PR #36, opened 2017-03-29
Fixes Open-Systems-Pharmacology/PK-Sim#78 - Crash when starting obs data import

rwmcintosh:
Tests are not included here because the only change is in the UI code.

### DIFF
--- a/src/OSPSuite.UI/Views/Importer/ImporterView.cs
+++ b/src/OSPSuite.UI/Views/Importer/ImporterView.cs
@@ -472,7 +472,7 @@ namespace OSPSuite.UI.Views.Importer
 
             if (newTables.Count == 0)
             {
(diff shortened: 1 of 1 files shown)


src/OSPSuite.UI/Views/Importer/ImporterView.cs
```diff
@@ -629,27 +627,31 @@ private void cleanColumnMappingControls()
       private void cleanMemory()
       {
          _namingView = null;
-         _openSourceFileControl.OnOpenSourceFile -= openSourceFileEvent;
-         _openSourceFileControl = null;
+         if (_openSourceFileControl != null)
```
rwmcintosh: That block is the real change.

rwmcintosh:
These event handlers had previously been leaking. I think in older versions
there was not an exception but a leak.

PR #39, opened 2017-04-11
Fixes #97 - Cannot zoom-out after zoom-in simulated-plot

rwmcintosh:
the barManager.SetPopupContextMenu is apparently disabling the click handler from the UxChartControl for right click.

### DIFF
--- a/src/OSPSuite.UI/Controls/UxChartControl.cs
+++ b/src/OSPSuite.UI/Controls/UxChartControl.cs
@@ -32,10 +32,11 @@ namespace OSPSuite.UI.Controls
          _clipboardTask = new ClipboardTask();
          _barManager = new BarManager {Form = this};
          _popupMenu = new PopupMenu(_barManager);
-         _barManager.SetPopupContextMenu(this, _popupMenu);
 
-         if (addDefaultPopup)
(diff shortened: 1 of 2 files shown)


src/OSPSuite.UI/Controls/UxChartControl.cs
```diff
@@ -32,10 +32,11 @@ public UxChartControl(bool addDefaultPopup = true)
          _clipboardTask = new ClipboardTask();
          _barManager = new BarManager {Form = this};
          _popupMenu = new PopupMenu(_barManager);
-         _barManager.SetPopupContextMenu(this, _popupMenu);
 
-         if (addDefaultPopup)
-            initializePopup();
+         if (!addDefaultPopup) return;
+
+         _barManager.SetPopupContextMenu(this, _popupMenu);
```
msevestre: should not we put this line in the initializePopup?

rwmcintosh: done

msevestre:
The reference to issue was not accurate. This is issue 
Open-Systems-Pharmacology/PK-Sim#97

PR #40, opened 2017-04-11
Fixes Open-Systems-Pharmacology/PK-Sim/#94 Cannot run PI 

msevestre:
and also Fixes Open-Systems-Pharmacology/PK-Sim/#96


msevestre:
@rwmcintosh How do we reference an external issue again?
Open-Systems-Pharmacology/PK-Sim/#96 does not seem to do the trick

rwmcintosh:
Open-Systems-Pharmacology/PK-Sim/#96
Open-Systems-Pharmacology/PK-Sim#96

PR #43, opened 2017-04-12
Fixes #42 preview origin data button too small

### LINKED ISSUE #42: Preview Original Data button in Importer is too small

msevestre:
I have also updated SX to the latest version

### DIFF
--- a/src/OSPSuite.UI/Views/Importer/ImporterView.cs
+++ b/src/OSPSuite.UI/Views/Importer/ImporterView.cs
@@ -70,12 +70,21 @@ namespace OSPSuite.UI.Views.Importer
             prevPage.Appearance.Header.Font = Fonts.NonSelectedTabHeaderFont;
       }
 
+      public override void InitializeResources()
+      {
+         base.InitializeResources();
+         layoutItemSelectRange.AdjustLongButtonSize();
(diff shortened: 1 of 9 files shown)


src/OSPSuite.UI/Views/Importer/ImporterView.cs
```diff
@@ -70,12 +70,21 @@ private void changeTabs(IXtraTabPage page, IXtraTabPage prevPage)
...
       }
 
+      public override void InitializeResources()
+      {
+         base.InitializeResources();
+         layoutItemSelectRange.AdjustLongButtonSize();
+         layoutItemImportAll.AdjustButtonSize();
+         layoutItemImport.AdjustButtonSize();
+         btnSelectRange.InitWithImage(ApplicationIcons.PreviewOriginData, Captions.Importer.PreviewExcelData);
+         resetImportButtonText();
```
msevestre: This was actually NEVER called when creating the view. e.g icons were not set properly. Never realized it

msevestre: I had to remove all buttons and add them again because for some reason, their default behavior had been changed and thus our standard method were not working. Always a joy to work with the importer

PR #73, opened 2017-05-08
WIP Open-Systems-Pharmacology/InstallationValidator#88

rwmcintosh:
WIP Open-Systems-Pharmacology/InstallationValidator#88

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -306,7 +306,7 @@ namespace OSPSuite.Assets
          public static readonly string TheUnitInformationMustBeEnteredOrConfirmed = "The unit information must be entered or confirmed.";
          public static readonly string TheMetaDataInformationMustBeEnteredOrConfirmed = "The meta data must be entered or confirmed.";
 
-    
+
          public class ToolTips
          {
             public static readonly string NamingPattern = "Set a pattern for renaming imported data";
@@ -1016,6 +1016,23 @@ namespace OSPSuite.Assets
             public static readonly string FontSizeDescription = "Font Size Description";
             public static readonly string FontSizeOrigin = "Font Size Origin";
          }
+      }
+
+      public static string FormatForRichEditExceptionView(string value, string fontFamily, double fontSize)
+      {
+         return @"
+            <style>
+               div {
+                   font-family: '" + fontFamily + @"';
+                  font-size: '" + fontSize + @"';
+               }
+
+               p {
+                  padding: 0px;
+                  margin: 5px;
+               } 
+            </style>
+            <div>" + value + "  </div>";
       }
    }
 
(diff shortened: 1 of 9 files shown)


src/OSPSuite.UI/Controls/UxRichEditControl.cs
```diff
@@ -15,5 +18,37 @@ private void onUnhandledException(object sender, RichEditUnhandledExceptionEvent
       {
          richEditUnhandledExceptionEventArgs.Handled = true;
       }
+
```
rwmcintosh: The RichEdit did not behave exactly as the label did during resize so IXtraResizableControl had to be implemented.

Do you think we should create a separate control for this purpose - making a richedit act like a label?

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -241,19 +241,31 @@ public static string DoYouWantToDeleteDirectory(string newDirectoryName)
          return $"Do you want to delete the directory '{newDirectoryName}' and continue?";
       }
 
+      private static void appendListItem(string listItem, StringBuilder sb)
```
msevestre: Should not we make those methods public somewhere so that then can be used from other components in the future?

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1005,6 +1017,23 @@ public static class FontAndSizeSettings
             public static readonly string FontSizeOrigin = "Font Size Origin";
          }
       }
+
+      public static string FormatForRichEditExceptionView(string caption, string fontFamily, double fontSize)
+      {
+         return @"
```
msevestre: String interpolation does not work here?

src/OSPSuite.UI/Controls/UxRichEditControl.cs
msevestre: do you mean like a HtmlLabel or sthg?

src/OSPSuite.UI/Controls/UxHtmlLabel.cs
```diff
@@ -0,0 +1,73 @@
...
+      public void Caption(string caption, string fontFamily = "", double? fontSize = 8.25)
+      {
+          Document.HtmlText = stylizeHtmlText(fontFamily, fontSize) + $"<div>{caption}</div>";
+      }
+
+      private static string stylizeHtmlText(string fontFamily, double? fontSize)
+      {
+         return $@"
+            <style>
+               div {{
```
rwmcintosh: Yeah, you can use string interpolation. Just need to double up the curly braces.

src/OSPSuite.UI/Controls/UxHtmlLabel.cs
```diff
@@ -0,0 +1,84 @@
...
+
+      Size IXtraResizableControl.MinSize => Size;
+
+      protected override void OnSizeChanged(EventArgs e)
+      {
+         base.OnSizeChanged(e);
+         RaiseChanged();
+      }
+
+      public void Caption(string caption, string fontFamily = "", double? fontSize = 8.25)
```
msevestre: Define a constant for that number here somewhere (UICOnstants probably?)

PR #75, opened 2017-05-10
Fixes #74 Update to UxHtmlLabel to make it usable in other components

### LINKED ISSUE #74: Update to UxHtmlLabel to make it usable in other components

msevestre:


src/OSPSuite.UI/Controls/UxHtmlLabel.cs
```diff
@@ -1,51 +1,40 @@
...
 using DevExpress.XtraEditors.Controls;
 using DevExpress.XtraRichEdit;
+using AutoSizeMode = DevExpress.XtraRichEdit.AutoSizeMode;
 
 namespace OSPSuite.UI.Controls
 {
    public class UxHtmlLabel : UxRichEditControl, IXtraResizableControl
    {
-      private static readonly object _layoutInfoChanged = new object();
-      private static string _fontFamily;
```
msevestre: Not sure why all those guys were static. This looks like a smell to me (why would all Label share the same variable, especially the layoutInfoChanged??)

src/OSPSuite.UI/Controls/UxHtmlLabel.cs
```diff
@@ -1,51 +1,40 @@
...
-      private static double _fontSize;
+      private readonly object _layoutInfoChanged = new object();
+      public string FontFamily { get; set; } = UIConstants.DEFAULT_HTML_FONT;
```
msevestre: Those properties should be available as set so that they can be updated from other components

rwmcintosh:
I think I recall that parameterless constructor has to be the first

PR #86, opened 2017-05-23
 	Fixes #84 ReportTask should export to XLXS if required

### LINKED ISSUE #84: ReportTask should export to XLXS if required
see https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/181

msevestre:


msevestre:
@Yuri05 We add a method for that already. It was just not used

PR #97, opened 2017-05-29
Fixes #96 - PI: Allow copying values from the results tab

### LINKED ISSUE #96: PI: Allow copying values from the results tab
Open-Systems-Pharmacology/MoBi#24

rwmcintosh:


src/OSPSuite.UI/Controls/UxMasterDetailGridView.cs
```diff
@@ -0,0 +1,13 @@
+using DevExpress.XtraGrid;
+
+namespace OSPSuite.UI.Controls
+{
+   public class UxMasterDetailGridView : UxGridView
+   {
+      protected override string ViewName => "UxMasterDetailGridView";
```
rwmcintosh: When I tried to apply this to UxGridView some exceptions were thrown from DevExpress. I created this class so I could isolate the behavior to just where we need these clones to be made from a subclass.

We need the grid control from the MultiplePIResultsView to create this kind of grid view when it makes a clone of optimizedParameterView.

msevestre: Not sure I understand why this subclass is required. I actually do not understand why we need the MasterXxxx . Can't we simply use UxGridView for all of that?
What exceptions are you referring to?

src/OSPSuite.UI/Controls/UxMasterDetailGridControl.cs
```diff
@@ -0,0 +1,32 @@
...
+      }
+
+      protected override void RegisterAvailableViewsCore(InfoCollection collection)
+      {
+         base.RegisterAvailableViewsCore(collection);
+         collection.Add(new UxGridViewInfoRegistrator());
+      }
+   }
+
+   public class UxGridViewInfoRegistrator : GridInfoRegistrator
```
msevestre: If we keep this extra UxMasterDetailGridView, then this class should be renamed

src/OSPSuite.UI/Controls/UxMasterDetailGridView.cs
rwmcintosh: The important detail is the ViewName property. That is how the grid control can create new views of the correct type. It's common to the view and a few other places in the subclass of gridcontrol.

I overrode the ViewName for UxGridView so that my subclass of grid would create UxGridView when cloning a OptimizedParameterView and I had NullReferenceExceptions from DevExpress during component resolution which were not easy to track down.

rwmcintosh: What I mean is, I think by changing that ViewName, we are introducing some new behavior in UxGridView that breaks something else.

rwmcintosh:
Such as a null reference exception being thrown from DevExpress during component creation.

I re-created it and it's from DataBrowser component.

PR #101, opened 2017-05-30
Fixes #99 - PK Parameters preview histogram jumbled

### LINKED ISSUE #99: PK Parameters preview histogram jumbled
Open-Systems-Pharmacology/PK-Sim#177

rwmcintosh:


src/OSPSuite.UI/Controls/UxHistogramControl.cs
```diff
@@ -79,11 +79,11 @@ public void Plot(ContinuousDistributionData distributionData, DistributionSettin
          if (!distributionData.IsRangeSpecified)
             return;
 
+         var templateView = SeriesTemplate.View as BarSeriesView;
```
rwmcintosh: This is the change. The other files are introducing the new Histogram Starter and refactoring starters so that they do not receive their presenter on construction, but only when the started.

src/OSPSuite.UI/Controls/UxHistogramControl.cs
```diff
@@ -79,11 +79,11 @@ public void Plot(ContinuousDistributionData distributionData, DistributionSettin
...
+         if (templateView != null)
+            templateView.BarWidth = distributionData.BarWidth;
+
          AxisX.WholeRange.SetMinMaxValues(distributionData.XMinValue, distributionData.XMaxValue);
-         foreach (var barSeriesView in barSeriesViews)
```
msevestre: Remove method barSeriesView. I don't think this is used anymore

rwmcintosh: done

PR #104, opened 2017-05-31
Fixes #103 Add compare hidden entities to comparison settings

### LINKED ISSUE #103: Add compare hidden entities to comparison settings
required for https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/158

msevestre:


msevestre:
Changes in designer is due to usage of UxCheckEdit instead of default CheckEdit

Issue #108, opened 2017-06-02
curve name does not contain simulation name after cloning


rwmcintosh:
Open-Systems-Pharmacology/PK-Sim#176

msevestre:
that's fixed correct?

rwmcintosh:
yes

PR #109, opened 2017-06-02
WIP - #108 curve name does not contain simulation name after cloning

rwmcintosh:
Some of this rename functionality can be shared if moved to core.

### DIFF
--- a/src/OSPSuite.Core/Services/CurveNamer.cs
+++ b/src/OSPSuite.Core/Services/CurveNamer.cs
@@ -50,12 +50,11 @@ namespace OSPSuite.Core.Services
 
       private IEnumerable<ICurve> curvesFromChartWithOriginalName(ISimulation simulation, ICurveChart chart)
       {
-         var curvesWithOriginalName = chart.Curves.Where(curve =>
+         return chart.Curves.Where(curve =>
          {
             var dataColumn = curve.yData;
             return hasOriginalName(simulation, dataColumn, curve);
          });
-         return curvesWithOriginalName;
       }
    }
 }
--- a/src/OSPSuite.Core/Services/CurveNamer.cs
+++ b/src/OSPSuite.Core/Services/CurveNamer.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
 using OSPSuite.Core.Chart;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Data;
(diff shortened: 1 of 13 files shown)


src/OSPSuite.Core/Services/CurveNamer.cs
```diff
@@ -0,0 +1,63 @@
...
+      }
+
+      private bool hasOriginalName(ISimulation simulation, DataColumn dataColumn, ICurve curve)
+      {
+         var originalCurveName = curveNameForDataColumn(simulation, dataColumn);
+         return string.Equals(curve.Name, originalCurveName);
+      }
+
+      public IReadOnlyList<ICurve> CurvesWithOriginalName(ISimulation simulation, IEnumerable<ICurveChart> curveCharts)
+      {
```
msevestre: That can become a one line by simply returning curveCharts.SelectMany(x=>curvesFromChartWithOrigianlName(s, x));

src/OSPSuite.Core/Services/CurveNamer.cs
```diff
@@ -0,0 +1,63 @@
...
+         curveCharts.Each(chart =>
+         {
+            curvesHavingOriginalName.AddRange(curvesFromChartWithOriginalName(simulation, chart));
+         });
+         return curvesHavingOriginalName;
+      }
+
+      private IEnumerable<ICurve> curvesFromChartWithOriginalName(ISimulation simulation, ICurveChart chart)
+      {
+         var curvesWithOriginalName = chart.Curves.Where(curve =>
```
msevestre: Return right away

src/OSPSuite.Core/Services/CurveNamer.cs
```diff
@@ -0,0 +1,63 @@
...
+using OSPSuite.Core.Domain.Data;
+using OSPSuite.Core.Domain.Mappers;
+using OSPSuite.Utility.Extensions;
+
+namespace OSPSuite.Core.Services
+{
+   public interface ICurveNamer
+   {
+      string CurveNameForColumn(ISimulation simulation, DataColumn dataColumn);
+      IReadOnlyList<ICurve> CurvesWithOriginalName(ISimulation simulation, IEnumerable<ICurveChart> curveCharts);
```
msevestre: Not sure exactly as to what that method is doing. Either rename the method or add a comment explaining what Original means in regards to the simulation

src/OSPSuite.Core/Services/CurveNamer.cs
msevestre: you can return here right away instead of saving in local variable and returning the variable. Good to merge after that

PR #111, opened 2017-06-05
Fixes #110 adjust control heights

### LINKED ISSUE #110: AdjustControlHeight does not work when shrinking the control height
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/f75c2b2f6a639a8a8f27e4937358ad7819e87ba0/src/OSPSuite.UI/Extensions/LayoutControlItemExtensions.cs#L66

if the ControlMaxSize becomes small thant eh ControlMinSize (temporarilly, the value is not taken into account)

msevestre:


msevestre:
fixing that issue plus updating all dlls to latest (especially SmartXls that just expired)

src/OSPSuite.UI/Extensions/LayoutControlItemExtensions.cs
```diff
@@ -65,23 +73,18 @@ public static void AdjustControlSize(this LayoutControlItem layoutControlItem, i
          layoutControlItem.SizeConstraintsType = SizeConstraintsType.Custom;
          layoutControlItem.ControlMaxSize = size;
          layoutControlItem.ControlMinSize = size;
+         //In case the control size became smaller that the previous min size, the max size is ignored. So we set it again 
+         layoutControlItem.ControlMaxSize = size;
```
msevestre: This is the fix. If shrinking the control, ControlMinSize may be bigger than ControlMaxSize and the value is not accepted . Could be a new behavior from DevExpress...

src/OSPSuite.UI/Views/DisplayUnitsView.cs
```diff
@@ -80,17 +80,14 @@ public void BindTo(IEnumerable<DefaultUnitMapDTO> defaultUnits)
...
-      }
+      public override bool HasError => _gridViewBinder.HasError;
 
       public override void InitializeResources()
       {
          base.InitializeResources();
-         layoutItemAddUnitMap.AdjustButtonSize(UIConstants.Size.BUTTON_WIDTH, UIConstants.Size.LARGE_BUTTON_HEIGHT);
-         layoutItemLoadUnits.AdjustButtonSize(UIConstants.Size.BUTTON_WIDTH, UIConstants.Size.LARGE_BUTTON_HEIGHT);
-         layoutItemSaveUnits.AdjustButtonSize(UIConstants.Size.BUTTON_WIDTH, UIConstants.Size.LARGE_BUTTON_HEIGHT);
+         layoutItemAddUnitMap.AdjustTallButtonSize();
```
msevestre: Just created that method so that the knowledge of what is small, tall large etc. is only in one place

msevestre:
Merging as it will be required to get further build working

PR #131, opened 2017-06-12
Fixes #89 - Update: ChartTest program

### LINKED ISSUE #89: Update: ChartTest program
* Data Repository always creates the same curves. This is ok for this one problem we try to solve but will be an issue if we try to solve/understand other problems with this tool. Add some randomization (I think the previous code was doing a great job at generating curves that were not monotonous)

* Generated Observed Data do not represent are usual observed data. We need to have data with Errors, and less point that the Simulation data. Especially, the number of point and X values should not always be the same

* Last but not least, we also need some curves with GeoMean and Arithmetical Mean set. This is one of the feature that the chart is provided and we cannot test it anymore

rwmcintosh:


tests/OSPSuite.Starter/Tasks/DataRepositoryCreator.cs
```diff
@@ -70,36 +82,106 @@ private static IQuantity getQuantityFromOrganism(IContainer organism)
...
+         for (var i = 0; i < numberOfCalculations; i++)
          {
-            dataRepository.Add(calculationColumnFor(baseGrid, i, model, columnOrigins));
+            dataRepository.Add(createPrimaryColumnFor(baseGrid, i, model, columnOrigins));
          }
          return dataRepository;
       }
 
-      public IEnumerable<DataRepository> CreateObservationRepositories(int numberOfObservations, IContainer model, int index)
+      private DataColumn createAuxiliaryColumn(DataColumn column, int index, IContainer model, AuxiliaryType columnType, ColumnOrigins columnOrigins, Func<double, float> calculateAuxiliaryFactor)
```
rwmcintosh: When you create the auxiliary column, you supply a Func that will take a random value from 0 - 1 and turn it into a factor that is multiplied by the column value to give the auxiliary value.

rwmcintosh:
I think the calculations make sense with auxiliary types. Little tough to verify until the binders are available.

tests/OSPSuite.Starter/Tasks/DataRepositoryCreator.cs
msevestre: hum. this is only valid for Arithmetic STD. This should not be used for GeoSTD

rwmcintosh: Are you talking about the method in general, or the factor?

msevestre:
```
GeoMean => value = Mean*Factor and Mean/Factor
ArithMEan => value = Mean + factor and Mean - factor
```

so you should only multiply with the factor for arithmean. Otherwise use the factor as is

PR #132, opened 2017-06-12
Fixes #121 Reset of axis should also remove min and max values if set

### LINKED ISSUE #121: Reset of axis should also remove min and max values if set
for example when removing all curves, min and max remains as well as display unit

msevestre:


src/OSPSuite.Core/Chart/Axis.cs
```diff
@@ -134,6 +134,13 @@ public IDimension Dimension
          }
       }
 
+      public void Reset()
```
msevestre: This is the actual reset code called when the curve is being removed and axis reseted

src/OSPSuite.Core/Chart/CurveChart.cs
```diff
@@ -240,10 +236,10 @@ private Axis addNewAxisFor(AxisTypes axisType)
       private void updateAxesForRemovedCurve(AxisTypes yAxisType)
       {
          if (!Curves.Any())
-            _axes[AxisTypes.X].Dimension = null;
+            _axes[AxisTypes.X].Reset();
```
msevestre: This is the call

tests/OSPSuite.Core.Tests/Core/CurveChartSpecs.cs
```diff
@@ -1,20 +1,37 @@
...
+      protected Curve _curveOnYAxis;
+      protected Curve _curveOnY2Axis;
+      protected DataRepository _obsData1;
+      protected DataRepository _obsData2;
+      protected Axis _xAxis;
+      protected Axis _yAxis;
+
       protected override void Context()
       {
-         sut = new CurveChart();
```
msevestre: It's time to get more coverage here. Removing usage of fake and creating real data

tests/OSPSuite.Core.Tests/Core/CurveChartSpecs.cs
```diff
@@ -50,27 +67,31 @@ public void the_relative_legend_indicies_should_be_adjusted()
...
       protected override void Because()
       {
-         sut.SetyData(_curve, A.Fake<DataColumn>(), A.Fake<IDimensionFactory>());
+         _y2Axis = sut.AddNewAxisFor(AxisTypes.Y2);
       }
 
       [Observation]
       public void the_default_y_axis_scaling_should_be_used_for_the_new_axis()
       {
-         sut.Axes.First().Scaling.ShouldBeEqualTo(Scalings.Linear);
```
msevestre: This was wrong. First returns X Axis in that case but the test should have tested Y axis scaling....

PR #135, opened 2017-06-12
Fixes #119 - Why do we need to call RefreshAxisBinder twice

### LINKED ISSUE #119: Why do we need to call RefreshAxisBinder twice
There is something funky going on with ranges and axis

rwmcintosh:


src/OSPSuite.UI/Binders/AxisBinder.cs
```diff
@@ -174,14 +174,15 @@ private void adjustAxisMinMax()
       private void setAxisRange(bool sideMarginsEnabled, Size diagramSize)
       {
          var axisWidthInPixel = Axis.AxisType == AxisTypes.X ? diagramSize.Width : diagramSize.Height;
+         _axisView.Logarithmic = Axis.Scaling == Scalings.Log;
```
rwmcintosh: First problem was that the logarithmic property was not set before calculating the range. When going from log-linear that meant the 0.7 side margin was applied.

Because we responded to many events in the past, this was always calculated multiple times. After the first, this flag was set correctly at method entry.

src/OSPSuite.Presentation/Presenters/Charts/ChartDisplayPresenter.cs
```diff
@@ -404,14 +404,11 @@ private void updateChart()
...
 
-            RefreshAxisBinders();
-
             //maybe can be done better
             View.ReOrderLegend();
          }
          updateViewLayout();
 
-         //TODO WHY DO WE NEED IT TWICE ? 
          RefreshAxisBinders();
```
rwmcintosh: This has to be called outside the updater because when resetting the axis visual range, first the axis whole range is set to auto, then it's automatically calculated extents are used to set the visual range.

When we perform the refresh inside the updater, the extents are not calculated it seems, so we always set the visual range min to 0. This is corrected for log scale to Float.Epsilon.

msevestre:
nice

PR #136, opened 2017-06-12
Fixes #126 edit of a single curve or axis

### LINKED ISSUE #126: Edit of single curve or axis via SingleEditPresenter should also trigger chart update

msevestre:


src/OSPSuite.Presentation/Presenters/Charts/ChartDisplayPresenter.cs
```diff
@@ -580,30 +573,23 @@ private bool canHandle(ChartEvent chartEvent)
...
+      public void ActivateFirstContextMenuEntryFor(Curve curve)
       {
-         _axisContextMenuFactory.CreateFor(axis, this).Show(_view, popupLocation);
+         _contextMenuFactory.CreateFor(viewItemFor(curve), this).ActivateFirstMenu();
       }
 
-      public void ShowContextMenu(CurveChart curveChart, Point popupLocation)
+      public void ShowContextMenu(IViewItem viewItem, Point popupLocation)
       {
-         var contextMenu = _curveChartContextMenuFactory.CreateFor(curveChart, this);
```
msevestre: The whole context menu implementation was sub-optimal here. We were using one explicit factory per type whereas the presenter should only use one and delegate to the factory the creation of the context menu. This will make this presenter a bit lighter (in terms of dependencies that is)

msevestre:
There was no tests at all on those presenters. I did not add them yet but this is something that should be done as part of our redesign.

PR #142, opened 2017-06-13
Fixes #137 - Crash setting Max = 0 for log scale

### LINKED ISSUE #137: Chart: Crash when setting Max = 0 for an axis in log scale
see https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/223

rwmcintosh:
That was formerly causing StackOverflowException in the event loops.

Now that we only respond to property changes once, the programs don't crash, but there was still an exception stack shown to the user because you cannot set a DevExpress Axis visual or whole range with the min > max. 

It should also not be possible to set the max to 0 for log scale. I thought about the same rule for min, but there's already an overriding behavior when you try to set min = 0 for log scale.

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -1418,6 +1418,23 @@ namespace OSPSuite.Assets
          var separator = "\n\t- ";
          return $"A circular reference was found in formula of {entityType.ToLower()} '{entity}' with path '{entityPath}'\nPlease check the direct or indirect references:{separator}{references.ToString(separator)}";
       }
+
+      public static string AxisMaxMustBeGreaterThanOrEqualToAxisMin(float? axisMinimumValue)
+      {
+         var preamble = "The axis maximum value should be greater than or equal to the axis minimum value";
+         return axisMinimumValue.HasValue ? $"{preamble} '{axisMinimumValue}'" : preamble;
+      }
+
+      public static string AxisMinMustBeLessThanOrEqualToAxisMax(float? axisMaximumValue)
+      {
+         var preamble = "The axis minimum value should be less than or equal to the axis maximum value";
+         return axisMaximumValue.HasValue ? $"{preamble} '{axisMaximumValue}'" : preamble;
+      }
+
+      public static string AxisMaxMustBeGreaterThanZero()
+      {
+         return "The axis maximum must be greater than 0";
+      }
    }
 
    public static class Rules
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -1431,10 +1431,7 @@ namespace OSPSuite.Assets
          return axisMaximumValue.HasValue ? $"{preamble} '{axisMaximumValue}'" : preamble;
       }
 
-      public static string AxisMaxMustBeGreaterThanZero()
-      {
-         return "The axis maximum must be greater than 0";
-      }
+      public static readonly string LogAxisMaxMustBeGreaterThanZero = "Loagarithmic axis maximum must be greater than 0";
    }
 
    public static class Rules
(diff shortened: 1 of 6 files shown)


src/OSPSuite.UI/Binders/AxisBinder.cs
```diff
@@ -163,10 +163,14 @@ private void adjustAxisMinMax()
          else // somebody inserted a limit...
          {
             if (!Axis.Min.HasValue)
-               Axis.Min = Convert.ToSingle(_axisView.WholeRange.MinValue);
+            {
```
rwmcintosh: Here it was possible that user sets an explicit value on one property and the other property automatically calculates an inconsistent value (eg. min > max).

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1418,6 +1418,23 @@ public static string CircularReferenceFoundInFormula(string entity, string entit
...
+      }
+
+      public static string AxisMinMustBeLessThanOrEqualToAxisMax(float? axisMaximumValue)
+      {
+         var preamble = "The axis minimum value should be less than or equal to the axis maximum value";
+         return axisMaximumValue.HasValue ? $"{preamble} '{axisMaximumValue}'" : preamble;
+      }
+
+      public static string AxisMaxMustBeGreaterThanZero()
+      {
```
msevestre: This is for log scale only. Add the info to the message and create a static readonly prop (no parameter here)

src/OSPSuite.Core/Chart/Axis.cs
```diff
@@ -54,6 +58,45 @@ public Axis(AxisTypes axisType)
...
+            yield return minLessThanOrEqualToMax;
+            yield return maxGreaterThanZero;
+         }
+
+         private static IBusinessRule maxGreaterThanOrEqualToMin { get; } = maxGreaterThanOrEqualToMinForAxis();
+         private static IBusinessRule minLessThanOrEqualToMax { get; } = minLessThanOrEqualToMaxForAxis();
+         private static IBusinessRule maxGreaterThanZero { get; } = maxGreaterThanZeroForLogarithmicAxis();
+
+         private static IBusinessRule minLessThanOrEqualToMaxForAxis()
+         {
```
msevestre: use shorthand syntax =>?

src/OSPSuite.UI/Binders/AxisBinder.cs
```diff
@@ -163,10 +163,14 @@ private void adjustAxisMinMax()
...
+               var rangeMin = Convert.ToSingle(_axisView.WholeRange.MinValue);
+               Axis.Min = Axis.Max.HasValue ? Math.Min(Axis.Max.Value, rangeMin) :  rangeMin;
```
msevestre: Can we add some tests for this new behavior?

tests/OSPSuite.Core.Tests/Core/AxisSpecs.cs
```diff
@@ -12,19 +13,29 @@ public abstract class concern_for_Axis : ContextSpecification<Axis>
...
-      protected override void Context()
+      [Observation]
+      public void invalid_for_max_less_than_min()
       {
-         sut = new Axis(_axisType);
+         new Axis(AxisTypes.X)
+         {
+            Min = 0,
+            Max = -1
+         }.IsValid().ShouldBeEqualTo(false);
```
msevestre: ShouldBeFalse

tests/OSPSuite.Core.Tests/Core/AxisSpecs.cs
```diff
@@ -12,19 +13,29 @@ public abstract class concern_for_Axis : ContextSpecification<Axis>
...
-      public void test_for_type_should_be_true()
+      public void invalid_for_log_max_equal_to_0()
       {
-         sut.IsYAxis.ShouldBeTrue();
+         new Axis(AxisTypes.X)
+         {
+            Scaling = Scalings.Log,
+            Min = 0,
+            Max = 0
+         }.IsValid().ShouldBeEqualTo(false);
```
msevestre: ShouldBeFalse

PR #149, opened 2017-06-14
Fixes #148 Switch lin/log in simple chart should update chart

### LINKED ISSUE #148: Switch lin/log in simple chart should update chart

msevestre:


src/OSPSuite.Presentation/Presenters/Charts/SimpleChartPresenter.cs
```diff
@@ -62,6 +62,11 @@ public interface ISimpleChartPresenter : IPresenter<ISimpleChartView>
       ///    Action run when a series point is hot tracked in the chart
       /// </summary>
       Action<int> HotTracked { set; }
+
+      /// <summary>
+      /// Refresh the display after external changes were made to the chart
+      /// </summary>
+      void Refresh();
```
msevestre: because we expose the CurveChart explicitely, I am pretty sure that there is some code in PKsim or Mobi that is modifying the chart directly AFTER it was bound to the display presenter. Those spots will have to call refresh

Issue #72, opened 2017-05-05
Improvement of chart performance 


msevestre:
see description of tasks in project https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/projects/1

PR #171, opened 2017-06-16
WIP #72 ensure that title font size for PI is actually shown

msevestre:


msevestre:
For this settings to be used, the preview needs to be set as well. Before the code was bypassing the presenter alltogehter and setting the settings right in the view..

Issue #177, opened 2017-06-19
PredictedVsObserved does not work as expected with multiple outputs


msevestre:
Using test program:
* Start PI
* Add one output amount
* Add one output fraction
* select some parameters to optimize
* Run
* Open predicted vs observed. There is no curve. 

The problem is that the code was relying before on some stuff to happen with events and this is not happening anymore. This should not be required anyways.

The problem that I see is that we set up the chart correctly (in my case to use X as fraction) but then, a call to `_predictedVsObservedChartService.SetXAxisDimension(observationColumns, Chart);`  in `ParameterIdentificationPredictedVsObservedChartPresenter` reset the X axis to amount the chart is not ok anymore. Probably some explicit code needs to be added

msevestre:
@rwmcintosh : I am going to push some code on a branch so that you can have 2 outputs in the PI

PR #179, opened 2017-06-19
WIP #177: Ensure that fraction can also be selected as PI output

msevestre:


msevestre:
@rwmcintosh This is the branch you can use

PR #183, opened 2017-06-19
Fixes #182 data browser presenter clear

### LINKED ISSUE #182: DataBrowserPresenter. Clear should not delete binding

msevestre:


rakefile.rb
```diff
@@ -28,7 +28,7 @@
 
 def copy_to_app(app_target_relative_path)
   app_target_path = File.join(solution_dir, app_target_relative_path)
-  source_dir = File.join(src_dir, 'OSPSuite.UI', 'bin', 'Debug')
+  source_dir = File.join(tests_dir, 'OSPSuite.Starter', 'bin', 'Debug')
```
msevestre: copying form starter instead of UI just to be sure that we have every single DLL

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
```diff
@@ -144,9 +144,9 @@ public void RemoveDataColumns(IEnumerable<DataColumn> dataColumns)
 
       public void Clear()
       {
-         _view.DeleteBinding();
          _dataColumnDTOCache.Clear();
          _allDataColumns.Clear();
+         bindToView();
```
msevestre: binding to the view instead of deleting the binding. This may avoid issue with references not being release properly

msevestre:
I need to move on with that. Going to merge

PR #187, opened 2017-06-20
Fixes #185 Applying chart template to chart causes ChartChanged event

### LINKED ISSUE #185: Applying chart template to chart causes ChartChanged event
In certain cases a template is created from a chart, then is applied back to the chart. For this sequence the ChartChanged event should not be raised because the chart is being created exactly as it was before.

msevestre:


msevestre:
as discussed

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -555,7 +551,9 @@ public void Handle(ChartUpdatedEvent chartUpdatedEvent)
             return;
 
          Refresh();
-         ChartChanged();
+
+         if (chartUpdatedEvent.PropogateChartChangeEvent)
+            ChartChanged();
```
msevestre: event propagated only if  flag is set to true (default)

PR #189, opened 2017-06-20
Fixes #188 ChartSettingsPresneter and ChartExportSettingsPresenter sh…

### LINKED ISSUE #188: ChartSettingsPresneter and ChartExportSettingsPresenter should notify change event only on user actions

msevestre:
…ould notify change event only on user actions

### DIFF
--- a/src/OSPSuite.Core/Chart/ChartFonts.cs
+++ b/src/OSPSuite.Core/Chart/ChartFonts.cs
@@ -1,9 +1,8 @@
-using OSPSuite.Utility.Reflection;
 using OSPSuite.Core.Domain;
 
 namespace OSPSuite.Core.Chart
 {
-   public class ChartFonts : Notifier
+   public class ChartFonts : MyNotifier
    {
       private int _axisSize;
       private int _legendSize;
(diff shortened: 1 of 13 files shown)


src/OSPSuite.Core/Chart/ChartFonts.cs
```diff
@@ -25,61 +24,37 @@ public ChartFonts()
       public int AxisSize
       {
          get => _axisSize;
-         set
-         {
-            _axisSize = value;
-            OnPropertyChanged(() => AxisSize);
-         }
+         set => SetProperty(ref _axisSize, value, () => AxisSize);
```
msevestre: this should already prevent some of the events that we saw in case of update because we will set the same property all the time. That being said, let's do it right

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -230,8 +230,8 @@ public class ChartEditorPresenter : AbstractCommandCollectorPresenter<IChartEdit
          _presentersWithColumnSettings = new List<IPresenterWithColumnSettings> {_dataBrowserPresenter, _curveSettingsPresenter, _axisSettingsPresenter};
          initPresentersWithColumnSettings();
 
-         _chartExportSettingsPresenter.StatusChanged += (o, e) => onChartPropertiesChanged();
-         _chartSettingsPresenter.StatusChanged += (o, e) => onChartPropertiesChanged();
+         _chartExportSettingsPresenter.ChartExportSettingsChanged += onChartPropertiesChanged;
```
msevestre: Instead of the very generic StatusChanged, the presenter have their own event

msevestre:
Also some tests were added to other presenters

PR #191, opened 2017-06-20
Fixes #190  Color drag and drop in CurveSettingsPresenter does not wo…

### LINKED ISSUE #190: Color drag and drop in CurveSettingsPresenter does not work anymore

msevestre:
…rk anymore


msevestre:
Remove usage of tags everywhere as well.

PR #195, opened 2017-06-21
Fixes #177 - Predicted vs. Observed not working for multi-curves

### LINKED ISSUE #177: PredictedVsObserved does not work as expected with multiple outputs

rwmcintosh:


src/OSPSuite.Presentation/Services/ParameterIdentifications/PredictedVsObservedChartService.cs
```diff
@@ -106,16 +111,24 @@ public void SetXAxisDimension(IEnumerable<DataColumn> observationColumns, Parame
...
 
       public IReadOnlyList<DataRepository> AddIdentityCurves(IEnumerable<DataColumn> observationColumns, ParameterIdentificationPredictedVsObservedChart chart)
       {
          var identityCurves = addIdentityCurves(observationColumns, chart).ToList();
-         chart.UpdateAxesVisibility();
+         UpdateAxesVisibility(chart);
          return identityCurves;
       }
 
+      public void UpdateAxesVisibility(ParameterIdentificationPredictedVsObservedChart chart)
```
msevestre: Is there a reason why you would move this concern outside of the Chart when you already have a specific chart object?

src/OSPSuite.Presentation/Services/ParameterIdentifications/PredictedVsObservedChartService.cs
```diff
@@ -132,10 +145,32 @@ private IEnumerable<DataRepository> addIdentityCurves(IEnumerable<DataColumn> ob
...
+      private IDimension mostFrequentDimension(IReadOnlyList<DataColumn> columns)
+      {
+         var preferredDimension = mostFrequentDimensionFrom(columns.Where(isPreferredDimension).Select(mergedDimensionsFor));
+
+         if (preferredDimension != null)
+            return preferredDimension;
+
+         var dimensions = columns.Select(mergedDimensionsFor).ToList();
+
+         var firstGrouping = mostFrequentDimensionFrom(dimensions);
```
msevestre: return right away

PR #196, opened 2017-06-21
194 edit single axis cancel

msevestre:


msevestre:
* Adding some missing tests
* Ensure that naming is consistent (x vs X, BindTo vs BindToSource etc..)
* Remove property set in Curve when this should not be called for consistency reason

src/OSPSuite.Presentation/Presenters/Charts/SingleCurveSettingsPresenter.cs
```diff
@@ -11,21 +11,23 @@ public interface ISingleCurveSettingsPresenter : IPresenter<ISingleCurveSettings
...
       {
-         _view.BindTo(curve);
+         _cloneCurve = curve.Clone();
+         _view.BindTo(_cloneCurve);
 
          _view.Display();
          if(_view.Canceled)
             return;
 
+         curve.UpdateFrom(_cloneCurve);
```
msevestre: This was the original intent here. Ensure that when editing a single axis or curve, the user can cancel the edit

PR #204, opened 2017-06-27
Fixes #203 PI-Refresh problem when PI run is much smaller than refres…

### LINKED ISSUE #203: PI-Refresh problem when PI run is much smaller than refresh time

msevestre:
…h time


msevestre:
@rwmcintosh This is what we discussed this morning

PR #210, opened 2017-06-28
Fixes #230 Export of data to Excel from Chart does not export the cur…

### LINKED ISSUE #230: Automatic filling of "value description" and "value origin mode"  after Parameter Identification
Set "value origin mode"  to “Parameter Identification” and "value description" to  Name and Date of Parameter Identification automatically for those parameter values, which are transferred to simulation.

msevestre:
…ves using the selected units


msevestre:
Interestingly, we had a few issues here
* Export to csv could export in a unit that is not the base unit (should never be the case)
* Export to excel would not use the merge dimension (instead the column dimension would be used)

src/OSPSuite.Core/Domain/Services/IDataRepositoryTask.cs
```diff
@@ -14,19 +15,24 @@ public interface IDataRepositoryTask
       ///    Returns one datable for each base grid and associated columns representing the content of the repository (one column
       ///    for each DataColumn)
       ///    The column headers will be set to the column name();
-      ///    If the flag <paramref name="formatOutput" /> is set to true, the value will be formatted using the number of decimal
+      ///    If the flag <paramref name="formatOutput" /> is set to <c>true</c> (default is <c>false</c>), the value will be formatted using the number of decimal
       ///    places defined in the application
+      ///    If the flag <paramref name="useDisplayUnit" /> is set to <c>true</c> (default is <c>true</c>), the value will be exported in the display unit of the
```
msevestre: I have added this flag (default is true to keep the same behavior as before), that specifies whether the display unit should be used or not

src/OSPSuite.Core/Services/CurveChartExportTask.cs
```diff
@@ -0,0 +1,45 @@
...
+      public CurveChartExportTask(IDialogCreator dialogCreator, IDataRepositoryTask dataRepositoryTask, IDimensionFactory dimensionFactory)
+      {
+         _dialogCreator = dialogCreator;
+         _dataRepositoryTask = dataRepositoryTask;
+         _dimensionFactory = dimensionFactory;
+      }
+
+      public void ExportToExcel(CurveChart chart)
+      {
+         if (chart == null) return;
```
msevestre: Created a service for that just to isolate some behavior

src/OSPSuite.Core/Services/CurveChartExportTask.cs
```diff
@@ -0,0 +1,45 @@
...
+         var visibleCurves = chart.Curves.Where(x => x.Visible).ToList();
+         if (!visibleCurves.Any()) return;
+
+         var fileName = _dialogCreator.AskForFileToSave(Captions.ExportChartToExcel, Constants.Filter.EXCEL_SAVE_FILE_FILTER, Constants.DirectoryKey.REPORT, chart.Name);
+         if (string.IsNullOrEmpty(fileName)) return;
+
+         var dataColumnCache = new Cache<DataColumn, Curve>(onMissingKey: x => null);
+         visibleCurves.Each(curve => dataColumnCache[curve.yData] = curve);
+         _dataRepositoryTask.ExportToExcel(dataColumnCache.Keys, fileName, col => dataColumnCache[col]?.Name ?? col.Name, _dimensionFactory.MergedDimensionFor);
```
msevestre: Using the merge dimension when exporting

src/OSPSuite.Infrastructure/Services/SimulationResultsExporter.cs
```diff
@@ -27,7 +27,7 @@ public SimulationResultsExporter(IDataRepositoryTask dataRepositoryTask, IQuanti
 
       public Task ExportToCsvAsync(ISimulation simulation, DataRepository results, string fileName)
       {
-         var dataTable = _dataRepositoryTask.ToDataTable(results, x => _quantityDisplayPathMapper.DisplayPathAsStringFor(simulation, x)).First();
+         var dataTable = _dataRepositoryTask.ToDataTable(results, x => _quantityDisplayPathMapper.DisplayPathAsStringFor(simulation, x), x => x.Dimension, useDisplayUnit: false).First();
```
msevestre: Do not use display unit here!!

PR #220, opened 2017-07-04
Fixes #78 Remove Resizable hack form UxHtmlLabel

### LINKED ISSUE #78: Remove Resizable hack form UxHtmlLabel

msevestre:


msevestre:
@Yuri05 We tried to use an RichTextEdit html for label but it did not work as expected. Instead, I am formatting the text to be either html (real html) or devexpress html that can be displayed in a label (the same text is also used as pure HTML in the InstallationValidator)

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -241,31 +241,53 @@ public static string DoYouWantToDeleteDirectory(string newDirectoryName)
          return $"Do you want to delete the directory '{newDirectoryName}' and continue?";
       }
 
-      public static void AppendListItem(string listItem, StringBuilder sb)
+      public static void AppendListItem(StringBuilder sb, bool html, string listItem, int index)
```
Yuri05: where is `int index` used?

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -241,31 +241,53 @@ public static string DoYouWantToDeleteDirectory(string newDirectoryName)
...
       {
-         sb.Append($"<li>{listItem}</li>");
+         if(html)
+            sb.Append($"<li>{listItem}</li>");
+         else
+            sb.AppendLine($"   {index}: {listItem}");
```
msevestre: here?

Yuri05: lol

PR #226, opened 2017-07-20
Fixes #225 Delete multiple observed data. Message formatting looks st…

### LINKED ISSUE #225: Delete multiple observed data. Message formatting looks strange
Moreover, if no observed data cannot be deleted, the message should be a Message Ok and not Yes No

msevestre:
…range

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -188,7 +188,7 @@ namespace OSPSuite.Assets
          if (notDeleted.Count == 0)
             return prompt;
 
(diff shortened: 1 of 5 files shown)


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -188,7 +188,7 @@ public static string ReallyDeleteAllObservedData(IEnumerable<string> anythingNot
          if (notDeleted.Count == 0)
             return prompt;
 
-         return $"{prompt}\n\nWarning:\n\n{notDeleted.ToString("\n\n")}\n\nand will not be removed.";
+         return $"{prompt}\n\nWarning: Following observed data are still in used and will not be deleted: \n\n{notDeleted.ToString("\n\n")}";
```
Yuri05: in use

PR #251, opened 2017-08-17
Fixes #250 multiple project conversion

### LINKED ISSUE #250: Multiple Project conversion

msevestre:


src/OSPSuite.Core/Converter/IObjectConverter.cs
```diff
@@ -6,20 +6,20 @@ namespace OSPSuite.Core.Converter
 {
    public interface IObjectConverter : ISpecification<int>
    {
-      int Convert(object objectToUpdate);
-      int ConvertXml(XElement element);
+      (int convertedToVersion, bool conversionHappened) Convert(object objectToUpdate);
```
msevestre: Instead of only version, we return a flag indicating whether a conversion happened or not. This will be used in other converter in PKSim and MoBi to set flags accordingly

src/OSPSuite.Core/Converter/v5_4/Converter531To541.cs
```diff
@@ -1,38 +1,45 @@
...
          this.Visit(objectToUpdate);
-         return PKMLVersion.V5_4_1;
+         return (PKMLVersion.V5_4_1, _converted);
       }
 
-      public int ConvertXml(XElement element)
+      public (int convertedToVersion, bool conversionHappened) ConvertXml(XElement element)
       {
-         return PKMLVersion.V5_4_1;
+         return (PKMLVersion.V5_4_1, false);
```
msevestre: we return false when the method is a dummy.

src/OSPSuite.Core/Converter/v5_4/Converter531To541.cs
```diff
@@ -1,38 +1,45 @@
...
       }
 
       public void Visit(SimulationTransfer simulationTransfer)
       {
          var simulation = simulationTransfer.Simulation;
          convertSimulation(simulation);
+         _converted = true;
```
msevestre: Otherwise each public method sets the flag `_converted` to true

src/OSPSuite.Core/Converter/v5_5/Converter541To551.cs
```diff
@@ -28,26 +28,26 @@ public Converter541To551(IDimensionFactory dimensionFactory)
...
       {
          element.DescendantsAndSelfNamed("ParameterStartValues", "ParameterStartValuesBuildingBlock").Each(ConvertParameterStartValuesElement);
          element.DescendantsAndSelfNamed("MoleculeStartValues", "MoleculeStartValuesBuildingBlock").Each(ConvertMoleculeStartValueElement);
 
          //"ModelCoreSimulation" required for older project
          element.DescendantsAndSelfNamed("Simulation", "ModelCoreSimulation").Each(ConvertSimuationElement);
 
-         return PKMLVersion.V5_5_1;
+         //too many conversions going on in this converted. Assume conversion
+         return (PKMLVersion.V5_5_1, true);
```
msevestre: This is an old one and almost everything is changing. Return true. If we load such an old project, conversion happens the message will be shown even if no conversion happens. No big deal (and again, everything is converted :)

src/OSPSuite.Core/Converter/IObjectConverter.cs
msevestre: This is the new Syntax for C# 7 and implicit tuple. love it

msevestre:
Need to merge that now as I need to move forward with implementation in PK-Sim and MoBi

PR #269, opened 2017-10-27
Fixes #288 Allow FormulaTask to resolve dyanmic formula in a given co…

msevestre:
…ntainer


msevestre:
@Yuri05 This should be useful. Method should be called once the individual has been created. Obviously this is not enough but this is the first step

PR #272, opened 2017-11-07
Fixes #271 output selection hidden

### LINKED ISSUE #271: Output Selection is hidden when the output is selected
This is a bug due to DevExpress Update

msevestre:
@Yuri05 Thsoe Repository Items cannot be created once and used all the time anymore since DevExpress update. This actually makes sense and not sure why it worked like that before.
To avoid code repetition, I created a few extra method in the `UxRepositoryItemImageComboBox` class

### DIFF
--- a/src/OSPSuite.UI/Binders/OptimizedParametersBinder.cs
+++ b/src/OSPSuite.UI/Binders/OptimizedParametersBinder.cs
@@ -23,7 +23,6 @@ namespace OSPSuite.UI.Binders
       private GridViewBinder<OptimizedParameterDTO> _gridViewBinder;
       private readonly ValueDTOFormatter _valueFormatter = new ValueDTOFormatter();
       private readonly RepositoryItemPictureEdit _rangeRepository = new RepositoryItemPictureEdit();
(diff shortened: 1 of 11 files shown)


src/OSPSuite.UI/Binders/OptimizedParametersBinder.cs
```diff
@@ -73,9 +70,8 @@ private static void updateGridViewSelectionMode(GridView gridView)
 
       private RepositoryItem nameRepositoryFor(OptimizedParameterDTO optimizedParameterDTO)
       {
-         _nameRepository.Items.Clear();
-         _nameRepository.Items.Add(new ImageComboBoxItem(optimizedParameterDTO.Name, _imageListRetriever.ImageIndex(optimizedParameterDTO.BoundaryCheckIcon.IconName)));
-         return _nameRepository;
+         var nameRepository = new UxRepositoryItemImageComboBox(_gridViewBinder.GridView, _imageListRetriever);
```
msevestre: Instead of clearing the member, create a local one and add the item

PR #283, opened 2017-11-16
Fixes #282 pi chart update

### LINKED ISSUE #282: Parameter Identification: PredictedvsObserved Chart is not updated
see https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/468

msevestre:


src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationSingleRunAnalysisPresenter.cs
```diff
@@ -40,6 +40,7 @@ public ParameterIdentificationRunResult SelectedRunResults
          {
             UpdateTemplateFromChart();
             updateSelectedRunResults(value);
+            UpdateChartFromTemplate();
```
msevestre: This was missing. We were saving the template but never loading from template. It was working implicitly before probably...somehow

src/OSPSuite.UI/Extensions/ChartControlExtensions.cs
```diff
@@ -84,6 +84,14 @@ private static void setFontAndSizeSettings(ChartControl chartControl, int width,
 
       public static void CopyToClipboard(this UxChartControl chartControl, string watermark)
       {
+         if (string.IsNullOrEmpty(watermark))
+         {
+            //Bug in devexpress that does not keep the color in bar charts
+            chartControl.CopyChartToClipboard(chartControl);
```
msevestre: This is the change for the copy issue. Actually the reason is that the clone method does not copy the event handlers. Colors are set with event handlers for bars so this is the reason. The fix is to iterate over all series and copy the color. Too risky for now => e.g. draft plot may have wrong colors :)

PR #292, opened 2017-12-04
Fixes #291 Model should be ableo to retrieve MolWeight parameters for…

### LINKED ISSUE #291: Model should be ableo to retrieve MolWeight parameters for parameter defined in a Molecule
see https://github.com/Open-Systems-Pharmacology/MoBi/issues/209

and 
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/f75c2b2f6a639a8a8f27e4937358ad7819e87ba0/src/OSPSuite.Core/Domain/Model.cs#L55

msevestre:
… parameter defined in a Molecule

### DIFF
--- a/src/OSPSuite.Core/Domain/Model.cs
+++ b/src/OSPSuite.Core/Domain/Model.cs
@@ -47,7 +47,6 @@ namespace OSPSuite.Core.Domain
          if (Root == null || quantity == null)
             return null;
 
-         //Mol weight only available for molecule amount and observer
          var moleculeName = quantity.IsAnImplementationOf<IMoleculeAmount>() ? 
             quantity.Name : 
             quantity.ParentContainer?.Name;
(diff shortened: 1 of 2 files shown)


src/OSPSuite.Core/Domain/Model.cs
```diff
@@ -35,39 +37,32 @@ public IContainer Neighborhoods
...
+            Root?.Add(_neighborhoods);
          }
-         get { return _neighborhoods; }
+         get => _neighborhoods;
       }
 
       public double? MolWeightFor(IQuantity quantity)
       {
-         if (Root == null)
+         if (Root == null || quantity == null)
```
msevestre: This method was too restrictive I believe. It was returning the Molweight of the associated quantity only for `MoleculeAmount` AND `Observer`. But it should be possible to retrieve the MolWeight for all other types of quantity (I think the last remaining one is `Parameter`, but no need to be explicit about that here)

msevestre:
@Yuri05 review please

src/OSPSuite.Core/Domain/Model.cs
```diff
@@ -35,39 +37,32 @@ public IContainer Neighborhoods
...
             return null;
 
          //Mol weight only available for molecule amount and observer
```
Yuri05: the comment is not valid anymore

msevestre:
done

PR #297, opened 2017-12-15
Fixes #227 new selection field

### LINKED ISSUE #227: New selection field for "value origin mode"
Insert a selection field  and corresponding enumeration for “Value origin mode” with (perhaps iconized) entries e.g. for PK-Sim-DB, Article, Report, Internet, Project data, Parameter Identification, Manual Fit, Assumption, Unknown.

Background:
For model building documentation it is important to document the origin of all parameter values. 
Value description should be used to document "origin mode" and data source of all values carefully.

Origin mode should be entered by standardized selection (instead of typing it manually), which is easier for the user and allows filtering. By usage of different icons the space for display of the different origin modes can be minimized.

msevestre:


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -553,6 +553,18 @@ public static string WithParent(int uniqueIndex)
          }
       }
 
+      public static class ValueOrigins
```
msevestre: Those are all default stuff that we can change once we have the latest and greatest documentation. All icons are also dummy one at the moment

src/OSPSuite.Core/Comparison/QuantityDiffBuilder.cs
```diff
@@ -21,7 +21,7 @@ public override void Compare(IComparison<IQuantity> comparison)
          CompareValues(x => x.Dimension, x => x.Dimension, comparison);
          CompareValues(x => x.QuantityType, x => x.QuantityType, comparison);
          CompareValues(x => x.NegativeValuesAllowed, x => x.NegativeValuesAllowed, comparison);
-         CompareStringValues(x => x.ValueDescription, Captions.Comparisons.ValueDescription, comparison);
+        //TODO for MSD CompareStringValues(x => x.ValueDescription, Captions.Comparisons.ValueDescription, comparison);
```
msevestre: There is an entry for comparison so we'll take care of that there

src/OSPSuite.Core/Domain/Builder/StartValue.cs
```diff
@@ -125,52 +112,32 @@ private void entityFullPathToComponents(IObjectPath fullPath)
...
       public double? StartValue
       {
-         get { return _startValue; }
-         set
-         {
-            _startValue = value;
-            OnPropertyChanged(() => StartValue);
-         }
+         get => _startValue;
+         set => SetProperty(ref _startValue, value);
```
msevestre: Using the SetProperty method when not used already

src/OSPSuite.Core/Domain/ValueOrigin.cs
```diff
@@ -0,0 +1,31 @@
+﻿namespace OSPSuite.Core.Domain
+{
+   public class ValueOrigin
```
msevestre: This is the new class. Fairly simple with the OriginType and description together

src/OSPSuite.Core/Domain/ValueOriginType.cs
```diff
@@ -0,0 +1,65 @@
...
+      Unknown,
+   }
+
+   public static class ValueOriginTypes
+   {
+      private static readonly Cache<ValueOriginTypeId, ValueOriginType> _allValueOrigins = new Cache<ValueOriginTypeId, ValueOriginType>(x => x.Id);
+
+      public static IReadOnlyCollection<ValueOriginType> AllValueOrigins => _allValueOrigins;
+
+      public static ValueOriginType Assumption = create(ValueOriginTypeId.Assumption, Captions.ValueOrigins.Assumption, ApplicationIcons.AssumptionValueOrigin);
```
msevestre: Our types here. Usual extended enum so that we have icon and display available

tests/OSPSuite.Starter/Forms/GridViewForm.cs
```diff
@@ -36,19 +43,47 @@ private void initializeBinding()
...
             parameter.Formula = new ConstantFormula(i);
             yield return parameter;
          }
       }
    }
-}
+
+   class ParameterDTO : Parameter
+   {
+      public string ValueDescription
```
msevestre: Not sure if we will need those guys directly into the Parameter or not. This belongs more in a DTO but let's be ready to move that in Quantity if need be

PR #299, opened 2017-12-18
Fixes #230 Automatic filling of value description and value origin mo…

### LINKED ISSUE #230: Automatic filling of "value description" and "value origin mode"  after Parameter Identification

msevestre:
…de after Parameter Identification

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -563,6 +563,7 @@ namespace OSPSuite.Assets
          public static string ParameterIdentification = "Parameter Identification";
          public static string Report = "Report";
          public static string Unknown = "Unknown";
+         public static string Undefined = "Undefined";
       }
 
       public static class Reporting
@@ -914,6 +915,11 @@ namespace OSPSuite.Assets
          public static string SimulationDoesNotHaveOutputPath(string simulationName, string outputPath)
          {
             return $"Simulation {simulationName} does not have the output path {outputPath}";
+         }
+
+         public static string ValueUpdatedFrom(string parameterIdentification, string isoDate)
+         {
+            return $"Value updated from '{parameterIdentification}' on {isoDate}";
          }
       }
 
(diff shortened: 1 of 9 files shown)


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -915,6 +916,11 @@ public static string SimulationDoesNotHaveOutputPath(string simulationName, stri
          {
             return $"Simulation {simulationName} does not have the output path {outputPath}";
          }
+
+         public static string ValueUpdatedFrom(string parameterIdentification, string isoDate)
+         {
+            return $"Value updated from '{parameterIdentification}' on {isoDate}";
```
msevestre: That text may have to be revised. I just want to have the function in at the moment

src/OSPSuite.Core/Commands/UpdateValueOriginCommand.cs
```diff
@@ -0,0 +1,55 @@
+﻿using OSPSuite.Core.Commands.Core;
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Builder;
+using Command = OSPSuite.Assets.Command;
+
+namespace OSPSuite.Core.Commands
+{
+   public class UpdateValueOriginCommand : OSPSuiteReversibleCommand<IOSPSuiteExecutionContext>
```
msevestre: I have created a command so that undoing the action of setting the PI parameter will also undo the automatic set.
I would not create a command however for VAlueOrigin set manually (e.g. by the user as opposed to programatically)

src/OSPSuite.Core/Commands/UpdateValueOriginCommand.cs
```diff
@@ -0,0 +1,55 @@
...
+      {
+         var oldValueOriginDisplay = _valueOrigin.ToString();
+         _valueOrigin.Type = _newType;
+         _valueOrigin.Description = _newDescription;
+         Description = Command.UpdateValueOriginFrom(oldValueOriginDisplay, _valueOrigin.ToString());
+      }
+
+      protected override void ClearReferences()
+      {
+         //we keep the reference alive to allow for undo
```
msevestre: I think this is ok to keep a reference to the ValueOrigin being changed.

src/OSPSuite.Core/Commands/UpdateValueOriginCommand.cs
Yuri05: >I would not create a command however for VAlueOrigin set manually (e.g. by the user as opposed to programatically)

Why not? If a value is reseted to an older one, Value Origin should be reseted too, don't you think?

msevestre: I would create a command ONLY for value origin set automatically. Right now we do not have a command for Value Description and I think this is correct (Commands are only for actions that change the model output)

In this case, we update the valueorigin in code (without the user knowing about it). Undo (and not reset) should also undo the automatic change

Yuri05: ok, let's accept it for the moment.
might need further discussion in the future

PR #308, opened 2018-01-15
Fixes #307 add default flag to value origin

### LINKED ISSUE #307: Add the Default flag to ValueOrigin

msevestre:


msevestre:
@Yuri05 FYI: Merging as I need to move forward but it is based on our discussion

PR #318, opened 2018-01-19
Fixes #315 method to enable disable colum in binder and Fixes #313

### LINKED ISSUE #313: Add Id as nullable int as value origin so that it is possible to track down the default valueorigin

### LINKED ISSUE #315: ValueOriginBinder should receive a method that will be used to decide if a cell is enabled or not
Default will be true but this will allow the calling app to decide if a given value origin should be editable.
This is especially useful when editing a parameter in simulation whose ValueOrigin should be in sync with other (such as MW and all Halogens etc..)

msevestre:


src/OSPSuite.Assets/ApplicationIcons.cs
```diff
@@ -457,16 +457,19 @@ public static class ApplicationIcons
...
-      public static readonly ApplicationIcon ManualFitValueOrigin = AddNamedIcon(Icons.ManualFitValueOrigin, "ManualFitValueOrigin");
-      public static readonly ApplicationIcon ParameterIdentificationValueOrigin = AddNamedIcon(Icons.ParameterIdentificationValueOrigin, "ParameterIdentificationValueOrigin");
-      public static readonly ApplicationIcon PublicationValueOrigin = AddNamedIcon(Icons.PublicationValueOrigin, "PublicationValueOrigin");
-      public static readonly ApplicationIcon UnknownValueOrigin = AddNamedIcon(Icons.UnknownValueOrigin, "UnknownValueOrigin");
-      public static readonly ApplicationIcon InVitroAssayValueOrigin = AddNamedIcon(Icons.OtherValueOrigin, "InVitroAssayValueOrigin");
-      public static readonly ApplicationIcon MeasurementValueOrigin = AddNamedIcon(Icons.OtherValueOrigin, "MeasurementValueOrigin");
-      public static readonly ApplicationIcon OtherValueOrigin = AddNamedIcon(Icons.OtherValueOrigin, "OtherValueOrigin");
+      public static readonly ApplicationIcon ValueOriginMethodAssumption = AddNamedIcon(Icons.ValueOriginMethodAssumption, "ValueOriginMethodAssumption");
```
msevestre: Update icons according to new list

src/OSPSuite.Assets/OSPSuite.Assets.csproj
```diff
@@ -1033,14 +1033,43 @@
...
-    <None Include="Icons\AssumptionValueOrigin.ico" />
-    <None Include="Icons\DatabaseValueOrigin.ico" />
-    <None Include="Icons\InternetValueOrigin.ico" />
-    <None Include="Icons\ManualFitValueOrigin.ico" />
-    <None Include="Icons\OtherValueOrigin.ico" />
-    <None Include="Icons\ParameterIdentificationValueOrigin.ico" />
-    <None Include="Icons\PublicationValueOrigin.ico" />
-    <None Include="Icons\UnknownValueOrigin.ico" />
+    <None Include="Icons\ValueOriginMethodAssumption.ico" />
+  </ItemGroup>
```
msevestre: How VS is organizing file is still a mystery for me :

src/OSPSuite.Core/Domain/ValueOrigin.cs
```diff
@@ -53,6 +98,33 @@ public void UpdateFrom(ValueOrigin valueOrigin)
 
       public string Display => defaultDisplay(this);
 
+      public int CompareTo(ValueOrigin other)
```
msevestre: Needs to implement IComparable and override Equals to allow filtering in UI. 
Because two value origin are the same if they have the same method, source and description, that's what I used to implement comparable. That should be ok as long as we do not do funky stuff with ValueORigins in hash or in dictionary. In that case, we will need to create ValueOriginDTO... but rather not to for such a simple object

src/OSPSuite.Core/Domain/ValueOrigin.cs
```diff
@@ -78,5 +150,24 @@ private static bool isUndefined(ValueOrigin valueOrigin)
...
+
+            if (isUndefined(this))
+               _key = string.Empty;
+
+            _key = new[]
+            {
+               Source.Id.ToString(), Method.Id.ToString(), Description
+            }.Where(x => !string.IsNullOrWhiteSpace(x)).ToString("-");
+
+            return _key;
```
msevestre: I did not want to have to compute the key all the time. so it's only done once and reset every time a property is changed

src/OSPSuite.UI/Binders/ValueOriginBinder.cs
```diff
@@ -74,21 +76,35 @@ private void updateValueOrigin(bool canceled)
...
          initializeToolTip(_gridView.GridControl);
       }
 
+      private void onShowingEditor(object sender, CancelEventArgs e)
+      {
+         if (!ColumnIsValueOrigin(_gridView.FocusedColumn))
+            return;
+
+         var withValueOrigin = _gridViewBinder.FocusedElement;
+         e.Cancel = !_valueOriginEditableFunc(withValueOrigin);
```
msevestre: if the value cannot be edited then the editor is simply not shown

msevestre:
Going to merge that. Please review at your earliest convenience

PR #325, opened 2018-01-23
Fixes #324 ValueOrigin update from should not update Default state by…

### LINKED ISSUE #324: ValueOrigin update from should not update Default state by default

msevestre:
… default


msevestre:
Updating the Default flag everytime has some side effects in PKSim when trying to ensure that the value origin is the same for all parameters belonging to the same group (such as MoWeight or PKA). I think that in the context of a clone, everything should be copied (included Id and default). Otherwise, only the actual ValueOrigin {source, Method etc..}{ should be copied. Maybe we are missing a level here to describe this behavior propertly

msevestre:
@Yuri05 Going to merge for now. Let's discuss it

PR #330, opened 2018-01-24
Fixes #329 Move ValueOrigin to Quantity as it is required for Molecul…

### LINKED ISSUE #329: Move ValueOrigin to Quantity as it is required for MoleculeAmount

msevestre:
…eAmount


msevestre:
Moved ValueOrigin to Quantity. IsDefault is however still only in Parameter for now

Issue #236, opened 2017-07-26
In comparison display value and "value description" when overriding formula values


ju-rgen:
In case of overriding formula values by manual values (Parameter Start Values) display value and value description. 
(currently partly no differences are shown, partly only text “new value”.

msevestre:
Would it be enough to show the Value description difference as a new entry in the difference table? I am not sure how it should look like otherwise. Please update this requirement so that we can estimate it.

msevestre:
@Yuri05 Maybe you can also add your inputs here. Thanks

Yuri05:
One option would be just to display the difference in **value description** as additional entry
(can we make sure, that it's always displayed e.g. _after_ the entry for the difference in **value**?)

Another option would be to introduce the difference in value description as additional columns (_old value descr_, _new value descr_)-but I think it would overcomplicate both implementation and display of comparison

msevestre:
You would also need new columns for value modes (old and new)
In my opinion, new raw would be much easier indeed

msevestre:
ok rows will be implemented

msevestre:
@Yuri05 Can you explain how it differs from #235 ?

msevestre:
@Yuri05 No idea what needs to be implemented here . Looks a lot like #235. Wording is confusing at best. Closing. Please reopen if required

PR #341, opened 2018-02-12
Fixes #235 Optional display of value description and value origin mod…

### LINKED ISSUE #235: Optional display of “value description” and “value origin mode” in comparison
Allow selection of “Value Origin Mode” and “Value Description” fields in comparison view in order to see the origins of different values directly.

msevestre:
…e in comparison

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -634,6 +634,7 @@ namespace OSPSuite.Assets
 
          public static readonly string OnlyComputeModelRelevantProperties = "Only compare properties relevant to simulation results";
          public static readonly string CompareHiddenEntities = "Compare hidden entities (e.g. parameters)";
+         public static readonly string ShowValueOriginForChangedValues = "Show value origin for changed values";
          public static readonly string FormulaComparisonValue = "Values";
          public static readonly string FormulaComparisonFormula = "Formulas";
          public static readonly string RunComparison = "Start";
--- a/src/OSPSuite.Core/Comparison/ComparerSettings.cs
+++ b/src/OSPSuite.Core/Comparison/ComparerSettings.cs
@@ -45,12 +45,19 @@ namespace OSPSuite.Core.Comparison
       /// </summary>
       public bool CompareHiddenEntities { get; set; }
 
+      /// <summary>
+      /// Specifies if Value origin should be displayed for compared quantities. Default is <c>true</c>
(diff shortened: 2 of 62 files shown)


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -634,6 +634,7 @@ public static string FormulaComparisonModeDescription
 
          public static readonly string OnlyComputeModelRelevantProperties = "Only compare properties relevant to simulation results";
          public static readonly string CompareHiddenEntities = "Compare hidden entities (e.g. parameters)";
+         public static readonly string ShowValueOriginForChangedValues = "Show value origin for changed values";
```
msevestre: According to specifications: Only show value origin when value have changed

src/OSPSuite.Core/Comparison/ComparerSettings.cs
```diff
@@ -45,12 +45,19 @@ public class ComparerSettings
...
+      public bool ShowValueOrigin { get; set; }
+
+
       public ComparerSettings()
       {
          OnlyComputingRelevant = true;
          FormulaComparison = FormulaComparison.Formula;
          RelativeTolerance = Constants.DOUBLE_RELATIVE_EPSILON;
          CompareHiddenEntities = false;
+         ShowValueOrigin = true;
```
msevestre: Default true

src/OSPSuite.Core/Comparison/WithValueOriginComparison.cs
```diff
@@ -0,0 +1,33 @@
...
+
+         comparisonAction(comparison);
+
+         if (!shouldAddValueOriginToComparison(numberOfItemsInComparison, comparison))
+            return;
+
+         diffBuilder.CompareValues(x => x.ValueOrigin, x => x.ValueOrigin, comparison);
+      }
+
+      private bool shouldAddValueOriginToComparison(int numberOfItemsInComparisonBeforeStartingComparison, IComparison<T> comparison)
```
msevestre: Strategy that defines if ValueOrigin should be shown or not
If we compare everything, I would suggest to show the difference no matter if the show value origin flag is set or not. (otherwise there is NO way to see that difference)

Otherwise, if the comparison actually performed some changes, we return the setting value

msevestre:
I think #236 is also fixed with this one

msevestre:
@Yuri05 I need to merge that to move forward. Please review when you have time

PR #343, opened 2018-02-13
Fixes #342 Add method to reorder favorites

### LINKED ISSUE #342: Add method to Favorite object to reorder entries

msevestre:


msevestre:
Fun to implement. Goal is to ensure that a set of favorite entries can be rearranged in the Favorite list. Because we have a List, we can simply save the list on order in xml. No need to save index anywhere

tests/OSPSuite.Core.Tests/Core/FavoritesSpecs.cs
```diff
@@ -0,0 +1,113 @@
+﻿using OSPSuite.BDDHelper;
+using OSPSuite.BDDHelper.Extensions;
+using OSPSuite.Core.Domain;
+
+namespace OSPSuite.Core
+{
+   public abstract class concern_for_Favorites : ContextSpecification<Favorites>
```
msevestre: Pure BDD implementation here. Defined the tests first and was done when they were all green :)

msevestre:
@Yuri05 Will have to merge this one as well. Please review at your earliest convenience

Issue #240, opened 2017-07-26
When attaching a Parameter Identification to a Journal Page, the used simulations should be stored automatically


ju-rgen:
When attaching a Parameter Identification as Related Item to a Journal Page the used simulations should be stored as independent items. So later the Parameter Identification can be used with other version of the simulations AND the original Parameter Identification can be redone.

msevestre:
see https://github.com/Open-Systems-Pharmacology/MoBi/issues/157

msevestre:
Since items will be saved independently from each other, restoring means that the user will need to restore all required items (this won't be done automatically)
Also all observed data should probably be added as related item as well.
I imagine that this feature should also be available for sensitivity analyses

msevestre:
Label risky because restore might not work at all with the current concept since we are resetting id's when loading a related item back into a project. a PI however references Simulation (internally by ids) so this might require a lot of work to get it right

msevestre:
Probably restore in an empty project could work (or a project where a simulation with the same id as a saved simulation does not exist)

msevestre:
@Yuri05 In my opinion this does not really belongs to the MBD package. I think the implementation could be quite tricky to really complicated. I am not sure how you want to proceed with this task. 
We could invest up to 2d and see if we are making progress, at least in some simple use cases (reimporting in empty project etc..)

Yuri05:
>We could invest up to 2d and see if we are making progress...

:+1:

PR #348, opened 2018-02-15
Fixes #239 allow attachment of arbitrary files as related items

### LINKED ISSUE #239: Allow attachments of general files to a Journal Page
Allow attachments of other files like Excel files, pdf Reports etc. to a Journal page as Related Items.

This supports using the Working Journal as a comprehensive "single source" of model documentation.

msevestre:


src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -87,15 +87,17 @@ public static class Constants
...
       public const string STD_DEV_GEOMETRIC = "Geometric Standard Deviation";
       public const string STD_DEV_ARITHMETIC = "Arithmetic Standard Deviation";
       public const string AUXILIARY_TYPE = "AuxiliaryType";
       public const string FILE = "File";
       public const string SHEET = "Sheet";
-
       public const string DEFAULT_WATERMARK_TEXT = "DRAFT";
 
+      public const int RELATIVE_ITEM_FILE_SIZE_WARNING_THRESHOLD_IN_BYTES = 5_000_000; 
+      public const int RELATIVE_ITEM_MAX_FILE_SIZE_IN_BYTES = 50_000_000;
```
msevestre: 50MB max. 5 MB warning

src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -225,7 +227,9 @@ public static class Filter
          public static readonly string XLSX_EXTENSION = ".xlsx";
          public static readonly string TEXT_EXTENSION = ".txt";
          public static readonly string JSON_EXTENSION = ".json";
-         public static readonly string DOCX_EXTENSION = ".docs";
+         public static readonly string DOCX_EXTENSION = ".docx";
```
msevestre: hum....

src/OSPSuite.Core/Journal/JournalPage.cs
```diff
@@ -48,7 +48,8 @@ public virtual Content Content
          set
          {
             _content = value;
-            ContentId = _content.Id;
+            if (_content != null)
```
msevestre: Resetting the content of the related item when loading a file. No need to keep that in memory!

src/OSPSuite.Core/Journal/RelatedItemFactory.cs
```diff
@@ -39,22 +51,67 @@ public class RelatedItemFactory : IRelatedItemFactory
...
-            FullPath = _projectRetriever.CurrentProject.FilePath,
-            Discriminator = _applicationDiscriminator.DiscriminatorFor(relatedObject),
-            Description = _relatedItemDescriptionCreator.DescriptionFor(relatedObject)
          };
+
+         configurationFunc(relatedItem);
+         return relatedItem;
+      }
+
+      private ApplicationIcon iconFromFileExtension(string extension)
```
msevestre: Trying to be funky here to have different icons depending on file extensions so that you can quickly see what is which

src/OSPSuite.UI/Views/Journal/JournalPagePreviewView.cs
```diff
@@ -50,15 +49,17 @@ public void BindTo(JournalPageDTO journalPageDTO)
       {
          _journalPageDTO = journalPageDTO;
          _screenBinder.BindToSource(journalPageDTO);
-         updateControlVisibility(relatedItemVisible: journalPageDTO.RelatedItems.Any(), tagVisible: true);
+         updateControlVisibility(relatedItemVisible: true, tagVisible: true);
```
msevestre: This should always be visible know because of the add button!

src/OSPSuite.UI/Views/Journal/JournalPagePreviewView.cs
```diff
@@ -75,15 +76,15 @@ public void AddRelatedItemsView(IView view)
 
       private void updateControlVisibility(bool relatedItemVisible, bool tagVisible)
       {
-         layoutItemRelatedItems.Visibility = LayoutVisibilityConvertor.FromBoolean(relatedItemVisible);
-         layoutItemTags.Visibility = LayoutVisibilityConvertor.FromBoolean(tagVisible);
+         layoutGroupRelatedItems.Visibility = LayoutVisibilityConvertor.FromBoolean(relatedItemVisible);
+         layoutGroupItemTags.Visibility = LayoutVisibilityConvertor.FromBoolean(tagVisible);
```
msevestre: looks nicer in a group

msevestre:
@Yuri05 Need to merge this feature for #240

src/OSPSuite.Core/Domain/Constants.cs
Yuri05: I would define it as 5 * 1024 * 1024 and 50 * 1024 * 1024

src/OSPSuite.Core/Journal/RelatedItemFactory.cs
Yuri05: Should we also have explicit icons for some fore file types?
* Matlab (*.m, *.mat, *.fig)
* R (*.r, *.rd)
* pkml

msevestre: Sure

src/OSPSuite.Core/Domain/Constants.cs
msevestre: yep ok. But I need to convert back then to MB when displaying to user.

PR #354, opened 2018-02-19
Fixes #326 PI: random start values are not generated

### LINKED ISSUE #326: PI: random start values are not generated
Parameter Identification -> Configuration -> Options -> Multiple optimization (Randomize start values) 

Problem: Initial values are not randomized
Version: PK-Sim 7.2.1. Standalone version
snap

msevestre:


msevestre:
@Yuri05 I had to update freaking SmartXLS so that the build would pass. Aside from that, fix is trivial and as far as I can tell, bug it not new [same line in 7.1](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/v7.1.0.99/src/OSPSuite.Presentation/Mappers/ParameterIdentifications/ParameterIdentificationRunResultToRunResultDTOMapper.cs#L530)

src/OSPSuite.Presentation/Mappers/ParameterIdentifications/ParameterIdentificationRunResultToRunResultDTOMapper.cs
```diff
@@ -56,16 +56,16 @@ private OptimizedParameterDTO mapFrom(OptimizedParameterValue optimizedParameter
...
       }
 
-      private OptimizedParameterDTO mapFrom(IdentificationParameter identificationParameter, double optimalValue)
+      private OptimizedParameterDTO mapFrom(IdentificationParameter identificationParameter, double optimalValue, double startValue)
       {
          var dto = new OptimizedParameterDTO
          {
             Name = identificationParameter.Name,
             OptimalValue = mapFrom(optimalValue, identificationParameter.StartValueParameter),
-            StartValue = mapFrom(identificationParameter.StartValue, identificationParameter.StartValueParameter),
```
msevestre: We were always taking the start value from the IdentificationParameter object instead of the actual start value of the result

src/OSPSuite.Presentation/Mappers/ParameterIdentifications/ParameterIdentificationRunResultToRunResultDTOMapper.cs
```diff
@@ -56,16 +56,16 @@ private OptimizedParameterDTO mapFrom(OptimizedParameterValue optimizedParameter
          if (identificationParameter == null)
             return null;
 
-         return mapFrom(identificationParameter, optimizedParameterValue.Value);
+         return mapFrom(identificationParameter, optimizedParameterValue.Value, optimizedParameterValue.StartValue);
```
msevestre: Start value needs to be passed in the method

PR #358, opened 2018-02-21
Fixes #357 side margine emable won't work

### LINKED ISSUE #357: Side Margins Enabled will not work if unchecked in Parameter Identification
see https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/562

msevestre:


src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -26,8 +26,7 @@ namespace OSPSuite.Presentation.Presenters.Charts
    ///    - ChartOptions for editing properties of chartSettings.
    /// </summary>
    public interface IChartEditorPresenter : IPresenter<IChartEditorView>,
-      IListener<ChartUpdatedEvent>,
-      IListener<ChartPropertiesChangedEvent>
```
msevestre: This is not required in this component. The event is only raise from this presenter so we can call the handle directly

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -230,7 +229,7 @@ public class ChartEditorPresenter : AbstractCommandCollectorPresenter<IChartEdit
          _dataBrowserPresenter.SelectionChanged += (o, e) => onSelectionChanged(e.Columns);
 
          _chartExportSettingsPresenter.ChartExportSettingsChanged += (o,e)=> onChartPropertiesChanged();
-         _chartSettingsPresenter.ChartSettingsChanged += (o, e) => onChartPropertiesChanged();
+         _chartSettingsPresenter.ChartSettingsChanged += (o, e) => updateChart();
```
msevestre: Need to call full update when settings are changing (full update redraws everything). Indeed, sideMarginEnabled updates axis and not only chart

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -281,6 +280,7 @@ private void updateChart()
       private void onChartPropertiesChanged()
       {
          _eventPublisher.PublishEvent(new ChartPropertiesChangedEvent(Chart));
+         ChartChanged();
```
msevestre: calling chartchanged directly

PR #371, opened 2018-03-01
Fixes #370 add qualification plan to classification type

### LINKED ISSUE #370: Add QualificationPlan classification type

msevestre:


msevestre:
Required to create the structure in PKSim. We should think about making it more generic and defining those constants as string or at least provided a way to add app specific folder and classification

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -143,6 +143,7 @@ public static class Captions
       public static readonly string ObservedDataFolder = "Observed Data";
       public static readonly string ParameterIdentificationFolder = "Parameter Identifications";
       public static readonly string SensitivityAnalysisFolder = "Sensitivity Analyses";
+      public static readonly string QualificationPlanFolder = "Qualification Plans";
       public static readonly string LLOQ = "LLOQ";
```
Yuri05: I think we will have only 1 Qualification plan per project. but still an open question so far, so let's keep it

src/OSPSuite.Core/Domain/ClassificationType.cs
```diff
@@ -7,6 +7,7 @@ public enum ClassificationType
       Simulation,
       Comparison,
       ParameterIdentification,
-      SensitiviyAnalysis
+      SensitiviyAnalysis,
+      QualificationPlan
```
Yuri05: Typo: Sensitivi**t**yAnalysis

msevestre: crap!

msevestre: well this is old here. I do not want to change it because this would require project conversion

PR #376, opened 2018-04-05
Fixes #295 add option to disable circular reference check

### LINKED ISSUE #295: Add an option to disable Circular Reference Check
This should be enabled by default but some Models are so complicated, that the current implementation blows off the memory and also takes quite some time. (@PavelBal knows what I am talking about)

msevestre:


msevestre:
@Yuri05 PLease review

PR #387, opened 2018-04-19
Fixes #386 Favorite parameters that should not be variable should be …

### LINKED ISSUE #386: Favorite parameters that should not be variable should be ignored in PI when creating a PI automatically
required for https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/316

msevestre:
…ignored in PI when creating a PI automatically

### DIFF
--- a/src/OSPSuite.Presentation/Services/ParameterIdentificationTask.cs
+++ b/src/OSPSuite.Presentation/Services/ParameterIdentificationTask.cs
@@ -2,7 +2,6 @@
 using System.Collections.Generic;
 using System.Linq;
 using OSPSuite.Assets;
-using OSPSuite.Utility.Extensions;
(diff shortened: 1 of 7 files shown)


src/OSPSuite.Presentation/Services/ParameterIdentificationTask.cs
```diff
@@ -146,6 +160,9 @@ public void AddSimulationsTo(ParameterIdentification parameterIdentification, IE
 
       public void AddParameterTo(ParameterIdentification parameterIdentification, IParameter parameter)
       {
+         if (!_parameterSelector.CanUseParameter(parameter))
```
msevestre: The real only change. Parameter can only be added to PI if it can actually be used in a PI :)

PR #394, opened 2018-04-24
Fixes #393 PSV und MSV difference. Instead of Absent and present the …

### LINKED ISSUE #393: PSV und MSV difference. Instead of Absent and present the value of the present value should be shown

msevestre:
…value of the present value should be shown

This ensures that I can display more info on difference. Implemented at the moment only for PSV and MSV
image

(showing diff in value using value in display unit or formula with formula string)

### DIFF
--- a/src/OSPSuite.Core/Comparison/BuildingBlockDiffBuilder.cs
+++ b/src/OSPSuite.Core/Comparison/BuildingBlockDiffBuilder.cs
@@ -9,6 +9,7 @@ namespace OSPSuite.Core.Comparison
       private readonly ObjectBaseDiffBuilder _objectBaseDiffBuilder;
       private readonly EnumerableComparer _enumerableComparer;
       private readonly Func<T, object> _equalityProperty;
+      protected Func<T, string> _presentObjectDetailsFunc;
 
       protected BuildingBlockDiffBuilder(ObjectBaseDiffBuilder objectBaseDiffBuilder,
          EnumerableComparer enumerableComparer, Func<T, object> equalityProperty)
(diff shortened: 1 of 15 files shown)


src/OSPSuite.Core/Comparison/BuildingBlockDiffBuilder.cs
```diff
@@ -76,18 +77,40 @@ public ObserverBuildingBlocksDiffBuilder(ObjectBaseDiffBuilder objectBaseDiffBui
...
+      private readonly UnitFormatter _unitFormatter = new UnitFormatter();
+
+      protected StartValueBuildingBlockDiffBuilder(ObjectBaseDiffBuilder objectBaseDiffBuilder, EnumerableComparer enumerableComparer) :
+         base(objectBaseDiffBuilder, enumerableComparer, x => x.Path)
+      {
+         _presentObjectDetailsFunc = startValueDisplayFor;
+      }
+
+      private string startValueDisplayFor(TStartValue startValue)
+      {
```
msevestre: This is the method to work on displaying the difference for missing objects

src/OSPSuite.Core/Domain/ObjectBase.cs
```diff
@@ -49,15 +49,8 @@ protected ObjectBase(string id)
...
-         set
-         {
-            if (string.Equals(_name, value))
-               return;
-
-            _name = value;
-            OnPropertyChanged(() => Name);
-         }
+         get => _name;
+         set => SetProperty(ref _name, value);
```
msevestre: We have a shortcut method for that. Just using it

src/OSPSuite.Presentation/Mappers/DiffItemToDiffItemDTOMapper.cs
```diff
@@ -64,16 +64,24 @@ public void Visit(MissingDiffItem missingDiffItem)
          if (missingDiffItem.MissingObject2 != null)
          {
             value1 = Captions.Comparisons.Absent;
-            value2 = Captions.Comparisons.Present;
+            value2 = presentWithDetails(missingDiffItem.PresentObjectDetails);
          }
          else if (missingDiffItem.MissingObject1 != null)
          {
-            value1 = Captions.Comparisons.Present;
+            value1 = presentWithDetails(missingDiffItem.PresentObjectDetails);
```
msevestre: Add details if defined in Diff object item

PR #398, opened 2018-04-26
Fixes #397 Simulation export to json: Do not save path of simulation

### LINKED ISSUE #397: Simulation export to json: Do not save path of simulation

msevestre:


msevestre:
@Yuri05 FYI: I was exporting the name of all simulation before in the path. Just redundant information at that stage because we have the name of the simulation from the file naem

PR #403, opened 2018-05-03
 Fixes #402 full version info 

### LINKED ISSUE #402: support full version info in configuration

msevestre:


src/OSPSuite.Infrastructure/Configuration/OSPSuiteConfiguration.cs
```diff
@@ -42,19 +42,21 @@ public abstract class OSPSuiteConfiguration : IApplicationConfiguration
...
       public abstract string ApplicationFolderPathName { get; }
       public virtual string LicenseAgreementFilePath { get; } = Constants.Files.LICENSE_AGREEMENT_FILE_NAME;
+      private readonly bool _isReleasedVersion;
 
       protected OSPSuiteConfiguration()
       {
          var assemblyVersion = AssemblyVersion;
          MajorVersion = version(assemblyVersion.Minor);
          Version = $"{MajorVersion}.{assemblyVersion.Build}";
          ReleaseDescription = retrieveReleaseDescription();
```
msevestre: Released Description used to be empty in release and EAP , alpha beta otherwise.
Now this will changed. It will be something like 7.2.3.4 in release and 7.2.3.4 EAP in EAP

The reason is that many tools read the value of AssemblyInformatiomal Attribute and if it's there, use it instead of the Version. That's why the version shown in the CLI is EAP or empty

src/OSPSuite.Infrastructure/Configuration/OSPSuiteConfiguration.cs
```diff
@@ -65,21 +67,19 @@ protected OSPSuiteConfiguration()
...
+         return _isReleasedVersion ? $"{ProductNameWithTrademark} {MajorVersion}" : $"{ProductNameWithTrademark} {ReleaseDescription}";
       }
 
       private string fullVersionFrom(int revision)
       {
-         if (string.IsNullOrEmpty(ReleaseDescription))
-            return $"{Version} - Build {revision}";
-
-         return $"{Version}.{revision} - {ReleaseDescription}";
+         return _isReleasedVersion ? $"{Version} - Build {revision}" : ReleaseDescription;
```
msevestre: Same as before, for released, otherwise whatever is in ReleaseDescription

src/OSPSuite.Infrastructure/Configuration/OSPSuiteConfiguration.cs
```diff
@@ -65,21 +67,19 @@ protected OSPSuiteConfiguration()
...
-         var productDisplayName = $"{ProductNameWithTrademark} {MajorVersion}";
-         if (string.IsNullOrEmpty(ReleaseDescription))
-            return productDisplayName;
+         return string.Equals($"{version}.{revision}", releaseDescription);
+      }
 
-         return $"{productDisplayName} - {ReleaseDescription}";
+      private string retrieveProductDisplayName()
+      {
+         return _isReleasedVersion ? $"{ProductNameWithTrademark} {MajorVersion}" : $"{ProductNameWithTrademark} {ReleaseDescription}";
```
msevestre: We add the name in the header plus either X.Y or whateever is in Releas eDescription

Issue #406, opened 2018-05-05
Value origin: Method ordering


Yuri05:
In Value origin/Method: put **Other** before **Unknown**
(to be consistent with Value origin/Source)

msevestre:
ALso going to move In vivo and in vitro above pi so that we have values orderes alphabetically

PR #410, opened 2018-05-07
Fixes #683 typi and #406 ordering

msevestre:


msevestre:
Also updated SmartXls. Please review

PR #418, opened 2018-05-11
Fixes #416 visible option in chart manager

### LINKED ISSUE #416: Visible column is not available when managing chart templates

msevestre:


src/OSPSuite.Presentation/DTO/Charts/CurveTemplateDTO.cs
```diff
@@ -70,6 +70,12 @@ public Color Color
          set => CurveTemplate.CurveOptions.Color = value;
       }
 
+      public bool Visible
```
msevestre: We had forgotten that one I guess

src/OSPSuite.UI/Binders/ValueOriginBinder.cs
```diff
@@ -120,7 +120,7 @@ private void updateValueOrigin(bool canceled)
          _gridView.RowCellStyle += updateRowCellStyle;
          _gridView.CustomDrawCell += customDrawCell;
 
-         _valueOriginColumn = _gridViewBinder.AutoBind(x => x.ValueOrigin)
+         _valueOriginColumn = _gridViewBinder.Bind(x => x.ValueOrigin)
```
msevestre: this is not required and does not solve the problem mentioned

PR #419, opened 2018-05-11
Fixes #417 many data points and scrolling

### LINKED ISSUE #417: Chart editor - many data points and scrolling
see https://github.com/Open-Systems-Pharmacology/MoBi/issues/268

msevestre:


src/OSPSuite.Presentation/Presenters/Charts/CurveSettingsPresenter.cs
```diff
@@ -151,7 +157,9 @@ public void Refresh()
 
       public void NotifyCurvePropertyChange(CurveDTO curveDTO)
       {
-         CurvePropertyChanged(this, new CurveEventArgs(curveDTO.Curve));
+         this.DoWithinLatch(() =>
```
msevestre: Notify curve changed (eg. color changed or visible )) would trigger a rebind which is not required

msevestre:
@Yuri05 I think this is a great change. Please review

PR #429, opened 2018-05-18
Fixes #428 parameter debug improvement. Fixes #427 Update SImModel

### LINKED ISSUE #427: Update to newest SimModel
Update to newest SimModel

### LINKED ISSUE #428: Parameter report creator should also export building block type and allow for caller configuration of table and row

msevestre:


msevestre:
@Yuri05 Merging
Update simmodel and ensure that caller can add column to the debug view. Because the export is used probably for some kind of reporting, I am not going to add the BB Column there but PKSim will use it

PR #445, opened 2018-08-01
Fixes #444 ensure that observed data meta data is updated

### LINKED ISSUE #444: Ensure that ObservedData MetaDataContent is not duplicated
see https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/904

msevestre:


src/OSPSuite.Infrastructure/Serialization/ORM/MetaData/ObservedDataMetaData.cs
```diff
@@ -2,13 +2,16 @@
...
-   public class ObservedDataMetaData: MetaData<string>, IUpdatableFrom<ObservedDataMetaData>
+   public class ObservedDataMetaData : MetaData<string>, IUpdatableFrom<ObservedDataMetaData>
    {
-       public virtual DataRepositoryMetaData DataRepository { get; set; }
+      public virtual DataRepositoryMetaData DataRepository { get; set; }
 
       public void UpdateFrom(ObservedDataMetaData source, ISession session)
       {
-         DataRepository = source.DataRepository;
+         if (DataRepository == null)
```
msevestre: The problem was that the reference was always updated, thus creating a new MetaDataContent object every time!!
(source being the object created on save that should be used to update the existing objects instead of creating new ones).

msevestre:
@Yuri05 Merging: Please review

PR #447, opened 2018-09-23
Proposal to update some of the package that were outdated. I'd like t…

msevestre:
…o ensure that we keep uptodate


msevestre:
Little bit of a syntax change with fakeiteasy. Rest is good otherwise

msevestre:
ok works now. Merging

PR #451, opened 2018-10-11
Fixes Open-Systems-Pharmacology/PK-Sim#899 Simulation Chart Editor: Changing the Layout not working

ju-rgen:
should work now.


msevestre:
@ju-rgen Even though the build is faukung because of the SmartXLS component, I am going to merge it and we can go from there. Thanks @ju-rgen

PR #465, opened 2018-11-12
Remove chm references and use docs.open-systems-pharmacology.org instead

### LINKED ISSUE #464: Remove chm from setup, add pdf instead. Open docs on F1

msevestre:
Fixes #464


src/OSPSuite.Presentation/Presenters/ContextMenus/CommonMenuBarButtons.cs
```diff
@@ -69,5 +70,14 @@ public static IMenuBarItem ClearHistory(MenuBarItemId menuBarItemId)
             .WithCommand<ClearHistoryUICommand>()
             .WithIcon(ApplicationIcons.ClearHistory);
       }
+
+      public static IMenuBarItem Help(MenuBarItemId menuBarItemId)
```
msevestre: This command is also a generic one and should be used by MoBi  and PKSim

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationPredictedVsObservedChartPresenter.cs
```diff
@@ -27,7 +27,6 @@ public class ParameterIdentificationPredictedVsObservedChartPresenter : Paramete
          base(view, chartPresenterContext, ApplicationIcons.PredictedVsObservedAnalysis, PresenterConstants.PresenterKeys.ParameterIdentificationPredictedVsActualChartPresenter)
       {
          _predictedVsObservedChartService = predictedVsObservedChartService;
-         view.HelpId = HelpId.Tool_PI_Analysis_PredictedVsObserved;
```
msevestre: nothing like that anymore

src/OSPSuite.UI/Views/Comparisons/ComparisonView.cs
```diff
@@ -111,7 +110,5 @@ public override void InitializeResources()
          lblNoDifference.Font = new Font(lblNoDifference.Font.FontFamily, 14);
          lblNoDifference.Appearance.TextOptions.HAlignment = HorzAlignment.Center;
       }
-
-      protected override int TopicId => HelpId.Tool_ComparisonBuildingBlocks;
```
msevestre: in theory we could pass sub url but then as soon as we change the site, the url is obsolete and we get a page not found error
image

I'd rather always land on the landing page

Yuri05: Let's keep it for 7.4 like this.

For 7.5 I would suggest context-specific navigation to the corresponding help page.
In order to avoid "page not found" problems, we could create a dictionary `Help topic <=> Docu URL`. Then checking URLs before every new release would prevent such a problem.

Additionally the dictionary could be placed also on GitHub (e.g. in the _docs_ repo) and adjusted also **between releases**. Then PK-Sim/MoBi would try to load the dictionary from GitHub first. If failed, use the version provided with the setup.
This would minimize the risk of missing help pages.

msevestre: No idea how one would go with implementing that (dynamically updating constant). But sounds good. I  still see the obvious that this page will go out of sync very quickly and thus result in out of date links.
Anyways food for thoughts for when we have nothing else to do

msevestre: actually it could probably be done differently, **independent from URL**
https://docs.open-systems-pharmacology.org/?q=Lipophilicity

You can pass a query parameter to the url... That way you could find the relevant topics right away

Yuri05: I would be in favor of pointing to the URL instead of the search...

msevestre: I wouldn't :)

PR #467, opened 2018-11-14
Fixes #466 base grid comma

### LINKED ISSUE #466: Entering a basegrid value using comma leads to a wrong message in some cases

msevestre:


src/OSPSuite.Presentation/Presenters/ObservedData/DataRepositoryDataPresenter.cs
```diff
@@ -90,9 +90,7 @@ public void AddData(int rowIndex)
       /// <returns></returns>
       public IEnumerable<string> GetCellValidationErrorMessages(int rowIndex, int columnIndex, string newValue)
       {
-         if (!float.TryParse(newValue, out var proposedValue))
```
msevestre: This is not required because the method throws an exception of the value is not a float. 
Moreover, this cannot happen because we have a formatting in place

tests/OSPSuite.Core.Tests/Presentation/DataRepositoryDataPresenterSpecs.cs
```diff
@@ -145,23 +145,6 @@ public void should_not_return_validation_messages()
...
    }
 
-   public class When_getting_validation_messages_for_invalid_value : concern_for_DataRepositoryDataPresenter
-   {
-      private List<string> _result;
-
-      protected override void Because()
-      {
-         sut.EditObservedData(_dataRepository);
-         _result = sut.GetCellValidationErrorMessages(0, 0, "blah").ToList();
```
msevestre: Cannot happen anymore

PR #500, opened 2019-02-28
Fixes #499 add calculation method diff builder

### LINKED ISSUE #499: Add calculation method diff builder
required for https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1027

msevestre:


src/OSPSuite.Core/Comparison/CalculationMethodCacheDiffBuilder.cs
```diff
@@ -0,0 +1,20 @@
...
+      private readonly EnumerableComparer _enumerableComparer;
+
+      public CalculationMethodCacheDiffBuilder(EnumerableComparer enumerableComparer)
+      {
+         _enumerableComparer = enumerableComparer;
+      }
+
+      public override void Compare(IComparison<CalculationMethodCache> comparison)
+      {
+         _enumerableComparer.CompareEnumerables(comparison, x => x, item => item.Category, missingItemType: Captions.CalculationMethod);
```
msevestre: Identity is per category as we should have on entry per category. Category missing will be displayed as present absent

src/OSPSuite.Core/Comparison/CalculationMethodDiffBuilder.cs
```diff
@@ -0,0 +1,14 @@
+﻿using OSPSuite.Core.Domain;
+
+namespace OSPSuite.Core.Comparison
+{
+   public class CalculationMethodDiffBuilder : DiffBuilder<CalculationMethod>
+   {
+      public override void Compare(IComparison<CalculationMethod> comparison)
+      {
+         //No need to compare name and display name as they will be correlated
+         CompareStringValues(x => x.DisplayName, x => x.DisplayName, comparison);
```
msevestre: I have ensured that DisplayName is never empty so we can really compare only display name in this case. Category is going to give the context

PR #502, opened 2019-03-01
Fixes #494 multiple observed data

### LINKED ISSUE #494: PI: allow multiple mapping of observed data sets
Currently an observed data set can be maped to max. 1 simulated curve.
Trying to map it to a second curve results in the error message below.

However there are plenty of situations where this kind of multimapping is required.
grafik

@sfrechen @twendl FYI

msevestre:


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1184,10 +1184,9 @@ public static string SensitivityParameterIsInvalid(string name)
       public static readonly string CovarianceMatrixCannotBeCalculated = "Covariance matrix cannot be calculated";
       public static readonly string CorrelationMatrixCannotBeCalculated = "Correlation matrix cannot be calculated";
 
-      public static string CannotSelectTheSamePartialProcessMoreThanOnce(string name)
-      {
-         return $"'{name}' cannot be selected more than once.";
-      }
+      public static string CannotSelectTheSamePartialProcessMoreThanOnce(string name) => $"'{name}' cannot be selected more than once.";
+
+      public static string CannotSelectTheObservedDataMoreThanOnce(string name) => $"'{name}' cannot be selected more than once for the same output.";
```
msevestre: We still have a restriction. Same output, same observed data. That does not make sense

src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -256,24 +257,20 @@ public static class Filter
          public static readonly string MATLAB_FILTER = FileFilter("Matlab�", MATLAB_EXTENSION);
          public static readonly string CSV_FILE_FILTER = FileFilter("CSV", CSV_EXTENSION);
          public static readonly string JSON_FILE_FILTER = FileFilter("Json", JSON_EXTENSION);
+         public static readonly string MARKDOWN_FILE_FILTER = FileFilter("Markdown", MARKDOWN_EXTENSION);
```
msevestre: I have added the markdown extension as well. Not required for this PR but I will need it in PKSim and Qualif runner

src/OSPSuite.Core/Domain/ParameterIdentifications/WeightedObservedData.cs
```diff
@@ -24,8 +24,17 @@ public WeightedObservedData(DataRepository observedData)
       public static implicit operator DataRepository(WeightedObservedData weightedObservedData) => weightedObservedData?.ObservedData;
 
       public string Name => ObservedData?.Name;
+
+      /// <summary>
+      ///    Identifier used to uniquely identify a weighted observed data. The same observed data could be used more than
+      ///    once in a mapping
+      /// </summary>
+      public int? Id { get; set; }
```
msevestre: That's my trick for display name unicity. An Id that is set everytime multiple instances of the same observed data is used
Default is null so that we are compatible with previous PR
if we rename the obs data, we do not need to do anythig

src/OSPSuite.Presentation/DTO/ParameterIdentifications/OutputMappingDTO.cs
```diff
@@ -68,44 +64,26 @@ public Scalings Scaling
...
-         {
-            get
-            {
-               return CreateRule.For<OutputMappingDTO>()
-                  .Property(x => x.ObservedData)
-                  .WithRule((x, y) => x.Mapping.DimensionsAreConsistent(y))
-                  .WithError(Error.OutputMappingHasInconsistentDimension);
-            }
-         }
+         private static readonly IBusinessRule _outputMatchesObservationDimension = CreateRule.For<OutputMappingDTO>()
```
msevestre: some code cleanup here

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationOutputMappingPresenter.cs
```diff
@@ -154,27 +159,57 @@ private IEnumerable<DataRepository> allPossibleObservedDataForOutput(SimulationQ
...
-            .OrderBy(x => x.Name);        
+            .OrderBy(x => x.Name);
       }
 
       public void ObservedDataSelectionChanged(OutputMappingDTO dto, DataRepository newObservedData, DataRepository oldObservedData)
       {
-         if (observedDataAlreadySelected(newObservedData))
+         var allOutputsUsingObservedData = _allOutputMappingDTOs.Where(x => Equals(x.ObservedData, newObservedData)).Except(new[] {dto}).ToList();
+
+         if (observedDataAlreadySelectedForSameOutput(dto.Output, newObservedData))
```
msevestre: selected for the same output and obs data

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationOutputMappingPresenter.cs
```diff
@@ -154,27 +159,57 @@ private IEnumerable<DataRepository> allPossibleObservedDataForOutput(SimulationQ
...
          {
             dto.ObservedData = oldObservedData;
             View.CloseEditor();
             throw new CannotSelectTheObservedDataMoreThanOnceException(newObservedData);
          }
 
+         var weightedObservedData = new WeightedObservedData(newObservedData)
+         {
+            Id = nextUniqueIdFor(allOutputsUsingObservedData)
```
msevestre: this is where I am created the next Id. As soon as the selection is performed

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/WeightedObservedDataFormatter.cs
```diff
@@ -0,0 +1,22 @@
...
+
+      public WeightedObservedDataFormatter(OutputMappingDTO outputMappingDTO)
+      {
+         _outputMappingDTO = outputMappingDTO;
+      }
+
+      public string Format(DataRepository observedData)
+      {
+         //Input value is not used as the observedData is already stored in the WeightedObservedDate. However the display needs to be adjusted
+         return _outputMappingDTO?.WeightedObservedData?.DisplayName;
```
msevestre: This ensures that we actually have the right display in the mapping view as well

msevestre:
@Yuri05 This is how it looks like
image


Maybe we should use (id) instead of - Id. Not sure. Somehow we need to be able to differentiate between same usage of obs data so that one can adjust weights

PR #520, opened 2019-03-26
Fixes #519 simmodel export

### LINKED ISSUE #519: SimModelExporter implementation: Export of SimModel representation is NOT thread safe

msevestre:


src/OSPSuite.Core/Serialization/SimModel/Services/SimulationExportCreator.cs
```diff
@@ -14,7 +14,7 @@
 
 namespace OSPSuite.Core.Serialization.SimModel.Services
 {
-   public interface ICreateExportModelVisitor
+   public interface ISimulationExportCreator
```
msevestre: I have renamed this guy because the fact this is is a visitor is an implementation detail

src/OSPSuite.Core/Serialization/SimModel/Services/SimulationExportCreator.cs
```diff
@@ -27,14 +27,14 @@ public interface ICreateExportModelVisitor
       SimulationExport CreateExportFor(IModel model, SimModelExportMode exportMode);
    }
 
-   public class CreateExportModelVisitor :
+   public class SimulationExportCreator :
```
msevestre: This class was not instantiated everytime. When exporting in //, the _idMap would be cleared etc... which would lead to very weird error. Now we always get a new instance=>no more issue

src/OSPSuite.Engine/Serialization/SimModelExporter.cs
```diff
@@ -49,7 +49,8 @@ private MatlabCodeWriteMode writeModeFrom(MatlabFormulaExportMode formulaExportM
 
       private SimulationExport createSimulationExport(IModelCoreSimulation simulation, SimModelExportMode simModelExportMode)
       {
-         var simulationExport = _createExportModelVisitor.CreateExportFor(simulation.Model, simModelExportMode);
+         var simulationExportCreator = _simulationExportCreatorFactory.Create();
```
msevestre: Here new instance created everytime instead of having a reference to a single export

tests/OSPSuite.Core.Tests/Core/SimModelExporterIntegrationTests.cs
```diff
@@ -40,9 +39,8 @@ protected override void Context()
          _objectPathFactory = IoC.Resolve<IObjectPathFactory>();
          var solverSettingsFactory = IoC.Resolve<ISolverSettingsFactory>();
          _solverSettings = solverSettingsFactory.CreateCVODE();
-         _tableFormulaToTableFormulaExportMapper = IoC.Resolve<ITableFormulaToTableFormulaExportMapper>();
-         _concentrationBasedFormulaUpdater = IoC.Resolve<IConcentrationBasedFormulaUpdater>();
-         sut = new SimModelExporter(new CreateExportModelVisitor(_objectPathFactory, _tableFormulaToTableFormulaExportMapper,_concentrationBasedFormulaUpdater), new ExportSerializer(_simModelXmlSerializerRepository));
+         _simulationExporterCreatorFactory = IoC.Resolve<ISimulationExportCreatorFactory>();
+         sut = new SimModelExporter(_simulationExporterCreatorFactory, new ExportSerializer(_simModelXmlSerializerRepository));
```
msevestre: I am not sure why the object is created like that but I don't have time to investigate all the weirdness as long as it works

PR #535, opened 2019-04-12
Fixes #534 In container criteria 

### LINKED ISSUE #534: Enable Sum Formula tag "within container"
For calculating sum formulas it would be handy to have an option to restrict summation to the current container where the sum parameter is located. So, in addition to have the condition: "tagged with <container name>" where the container name is explicitly given, it would be "tagged with local container".

This should select container parameters as well as local molecule and reaction parameters that are present in the local container.

Implementing this feature would improve and simplify maintenance of the blood coagulation simulator, probably also other things. This is my most favored request since there is no workaround, and manual maintenance of the formulas is error-prone.

msevestre:
I think I have it. Container criteria in general

### DIFF
--- a/src/OSPSuite.Core/Domain/Descriptors/DescriptorCriteria.cs
+++ b/src/OSPSuite.Core/Domain/Descriptors/DescriptorCriteria.cs
@@ -12,14 +12,14 @@ namespace OSPSuite.Core.Domain.Descriptors
          return this.ToString(Constants.AND.ToUpper(), " ").Trim(' ');
       }
 
-      public bool IsSatisfiedBy(Tags tags)
+      public bool IsSatisfiedBy(EntityDescriptor entityDescriptor)
       {
-         return this.Any() && this.All(condition => condition.IsSatisfiedBy(tags));
+         return this.Any() && this.All(condition => condition.IsSatisfiedBy(entityDescriptor));
       }
 
       public virtual bool IsSatisfiedBy(IEntity entity)
       {
-         return IsSatisfiedBy(new Tags(entity.Tags) {new Tag(entity.Name)});
+         return IsSatisfiedBy(new EntityDescriptor(entity));
       }
 
       public bool Equals(DescriptorCriteria other)
@@ -49,14 +49,14 @@ namespace OSPSuite.Core.Domain.Descriptors
       /// <param name="tag">Tag to remove</param>
       public void RemoveByTag<T>(string tag) where T : class, ITagCondition
       {
-         var conditionsToremove = (from conditions in this
+         var conditionsToRemove = (from conditions in this
                                    let tagConditions = conditions as T
                                    where tagConditions != null
                                    where tagConditions.Tag.Equals(tag)
                                    select conditions)
             .ToList();
 
-         conditionsToremove.Each(condition => Remove(condition));
+         conditionsToRemove.Each(condition => Remove(condition));
       }
 
       public DescriptorCriteria Clone()
--- a/src/OSPSuite.Core/Domain/Descriptors/EntityDescriptor.cs
+++ b/src/OSPSuite.Core/Domain/Descriptors/EntityDescriptor.cs
@@ -1,0 +1,18 @@
+﻿namespace OSPSuite.Core.Domain.Descriptors
+{
+   public class EntityDescriptor
+   {
+      public Tags Tags { get; set; }
+      public Tag Container { get; set; }
+
+      public EntityDescriptor()
+      {
+      }
+
+      public EntityDescriptor(IEntity entity)
+      {
+         Tags = new Tags(entity.Tags){new Tag(entity.Name)};
+         Container = new Tag(entity.ParentContainer?.Name);
+      }
+   }
+}
--- a/src/OSPSuite.Core/Domain/Descriptors/MatchTagCondition.cs
+++ b/src/OSPSuite.Core/Domain/Descriptors/MatchTagCondition.cs
@@ -6,24 +6,23 @@ namespace OSPSuite.Core.Domain.Descriptors
    {
       public string Tag { get; private set; }
 
+      public string Condition => Tag;
+
       [Obsolete("For serialization")]
       public MatchTagCondition()
       {
       }
 
-      public override string ToString()
+      public override string ToString() => Condition;
+
+      public MatchTagCondition(string tag)
       {
-         return Tag;
+         Tag = tag;
       }
 
-      public MatchTagCondition(string matchTag)
+      public bool IsSatisfiedBy(EntityDescriptor entityDescriptor)
       {
-         Tag = matchTag;
-      }
-
-      public bool IsSatisfiedBy(Tags tags)
-      {
-         return tags.Contains(Tag);
+         return entityDescriptor.Tags.Contains(Tag);
       }
 
       public IDescriptorCondition CloneCondition()
@@ -33,7 +32,7 @@ namespace OSPSuite.Core.Domain.Descriptors
 
       public void Replace(string keyword, string replacement)
       {
-         if(string.Equals(Tag,keyword))
+         if (string.Equals(Tag, keyword))
             Tag = replacement;
       }
 
@@ -47,16 +46,9 @@ namespace OSPSuite.Core.Domain.Descriptors
       public bool Equals(MatchTagCondition other)
       {
          if (other == null) return false;
-         return string.Equals(Tag,other.Tag);
+         return string.Equals(Tag, other.Tag);
       }
 
-      public override int GetHashCode()
-      {
-         //Match and NoMatch CANNOT return the same hash for the same tag
-         int hash = 29; //one prime number that should differ from the one in Not Match
-         if (Tag != null)
-            hash = hash + Tag.GetHashCode();
-         return hash;
-      }
+      public override int GetHashCode() => Condition?.GetHashCode() ?? 0;
    }
 }
--- a/src/OSPSuite.Core/Domain/Services/ModelReportCreator.cs
+++ b/src/OSPSuite.Core/Domain/Services/ModelReportCreator.cs
@@ -2,10 +2,9 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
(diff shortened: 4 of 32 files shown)


src/OSPSuite.Core/Domain/Descriptors/DescriptorCriteria.cs
```diff
@@ -12,14 +12,14 @@ public override string ToString()
...
+      public bool IsSatisfiedBy(EntityDescriptor entityDescriptor)
       {
-         return this.Any() && this.All(condition => condition.IsSatisfiedBy(tags));
+         return this.Any() && this.All(condition => condition.IsSatisfiedBy(entityDescriptor));
       }
 
       public virtual bool IsSatisfiedBy(IEntity entity)
       {
-         return IsSatisfiedBy(new Tags(entity.Tags) {new Tag(entity.Name)});
+         return IsSatisfiedBy(new EntityDescriptor(entity));
```
msevestre: Not all tags are the same. there is now a hiearchy of tags and you cannot just check everything

src/OSPSuite.Core/Domain/Descriptors/EntityDescriptor.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace OSPSuite.Core.Domain.Descriptors
+{
+   public class EntityDescriptor
+   {
+      public Tags Tags { get; set; }
```
msevestre: This could be extended to be much more powerful (for example entity type if we want to filter by type etc..)

src/OSPSuite.Core/Domain/Descriptors/MatchTagCondition.cs
```diff
@@ -47,16 +46,9 @@ public override bool Equals(object otherObject)
...
       {
          if (other == null) return false;
-         return string.Equals(Tag,other.Tag);
+         return string.Equals(Tag, other.Tag);
       }
 
-      public override int GetHashCode()
-      {
-         //Match and NoMatch CANNOT return the same hash for the same tag
-         int hash = 29; //one prime number that should differ from the one in Not Match
```
msevestre: THis was a bit cumbersome all of that. My goal years ago was to ensure that MatchTag "XX" and NotMatchTag "XX" where not seen as equal
Since they have a different display, we should  use that instead

src/OSPSuite.Core/Domain/Services/ModelReportCreator.cs
```diff
@@ -430,39 +431,8 @@ private void createEntityReport<T>(string separatorName, Action<T> reportFor)
       private IEnumerable<T> all<T>()
       {
          return from c in _allChildren
-                where c.IsAnImplementationOf<T>()
-                select c.DowncastTo<T>();
-      }
-
-      private void reportFor(IEnumerable<IDescriptorCondition> descriptorCriteria)
```
msevestre: this was not used anywhere. Gone

tests/OSPSuite.Core.Tests/Core/ModelConstructorIntegrationTests.cs
```diff
@@ -216,16 +216,27 @@ public void Observer_ContainerObs_1_should_have_been_created_under_the_molecule_
          moleculeContainer.ContainsName(observerName).ShouldBeTrue();
       }
 
+
+      [Observation]
+      public void Observer_InContainerObserver_should_have_been_created_under_the_molecule_properties_container_for_C_under_organism()
+      {
+         string observerName = "InContainerObserver";
+
+         var liver_plasma = _model.ModelOrganCompartment(ConstantsForSpecs.Lung, ConstantsForSpecs.Plasma);
```
msevestre: I have created an observer with criteria in Plasma with Parent Lung.... works :)

tests/OSPSuite.Core.Tests/Helpers/ModelHelperForSpecs.cs
```diff
@@ -273,6 +273,15 @@ private IObserverBuildingBlock getObservers()
          fractionObserver.ContainerCriteria = Create.Criteria(x => x.With(ConstantsForSpecs.Plasma));
          observers.Add(fractionObserver);
 
+
+         var observerInOrganism = _objectBaseFactory.Create<IContainerObserverBuilder>().WithName("InContainerObserver");
+         observerInOrganism.Dimension = _dimensionFactory.Dimension(Constants.Dimension.AMOUNT);
+         observerInOrganism.MoleculeList.ForAll = false;
+         observerInOrganism.AddMoleculeName("C");
+         observerInOrganism.Formula = InContainerObs(observers.FormulaCache);
+         observerInOrganism.ContainerCriteria = Create.Criteria(x => x.With(ConstantsForSpecs.Plasma).And.InContainer(ConstantsForSpecs.Lung));
```
msevestre: That's the criteria

src/OSPSuite.Core/Domain/Descriptors/EntityDescriptor.cs
```diff
@@ -0,0 +1,18 @@
...
+      public Tag Container { get; set; }
+
+      public EntityDescriptor()
+      {
+      }
+
+      public EntityDescriptor(IEntity entity)
+      {
+         Tags = new Tags(entity.Tags){new Tag(entity.Name)};
+         Container = new Tag(entity.ParentContainer?.Name);
```
Yuri05: :exclamation: I don't think that this implementation is correct. :exclamation: 

_InContainer-condition_ should check not only the _name_ but also the **_tags_** of the parent container.
(actually _only the tags_, because container-name is always automatically added into the tag-list of every container).

E.g. if a user would like to sum up something related to a group of containers, condition would rely not on the organ name, but on the corresponding tag (total volume of several "tumor" organs for example or whatever else).

src/OSPSuite.Core/Domain/Descriptors/EntityDescriptor.cs
```diff
@@ -0,0 +1,18 @@
...
+      public Tags Tags { get; set; }
+      public Tag Container { get; set; }
+
+      public EntityDescriptor()
+      {
+      }
+
+      public EntityDescriptor(IEntity entity)
+      {
+         Tags = new Tags(entity.Tags){new Tag(entity.Name)};
```
Yuri05: Should we name the property `PArentContainer `instead of just `Container`?

src/OSPSuite.Core/Domain/Descriptors/EntityDescriptor.cs
Yuri05: Actually I'm just asking myself if `EntityDescriptor.Container` should include the tags of **ALL parent containers in the hierarchy of the entity**.

E.g. `EntityDescriptor.Container` of _SmallIntestine/Mucosa/Duodenum/Intracellular/Volume_ would include the tags of 
* SmallIntestine
* SmallIntestine/Mucosa
* SmallIntestine/Mucosa/Duodenum and 
* SmallIntestine/Mucosa/Duodenum/Intracellular

Because that's what is already done in current models quite often: containers become tags of their parent containers in the hierarchy in order to enable e.g. passive transports:
grafikgrafik


If we implement `EntityDescriptor.Container` with the full hierarchy - there will be no need for this kind of "Tag inheritance" in the future!

msevestre: Interesting. I actually thought about both aspects yesterday as I was implementing it

I agree with the first one. However name would have to be added and not just tags (as you can see, it's added here as well for the entity itself)


I thought about #2 a bit and decided against it because if you want to get something in Liver for example, every sub compartment of liver would also be found as potential container. (You are thinking tree up, I am thinking tree down)

I think it would be very easy to suddenly create stuff (transport, observer) or collect stuff (for the sum Formula) that are not intended

Maybe it's not a big deal. We need to think about it a bit more. (Plus those criteria are created all the time that we check if a condition is satisfied...this could potentially be time consuming!)

msevestre: For the last point, we could delay evaluation of such tags until it's really needed. E.g. No need to retrieve the whole hierarchy for each entity if the condition does not even use "in container"

Yuri05:
>I thought about 2 a bit and decided against it because if you want to get something in Liver for example, every sub compartment of liver would also be found as potential container. (You are thinking tree up, I am thinking tree down)

If we introduce not only **InContainer**-condition but also **Not InContainer** (similiar to match/not match), then it would be easy to do this by exclusion of all subcontainers, e.g.
`InContainer(Liver) and Not InContainer(Periportal) and Not InContainer(Pericentral)...`

And in the most cases the music plays in the _lowest_ containers in the hierarchy, where all the molecule amounts are created (Plasma/Interstitial/Intracellular etc.)

Currently with the tags it would be the same: if we would like to create something in liver (but not in subcompartments), we would need the condition
`MatchCondition(Liver) and Not MatchCondition(Periportal) and Not MatchCondition(Pericentral)...`
(because every direct on indirect liver-subcontainer "inherits" the liver-tag)

E.g. this is how it's currently defined for those "special" liver observers (which combines
 liver zones stuff and are define only at the liver level)
(don't know why galbladder is in the list, but that's the idea: "include liver and exclude all direct or indirect subcontainers)
grafik

msevestre:
Yeah that would work.
Do you want me to spend some time on #2 a well? (E.g. add hierarchy) and define a NotInContainer

Yuri05:
>Do you want me to spend some time on #2 a well? (E.g. add hierarchy) and define a NotInContainer

Yes, would make sense.

msevestre:
Sounds good.
To summarize: two new conditions
1. InContainerCondition: Checks the tags of all parent container
2. NotInContainer: similar to previous but for exclusion

Evaluation will be lazy. E.g. construction of tags only performed if the criteria actually requires it.

I'll push a new PR on Monday hopefully

Yuri05:
> 1. InContainerCondition: Checks the tags of all parent container

The tags and the name.

>Evaluation will be lazy. E.g. construction of tags only performed if the criteria actually requires it.

Perfect :)

PR #536, opened 2019-04-15
 Fixes #534 InContainer, NotInContainer criteria

### LINKED ISSUE #534: Enable Sum Formula tag "within container"

msevestre:


src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -513,6 +513,7 @@ public static class ParameterExport
       public static readonly string NOT = "Not";
       public static readonly string AND = "and";
       public static readonly string IN_CONTAINER = "In container";
+      public static readonly string NOT_IN_CONTAINER = "Not in container";
```
msevestre: The next entry... Not in container

src/OSPSuite.Core/Domain/ContainerExtensions.cs
```diff
@@ -51,16 +51,6 @@ public static IEnumerable<IParameter> AllParameters(this IContainer container, F
          return container;
       }
 
-      internal static IEnumerable<IContainer> AllContainersFor(this IEnumerable<IContainer> containers, DescriptorCriteria containerCriteria)
```
msevestre: Those methods were internal and not used. They now go against what I have implemented so I removed them

src/OSPSuite.Core/Domain/Descriptors/EntityDescriptor.cs
```diff
@@ -1,18 +1,45 @@
...
       {
       }
 
-      public EntityDescriptor(IEntity entity)
+      public EntityDescriptor(IEntity entity, Func<IEntity, EntityDescriptor> createDescriptor)
       {
-         Tags = new Tags(entity.Tags){new Tag(entity.Name)};
-         Container = new Tag(entity.ParentContainer?.Name);
+         Tags = new Tags(entity.Tags) {new Tag(entity.Name)};
+         _lazyParentDescriptor = new Lazy<EntityDescriptor>(() => createDescriptor(entity.ParentContainer));
```
msevestre: The Lazy is a way to create an object only when instantiated the first time. Warning: It captures the context (e.g. entity will be referenced here so the object needs to be let go!)

src/OSPSuite.Core/Domain/Descriptors/EntityDescriptor.cs
```diff
@@ -1,18 +1,45 @@
...
+      }
+
+      private static EntityDescriptor createDescriptor(IEntity entity) => entity == null ? null : new EntityDescriptor(entity);
+
+      public Tags ParentContainerTags
```
msevestre: For the caller, this should be transparent/
When ParentContainerTag is retrieved, we actually create the tag if not availalble
Then we also retrieve the hierarchy of tags

src/OSPSuite.Core/Domain/Descriptors/EntityDescriptorMap.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace OSPSuite.Core.Domain.Descriptors
+{
+   public class EntityDescriptorMap<T> where T : IEntity
```
msevestre: Convenience method to have the entity AND the descriptor together. Entity could be saved in the descriptor but I felt this was easier to handle that way

src/OSPSuite.Core/Domain/Descriptors/EntityDescriptorMapList.cs
```diff
@@ -0,0 +1,21 @@
...
+   {
+      public EntityDescriptorMapList()
+      {
+      }
+
+      public EntityDescriptorMapList(IEnumerable<EntityDescriptorMap<T>> collection) : base(collection)
+      {
+      }
+
+      public IReadOnlyList<T> AllSatisfiedBy(DescriptorCriteria criteria)
```
msevestre: The list of such Map. Bonus it returns the entity satisfying a given criteria=> code much simpler

msevestre: The list of such Map. Bonus it returns the entity satisfying a given criteria=> code much simpler

src/OSPSuite.Core/Domain/Descriptors/TagCondition.cs
```diff
@@ -0,0 +1,64 @@
...
+      /// </summary>
+      string Tag { get; }
+
+      /// <summary>
+      ///    Returns the semantic display of the condition for the tag
+      /// </summary>
+      string Condition { get; }
+   }
+
+   public abstract class TagCondition : ITagCondition
```
msevestre: I was fedup of repeating the same code. I created an abstract class

src/OSPSuite.Core/Domain/Formulas/DynamicFormula.cs
```diff
@@ -53,10 +53,10 @@ protected override double CalculateFor(IEnumerable<IObjectReference> usedObjects
       /// <summary>
       ///   Expands the dynamic formula using the list of available object that can be used in the formula
       /// </summary>
-      public IFormula ExpandUsing(IReadOnlyList<IFormulaUsable> allFormulaUsable, IObjectPathFactory objectPathFactory, IObjectBaseFactory objectBaseFactory)
+      public IFormula ExpandUsing(EntityDescriptorMapList<IFormulaUsable> allFormulaUsable, IObjectPathFactory objectPathFactory, IObjectBaseFactory objectBaseFactory)
       {
          var explicitFormula = objectBaseFactory.Create<ExplicitFormula>().WithName(Name).WithFormulaString("0");
-         var allEntityToUse = allFormulaUsable.Where(Criteria.IsSatisfiedBy).ToList();
+         var allEntityToUse = allFormulaUsable.AllSatisfiedBy(Criteria);
```
msevestre: This ensure we do not compute the parent tags if not requried

src/OSPSuite.Core/Domain/Descriptors/NotInContainerCondition.cs
```diff
@@ -0,0 +1,23 @@
...
+
+      public NotInContainerCondition(string tag) : base(tag, Constants.NOT_IN_CONTAINER)
+      {
+      }
+
+      public override IDescriptorCondition CloneCondition() => new NotInContainerCondition(Tag);
+
+      public override bool IsSatisfiedBy(EntityDescriptor entityDescriptor)
+      {
+         return !entityDescriptor.ParentContainerTags.Contains(Tag);
```
msevestre: Only when accessing the ParentContainerTag would the descriptor be computed. Since InContainer and NotInContainer will hardly be used, this should ensure that there is no performance loss

src/OSPSuite.Core/Domain/Services/EventBuilderTask.cs
```diff
@@ -180,9 +187,9 @@ private void addApplicationTransportToModel(ITransportBuilder appTransport, IEnu
          }
       }
 
-      private IEnumerable<IContainer> sourceContainersFor(ITransportBuilder transport, IEnumerable<IContainer> allEventGroupParentChildContainers)
+      private IEnumerable<IContainer> sourceContainersFor(ITransportBuilder transport, EntityDescriptorMapList<IContainer> allEventGroupParentChildContainers)
       {
-         return allEventGroupParentChildContainers.Where(c => transport.SourceCriteria.IsSatisfiedBy(c));
+         return allEventGroupParentChildContainers.AllSatisfiedBy(transport.SourceCriteria);
```
msevestre: it also reads better :)

src/OSPSuite.Core/Domain/Services/ObserverBuilderTask.cs
```diff
@@ -52,7 +57,7 @@ public void CreateObservers(IBuildConfiguration buildConfiguration, IModel model
          }
          finally
          {
-            _allContainers.Clear();
+              _allContainerDescriptors = null;
```
msevestre: That should be enough to remove the reference here. We could clear but I don't think it's necessary

src/OSPSuite.Core/Extensions/EnumerableExtensions.cs
```diff
@@ -13,33 +15,40 @@ public static class EnumerableExtensions
...
-                 select valueRetriever(parameter)).Distinct().ToList();
+            select valueRetriever(parameter)).Distinct().ToList();
       }
 
       public static T MinimumBy<T, TKey>(this IEnumerable<T> theEnumerable, Func<T, TKey> func)
       {
          return theEnumerable.OrderBy(func).First();
       }
+
+      public static EntityDescriptorMapList<T> ToEntityDescriptorMapList<T>(this IEnumerable<T> entities) where T : class, IEntity
```
msevestre: This is really the only place where we create those lists. IT's all transparent for the user

tests/OSPSuite.Core.Tests/Core/EntityDescriptorSpecs.cs
```diff
@@ -0,0 +1,79 @@
...
+            return new EntityDescriptor(entity, CreateDescriptor);
+         }
+
+         sut = new EntityDescriptor(_entity, CreateDescriptor);
+      }
+
+      [Observation]
+      public void should_not_load_the_entity_descriptor_of_the_parent_containers_until_required()
+      {
+         _createDescriptorCount.ShouldBeEqualTo(0);
```
msevestre: Actually testing the lazy loading here.

tests/OSPSuite.Core.Tests/Core/ModelConstructorIntegrationTests.cs
```diff
@@ -218,15 +218,32 @@ public void Observer_ContainerObs_1_should_have_been_created_under_the_molecule_
...
+      public void Observer_NotInContainerObserver_should_have_been_created_under_the_molecule_properties_container_for_C_under_all_plasma_except_lung()
+      {
+         string observerName = "NotInContainerObserver";
+
+         var lung_plasma = _model.ModelOrganCompartment(ConstantsForSpecs.Lung, ConstantsForSpecs.Plasma);
+         var moleculeContainer = lung_plasma.GetSingleChildByName<IContainer>("C");
+         moleculeContainer.ContainsName(observerName).ShouldBeFalse();
+
+
+         var bone_plasma = _model.ModelOrganCompartment(ConstantsForSpecs.Bone, ConstantsForSpecs.Plasma);
```
msevestre: every Plasma except lung.. works :)

PR #540, opened 2019-04-16
Fixes #539 ensure that parsing of sum formula uses iteration pattern

### LINKED ISSUE #539: Ensure that Parsing of SumFormula also uses iteration pattern!

msevestre:


src/OSPSuite.Core/Domain/Formulas/Formula.cs
```diff
@@ -234,20 +233,40 @@ public string FormulaString
...
+      public virtual (bool valid, string validationMessage) IsValid()
+      {
+         return IsValid(FormulaString);
+      }
+
+      public virtual (bool valid, string validationMessage) IsValid(string formulaString)
+      {
+         try
+         {
+            Validate(formulaString);
```
msevestre: This API with the exception is very unnatural. Time to make it more user friendly

tests/OSPSuite.Core.Tests/Core/SumFormulaSpecs.cs
```diff
@@ -128,17 +121,37 @@ protected override void Context()
...
+         valid.ShouldBeTrue(message);
+      }
+   }
+
+   public class When_validating_a_sum_formula_using_the_wrong_iteration_variable : concern_for_SumFormula
+   {
+      [Observation]
+      public void should_return_that_the_formula_is_correct()
+      {
+         var (valid, message) = sut.IsValid("fu * P_#i * f_vas_#i");
```
msevestre: Here, using the wrong variable (P_#i instead of V_#i) fails validation

msevestre:
@Yuri05 Going to merge so that use it in core.

PR #542, opened 2019-04-18
Fixes #477 duplicate name

### LINKED ISSUE #477: PI: dublicate names possible when cloning
1) Create a PI and name it "x"
2) Create a second PI and name it "y".
3) Renaming "x" in "y" is not possible.
4) Cloning "x" and giving it the name "y" is possible, but should not be.

msevestre:


msevestre:
Update nhibernate here as well as part of thic commit

src/OSPSuite.Presentation/DTO/NonEmptyNameDTO.cs
```diff
@@ -43,11 +42,6 @@ protected static class AllRules
             .Property(item => item.Name)
             .WithRule((dto, name) => nameDoesNotContainerIllegalCharacters(name))
             .WithError(Error.NameCannotContainIllegalCharacters(Constants.ILLEGAL_CHARACTERS));
-
-         public static IBusinessRule NameShouldNotBeTheSame { get; } = CreateRule.For<NonEmptyNameDTO>()
```
msevestre: This did not belong in this guy. Rather in the rename guy

src/OSPSuite.Presentation/DTO/ObjectBaseDTO.cs
```diff
@@ -42,35 +44,20 @@ protected virtual bool IsNameUnique(string newName)
...
-
-         return description.StringIsNotEmpty();
+         return !DescriptionRequired || description.StringIsNotEmpty();
       }
 
       private static class ObjectBaseDTORules
       {
-         public static IBusinessRule UniqueName
-         {
-            get
```
msevestre: Create once per object before. Now created once only! this should help especially with the core ObectDTO!

src/OSPSuite.Presentation/DTO/RenameObjectDTOFactory.cs
```diff
@@ -0,0 +1,55 @@
...
+using OSPSuite.Utility.Extensions;
+
+namespace OSPSuite.Presentation.DTO
+{
+   public interface IRenameObjectDTOFactory
+   {
+      RenameObjectDTO CreateFor(IWithName objectBase);
+   }
+
+   public class RenameObjectDTOFactory : IRenameObjectDTOFactory
```
msevestre: That code was mostly in PKSim and some in MoBi. Now we have one place to manae those rules and PKSim can add specific stuff to it

PR #544, opened 2019-04-18
Fixes #543 clone and rename to same different case

### LINKED ISSUE #543: IT's possible to CLONE a PI and set the name to the same in other case

msevestre:


src/OSPSuite.Presentation/DTO/RenameObjectDTO.cs
```diff
@@ -7,6 +7,11 @@ public class RenameObjectDTO : ObjectBaseDTO
    {
       public string OriginalName { get; set; }
 
+      /// <summary>
+      /// Set to <c>true</c> (default), renaming a parameter using another case would be accepted. This is ok for rename scenario but not for clone scenario
+      /// </summary>
+      public bool AllowSameNameAsOriginalInDifferentCase { get; set; } = true;
```
msevestre: As I was testing, I realized it was possible to clone a PI and rename it using another case. It is not allowed for other builing block so this is a bug

msevestre: True: We are in a rename scenario
Clone: We are in a clone scenario

PR #546, opened 2019-04-19
Fixes #545 add path to bb ref

### LINKED ISSUE #545: Add Path to BuildingBlockRef

msevestre:


msevestre:
Actually remove the ref object from core because they don't belong here!

PR #550, opened 2019-04-24
518 infrastructure issue

msevestre:
@Yuri05 This has been bugging me for a very long time. Using core would bring TexReporter
using infrastructure would bring presnetation=>devexpress etc.
This is now all cleaned up. Could be even better but good for now


src/OSPSuite.Presentation/Presenters/Main/IMainViewPresenter.cs
```diff
@@ -13,9 +12,7 @@ public interface IMainViewPresenter :
       IListener<HeavyWorkFinishedEvent>,
       IListener<HeavyWorkStartedEvent>,
       IListener<RollBackStartedEvent>,
-      IListener<RollBackFinishedEvent>,
-      IListener<ReportCreationStartedEvent>,
```
msevestre: Those two events were implemented as abstract from the interface. Remove them. Detailled implementation in all apps

PR #555, opened 2019-04-26
Fixes #554 add missing props

### LINKED ISSUE #554: add NumberOfCompilations to reportconfiguraiton

msevestre:


src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -10,7 +10,7 @@ public static class Constants
    {
       public static readonly int PKML_VERSION = PKMLVersion.Current;
       public const string CVODES = "CVODES";
-      public const int SIM_MODEL_VERSION = 4;
+      public const int SIM_MODEL_VERSION = 5;
```
msevestre: I have also increased that guy. I didn't know about that field but since we have changed simmodel...

Yuri05: where is it used?

msevestre: No idea.. probably export to XML?

msevestre: https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/f8314455502e3c15ae58510f48d161d593b07651/src/OSPSuite.Core/Serialization/SimModel/DTO/SimulationExport.cs#L25

Yuri05: OK, so actually it's Simmodel-**XML** version. Should be only increased when XML-structure changes.

msevestre: ok. I will create a separate issue for that to RENAME the constant and set it to 4

PR #557, opened 2019-04-27
Fixes #556 simmodel version. Also add missing boolean

### LINKED ISSUE #556: Rename SIM_MODEL_VERSION to SIM_MODEL_XML_VERSION and set value to previous value (4)

msevestre:


src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -178,7 +176,8 @@ public static string ParameterCompoundType(int index)
             USE_PENALTY_FACTOR,
             IS_SMALL_MOLECULE,
             IS_LIVER_ZONATED,
-            USE_AS_SUSPENSION
+            USE_AS_SUSPENSION,
+            ENABLE_SUPERSATURATION
```
msevestre: This was missing here. Actually I am going to remove all those parameters from PKSim so that we do not have duplicates!

PR #559, opened 2019-05-07
Fixes #558 working journal changed on touch

### LINKED ISSUE #558: Beta 7.5: Working Journal altered
If I open a project with a working journal, but never open the journal for editing, but only save the project, the WJ file is altered. Build 4.

msevestre:


src/OSPSuite.UI/Views/Journal/JournalPageEditorView.cs
```diff
@@ -91,6 +91,9 @@ public void BindTo(JournalPageDTO journalPageDTO)
 
          _screenBinder.BindToSource(journalPageDTO);
 
+         //Ensure that the control is not marked as modified after binding. 
+         uxRichEditControl.Modified = false;
```
msevestre: This is the one variable that is sometimes true right after binding. This is used to check if something has changed before saving. I set it to false explicitly and all is well

PR #561, opened 2019-05-07
Fixes #560 sum formula check:

### LINKED ISSUE #560: Sumformula: Do not verify relative path since they are retrieved dynamically

msevestre:


src/OSPSuite.Core/Domain/Services/ModelValidator.cs
```diff
@@ -76,7 +66,11 @@ protected void CheckFormulaIn(IUsingFormula entity, IFormula formulaToCheck, Res
          var entityAbsolutePath = _objectPathFactory.CreateAbsoluteObjectPath(entity).ToPathString();
          var builder = _buildConfiguration.BuilderFor(entity);
          var objectWithError = builder ?? entity;
-         var entityType = _objectTypeResolver.TypeFor(entity);
+
+         // Dynamic formula may contain object path that will be resolved per instance. It cannot be checked here
+         if (formulaToCheck.IsDynamic())
```
msevestre: @Yuri05 This is the change here. Dynamic formula should bypass validation (the actual validation will happen when expanding dynamic formula)

So far we only had simple sum formula so we never had the issue :)

PR #565, opened 2019-05-13
564 export observed data

msevestre:
@Yuri05 The goal is to avoid that here
https://github.com/Open-Systems-Pharmacology/QualificationPlan/blob/master/examples/minimal/reporting%20engine%20input/ObservedData/Itraconazole%20600mg%20MD.csv

Where the SD Column is second

In the reporting Engine, there is no knowledge of AuxiliartyType, related columns ectc.. We have to work with conventions

Col 1: Time
Col 2: Observed Data
Col3: Error

### DIFF
--- a/src/OSPSuite.Infrastructure/Services/DataRepositoryTask.cs
+++ b/src/OSPSuite.Infrastructure/Services/DataRepositoryTask.cs
@@ -273,7 +273,7 @@ namespace OSPSuite.Infrastructure.Services
 
          allAuxiliariesColumns.Each(x => allColumnsNotDrug.Remove(x));
 
-         //First return the Drug column, then non drug columns that are not auxiliary columns then auxiliary columns as last implementation
+         //First return the drug columns, then non drug columns that are not auxiliary columns and last auxiliary columns
          return allDrugColumns.Union(allColumnsNotDrug).Union(allAuxiliariesColumns).ToList();
       }
 
--- a/src/OSPSuite.Infrastructure/Services/DataRepositoryTask.cs
+++ b/src/OSPSuite.Infrastructure/Services/DataRepositoryTask.cs
@@ -8,6 +8,7 @@ using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Data;
 using OSPSuite.Core.Domain.Services;
 using OSPSuite.Core.Domain.UnitSystem;
+using OSPSuite.Core.Extensions;
 using OSPSuite.Utility;
 using OSPSuite.Utility.Collections;
 using OSPSuite.Utility.Extensions;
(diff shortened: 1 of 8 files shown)


src/OSPSuite.Infrastructure/Services/DataRepositoryTask.cs
```diff
@@ -250,20 +249,32 @@ private static string tableNameFrom(DataColumn baseGrid)
...
+         allColumnsNotDrug.Each(x => allDrugColumns.Remove(x));
+
+         var allAuxiliariesColumns = allColumnsNotDrug
+            .Where(x => x.DataInfo.Origin.IsOneOf(ColumnOrigins.CalculationAuxiliary, ColumnOrigins.ObservationAuxiliary))
+            .ToList();
+
+         allAuxiliariesColumns.Each(x => allColumnsNotDrug.Remove(x));
+
+         //First return the drug columns, then non drug columns that are not auxiliary columns and last auxiliary columns
+         return allDrugColumns.Union(allColumnsNotDrug).Union(allAuxiliariesColumns).ToList();
```
msevestre: There is probably a clever way to do this but I could not think of one

tests/OSPSuite.Core.Tests/Core/DataRepositoryTaskSpecs.cs
```diff
@@ -86,6 +86,49 @@ public void should_have_set_the_name_of_the_data_repository_in_the_table()
...
+            DataInfo = {Origin = ColumnOrigins.Observation}
+         };
+         _col2 = new DataColumn("col2", _mass, _baseGrid1)
+         {
+            Values = new float[] {100, 200, 300},
+            DataInfo = {AuxiliaryType = AuxiliaryType.ArithmeticStdDev, Origin = ColumnOrigins.ObservationAuxiliary}
+         };
+
+         _col1.AddRelatedColumn(_col2);
+         _dataRepository.Add(_col2);
```
msevestre: Here: I had Error first and then conc. Before the export would have been as is. Now it follows the convention

tests/OSPSuite.Starter/Presenters/DataRepositoryTestPresenter.cs
```diff
@@ -42,10 +58,20 @@ public DataRepositoryTestPresenter(IDataRepositoryTestView view, IDataRepository
 
       public void Edit(DataRepository repository)
       {
+         _repository = repository;
          _chartPresenter.EditObservedData(repository);
          _dataPresenter.EditObservedData(repository);
          _metaDataPresenter.EditObservedData(repository);
       }
+
+      public Task ExportToCSV()
```
msevestre: Also added a way to test it from the UI...why not :)

src/OSPSuite.Infrastructure/Services/DataRepositoryTask.cs
Yuri05: What will allColumnsNotDrug contain? Metabolites, enzymes, transporters etc.?
Is this of importance, that allColumnsNotDrug are always exported after allColumnsDrug? (why?)

msevestre: This was a requirement from before. Nothing new here. The goal was to always have the drug name at the beginning because it is what matters when you export a simulation file. 

I don't want to change `THAT` behavior

PR #584, opened 2019-07-20
Fixes #583 add get parameter by path

### LINKED ISSUE #583: R-API: Add getParameters by path mtehods

msevestre:


msevestre:
@Yuri05 This is the behavior required for the R implementation usign wild cards

for example

`_task.AllParametersMatching(_organism, "*", "Intracellular", "Volume")` would return all intrecllular volumes in direct container of the organism 

whereas 
`_task.AllParametersMatching(_organism, "**", "Intracellular", "Volume")` would return all intracellular volumes of all direct and indirect children

Also works with `_task.AllParametersMatching(_organism, "**", "Intra*", "Vol*")`

or with `_task.AllParametersMatching(_organism, "**", "*", "Vol*")`

src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -14,6 +14,8 @@ public static class Constants
       public const int MAX_NUMBER_OF_POINTS_PER_INTERVAL = 5000;
       public const int MIN_NUMBER_OF_POINTS_PER_INTERVAL = 2;
       public const int MAX_NUMBER_OF_CHAR_IN_TABLE_NAME = 29;
+      public const string WILD_CARD = "*";
+      public const string WILD_CARD_REC = "**";
```
Yuri05: REC="Recursive"?

msevestre: @Yuri05 Yep. Pretty standard shortcut but I can expand if you prefer

msevestre:
@Yuri05 I found a bug. This is not ready quite yet

src/OSPSuite.Core/Domain/Constants.cs
msevestre: done

src/OSPSuite.R/Services/ContainerTask.cs
```diff
@@ -0,0 +1,89 @@
...
+               pattern.Add($"{ALL_BUT_PATH_DELIMITER}?"); // At least one occurence of a path entry => anything except ObjectPath.PATH_DELIMITER, repeated once
+            else if (string.Equals(entry, WILD_CARD_RECURSIF))
+               pattern.Add(".*"); //Match anything
+            else
+               pattern.Add(entry.Replace(WILD_CARD, ALL_BUT_PATH_DELIMITER));
+         }
+
+         var searchPattern = pattern.ToString($"\\{ObjectPath.PATH_DELIMITER}");
+         return $"^{searchPattern}$";
+      }
```
Yuri05: coole Implementierung :)

src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -14,6 +14,8 @@ public static class Constants
       public const int MAX_NUMBER_OF_POINTS_PER_INTERVAL = 5000;
       public const int MIN_NUMBER_OF_POINTS_PER_INTERVAL = 2;
       public const int MAX_NUMBER_OF_CHAR_IN_TABLE_NAME = 29;
+      public const string WILD_CARD = "*";
+      public const string WILD_CARD_RECURSIF = "**";
```
Yuri05: RECURSI**F** or RECURSI**VE** ?

Yuri05:
Question: is `**` supposed also to be combinable with parts of the path? 
I mean, within the SAME parameter. E.g. 
* should `AllParametersMatching(_organism, "Peri**al", "Volume")` return **Organism|Liver|Periportal|Volume** and **Organism|Liver|Pericentral|Volume**?
* should `AllParametersMatching(_organism, "**Duodenum", "Volume")` return **Organism|Lumen|Duodenum|Volume** and **Organism|SmallIntestine|Mucosa|Duodenum|Volume**?

Or do we want to prohibit something like `"Peri**al"` or `"**Duodenum"` ? (thus only `"**"` is allowed)

msevestre:
I think we should forbid it. This is not how generic path search works so I believe putting ** withing a path entry does not make sense. We could check for this and either
1/ throw an error
2/replace ** with * when within a path entry.

I'd suggest 1 because people need to use the right API

Yuri05:
> I'd suggest 1 because people need to use the right API

agree

src/OSPSuite.R/Services/ContainerTask.cs
```diff
@@ -0,0 +1,101 @@
...
+            return entity == null ? Array.Empty<T>() : new[] {entity};
+         }
+
+         var regex = new Regex(createSearchPattern(path), RegexOptions.IgnoreCase);
+         var parentContainerPath = $"{_entityPathResolver.FullPathFor(container)}{ObjectPath.PATH_DELIMITER}";
+
+         return container.GetAllChildren<T>(x => pathMatches(regex, parentContainerPath, x)).ToArray();
+      }
+
+      private void validate(string[] path)
```
msevestre: @Yuri05 Ok I am now checking that ** is not used as part of path.,
Message  is also quite explicit

PR #587, opened 2019-07-25
Wild card

msevestre:


src/OSPSuite.R/Services/ContainerTask.cs
```diff
@@ -79,15 +82,25 @@ private string createSearchPattern(string[] path)
...
-               pattern.Add($"{ALL_BUT_PATH_DELIMITER}?"); // At least one occurence of a path entry => anything except ObjectPath.PATH_DELIMITER, repeated once
+            if (string.Equals(entry, WILD_CARD)) { 
+               // At least one occurence of a path entry => anything except ObjectPath.PATH_DELIMITER, repeated once
+               pattern.Add($"{ALL_BUT_PATH_DELIMITER}?");
+               pattern.Add(PATH_DELIMITER);
+            }
             else if (string.Equals(entry, WILD_CARD_RECURSIVE))
+            {
                pattern.Add(".*"); //Match anything
+               pattern.Add(OPTIONAL_PATH_DELIMITER);
```
msevestre: @Yuri05 
This was a mistake. Someting like Liver, **, Volume should also return liver volume. This was not the case

PR #592, opened 2019-08-01
Fixes #591 add with display unit extensions

### LINKED ISSUE #591: Add WithDisplayUnitExtensions

msevestre:


msevestre:
@Yuri05 I removed some of the generics because it does not work with rCLR. They were not required anyways in those specific cases

PR #594, opened 2019-08-03
Fixes #593 dummy implementation of simulation runner

### LINKED ISSUE #593: Add SimulationRunner for R Interface

msevestre:


src/OSPSuite.Core/Domain/Data/SimulationResults.cs
```diff
@@ -18,7 +18,9 @@ public class SimulationResults : IEnumerable<IndividualResults>
 
       public virtual QuantityValues Time { get; set; }
 
-      //We need to add the objects in a thread safe manner to the list that is intrinsicly not thread safe
+      public virtual IndividualResults[] IndividualResultsAsArray() => AllIndividualResults?.ToArray();
```
msevestre: R does not work with other more advanced strucutres. We need to export as Array. Using ToArray does not work also because it's generic

src/OSPSuite.R/Api.cs
```diff
@@ -18,5 +18,9 @@ public static void InitializeOnce(ApiConfig apiConfig)
       }
 
       public static IContainerTask GetContainerTask() => IoC.Resolve<IContainerTask>();
+
+      public static ISimulationLoader GetSimulationLoader() => IoC.Resolve<ISimulationLoader>();
```
msevestre: Our one and only entry point to the API

src/OSPSuite.R/Services/SimulationRunner.cs
```diff
@@ -0,0 +1,85 @@
...
+
+namespace OSPSuite.R.Services
+{
+   public interface ISimulationRunner
+   {
+      Task<SimulationResults> RunSimulationAsync(IModelCoreSimulation simulation, SimulationRunOptions simulationRunOptions = null);
+      SimulationResults RunSimulation(IModelCoreSimulation simulation, SimulationRunOptions simulationRunOptions = null);
+   }
+
+   public class SimulationRunner : ISimulationRunner
```
msevestre: This is dummy implementation for now. It is required to get going until SimModel works

PR #604, opened 2019-09-05
602 export pk analysis

msevestre:


src/OSPSuite.Core/Batch/Mappers/SimulationResultsToBatchSimulationExportMapper.cs
```diff
@@ -10,7 +10,7 @@ namespace OSPSuite.Core.Batch.Mappers
 {
    public interface ISimulationResultsToBatchSimulationExportMapper
    {
-      BatchSimulationExport MapFrom(ISimulation simulation, DataRepository results);
+      BatchSimulationExport MapFrom(IModelCoreSimulation simulation, DataRepository results);
```
msevestre: @Yuri05 As discussed .Going the `IModelCoreSimulation` way

src/OSPSuite.Core/Domain/EntityExtensions.cs
```diff
@@ -12,6 +12,19 @@ public static class EntityExtensions
       }
 
       public static string EntityPath(this IEntity entity)
+      {
+         return entityPathAsListFor(entity).ToPathString();
+      }
+
+      public static string ConsolidatedPath(this IEntity entity)
```
msevestre: Being able to create a path without the first entry is a requirement as it is used in many scenarios within a simulation.

src/OSPSuite.Core/Domain/ISimulation.cs
```diff
@@ -1,38 +1,17 @@
 ﻿using System.Collections.Generic;
 using OSPSuite.Core.Chart;
-using OSPSuite.Core.Domain.Builder;
 
 namespace OSPSuite.Core.Domain
 {
-   public interface ISimulation : IObjectBase, ILazyLoadable, IAnalysable, IWithCreationMetaData, IWithModel, IUsesObservedData
+   public interface ISimulation : ILazyLoadable, IAnalysable, IUsesObservedData, IModelCoreSimulation
```
msevestre: I made ISimulation inherit from IModelCoreSimulation. 
Simplifies a lot of the migration

src/OSPSuite.Core/Domain/ModelCoreSimulationExtensions.cs
```diff
@@ -0,0 +1,37 @@
+﻿namespace OSPSuite.Core.Domain
+{
+   public static class ModelCoreSimulationExtensions
```
msevestre: I decided to create extension methods to ensure that more complex code won;'t be duplicated between pksim and mobi (a PKSim simulation inherits from PKSimBuildingBlock and as such won't be able to inherit ModelCoreSimulation)

src/OSPSuite.Core/Domain/OutputSelections.cs
```diff
@@ -14,22 +15,39 @@ public OutputSelections()
...
+      ///    Adds an output corresponding to the <paramref name="quantity" />. The corresponding <see cref="QuantitySelection" />
+      ///    will be added using a consolidated path
+      /// </summary>
+      public virtual void AddQuantity(IQuantity quantity)
       {
-         _allOutputs.Add(quantitySelection);
+         if (quantity == null)
+            return;
+
+         _allOutputs.Add(new QuantitySelection(quantity.ConsolidatedPath(), quantity.QuantityType));
```
msevestre: This methods is required when called from R

src/OSPSuite.Core/Domain/PopulationSimulationPKAnalyses.cs
```diff
@@ -46,7 +46,7 @@ public virtual bool HasPKParameterFor(string quantityPath, string pkParameter)
       /// </summary>
       public virtual IReadOnlyList<QuantityPKParameter> AllPKParametersFor(string quantityPath)
       {
-         return _pkAnalyses.Where(x => string.Equals(x.QuantityPath, quantityPath)).ToList();
+         return _pkAnalyses.Where(x => string.Equals(x.QuantityPath, quantityPath)).ToArray();
```
msevestre: R does not know how to work with list.
Based on the returned value (IReadOnlyList) it should be equivalent to return a ToArray

src/OSPSuite.Core/Domain/QuantityPKParameter.cs
```diff
@@ -11,10 +12,10 @@ public class QuantityPKParameter : IWithDimension, IWithName
       /// <summary>
       ///    The values. One item for each individual
       /// </summary>
-      public virtual List<float> Values { get; private set; }
+      public virtual float[] Values { get; private set; }
```
msevestre: Same here. Need to return an array (it should have been an array at the first place).
This will require a project conversion. In order to not forget, I created an issue in PKSim repo

src/OSPSuite.Core/Domain/Services/SimulationResultsToDataTableConverter.cs
```diff
@@ -0,0 +1,138 @@
...
+      {
+         return Task.Run(() => createResultDataToExport(simulationResults, simulation));
+      }
+
+      public Task<DataTable> PKAnalysesToDataTable(PopulationSimulationPKAnalyses pkAnalyses, IModelCoreSimulation simulation)
+      {
+         return Task.Run(() => createPKAnalysesDataToExport(pkAnalyses, simulation));
+      }
+
+      private DataTable createResultDataToExport(SimulationResults simulationResults, IModelCoreSimulation simulation)
```
msevestre: THis comes from PKSim directly. This is not ideal because the results are then twice in memory. For big export to CSV, this could create contention. However this works for now, especially with 64bit

src/OSPSuite.R/MinimalImplementations/LazyLoadTask.cs
```diff
@@ -0,0 +1,13 @@
+﻿using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Services;
+
+namespace OSPSuite.R.MinimalImplementations
+{
+   public class LazyLoadTask : ILazyLoadTask
+   {
+      public void Load<TObject>(TObject objectToLoad) where TObject : class, ILazyLoadable
+      {
+         //nothing to do
```
msevestre: no LazyLoading in R

src/OSPSuite.R/Services/SimulationRunner.cs
```diff
@@ -32,6 +33,7 @@ public Task<SimulationResults> RunSimulationAsync(IModelCoreSimulation simulatio
       {
          return Task.Run(() =>
          {
+            _simulationPersistableUpdater.UpdateSimulationPersistable(simulation);
```
msevestre: Persitable entities need to be updated based on settings in simulation

PR #606, opened 2019-09-12
605 population management

msevestre:


src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -401,6 +402,8 @@ public static class Population
          public const string VALUE_COLUMN = "Value";
          public const string PARAMETER_PATH_COLUMN = "ParameterPath";
          public const string INDIVIDUAL_ID_COLUMN = "IndividualId";
+         public const string RACE_INDEX = "RaceIndex";
```
msevestre: This all comes from PKSIm

src/OSPSuite.Core/Domain/Populations/Covariates.cs
```diff
@@ -0,0 +1,22 @@
+﻿using OSPSuite.Utility.Collections;
+
+namespace OSPSuite.Core.Domain.Populations
+{
+   public class Covariates
```
msevestre: We will have to do some mapping in PKSim as the covariates object used Gender and Population object that are filled directly from the database

src/OSPSuite.Core/Domain/Populations/ParameterValuesCache.cs
```diff
@@ -0,0 +1,188 @@
...
+         {
+            if (!_parameterValuesCache.Any())
+               return 0;
+            return _parameterValuesCache.First().Count;
+         }
+      }
+
+      private string internalKeyFor(string parameterPath)
+      {
+         if (_parameterValuesCache.Contains(parameterPath))
```
msevestre: This is the new implementation
Checking if the parameter path exists as is (always the case in PKSim)

in R, possibly the path will be used without unit even though it was loaded with unit (because we do not strop the unit)

src/OSPSuite.Core/Domain/Populations/ParameterValuesCache.cs
```diff
@@ -0,0 +1,188 @@
...
+            return parameterPath;
+
+         var allSupportedPath = AllParameterPaths();
+         var index = Array.FindIndex(allSupportedPath, x => x.StartsWith(parameterPath));
+         //try to strip the unit and see if it's possible to find a key exactly equal to the parameterPath without the unit
+         if (index >= 0 && string.Equals(allSupportedPath[index].StripUnit(), parameterPath))
```
msevestre: Checking if we find a key that starts with the full path of the parameter without the unit and verify that they are indeed equal when we strip the unit

src/OSPSuite.Core/Domain/Populations/ParameterValuesCache.cs
```diff
@@ -0,0 +1,188 @@
...
+      private void setValues<T>(string parameterPath, IReadOnlyList<T> newValues, Func<ParameterValues, Action<IReadOnlyList<T>>> addValuesFunc)
+      {
+         validateCount(parameterPath, newValues);
+         var parameterValues = ParameterValuesFor(parameterPath);
+         parameterValues.Clear();
+         addValuesFunc(parameterValues)(newValues);
+      }
+
+      private void validateCount<T>(string parameterPath, IReadOnlyList<T> values)
+      {
```
msevestre: This is to check that we do not create columns that do not have all the values over all individuals( required when adding stuff manually from R)

src/OSPSuite.Core/Domain/Populations/ParameterValuesCache.cs
```diff
@@ -0,0 +1,188 @@
...
+
+         if (Count == values.Count)
+            return;
+
+         throw new OSPSuiteException(Error.ParameterValuesDoNotHaveTheExpectedCount(parameterPath, Count, values.Count));
+      }
+
+      public virtual IReadOnlyCollection<ParameterValues> AllParameterValues => _parameterValuesCache;
+
+      public virtual void Merge(ParameterValuesCache parameterValuesCache, PathCache<IParameter> parameterCache)
```
msevestre: All this stuff taken from PKSim as is

src/OSPSuite.Infrastructure.Import/Services/IndividualValuesCacheImporter.cs
```diff
@@ -0,0 +1,166 @@
...
+using OSPSuite.Core.Domain.Populations;
+using OSPSuite.Core.Extensions;
+using OSPSuite.Core.Services;
+using OSPSuite.Infrastructure.Import.Extensions;
+using OSPSuite.Utility.Collections;
+using OSPSuite.Utility.Extensions;
+
+namespace OSPSuite.Infrastructure.Import.Services
+{
+   public interface IIndividualValuesCacheImporter
```
msevestre: Also comes from PKSim

src/OSPSuite.Infrastructure.Import/Services/IndividualValuesCacheImporter.cs
```diff
@@ -0,0 +1,166 @@
...
+         {
+            logger.AddError(e.FullMessage());
+            return new IndividualValuesCache();
+         }
+      }
+
+      private IndividualValuesCache withPathsContainingUnitsUpdated(IndividualValuesCache individualValuesCache, PathCache<IParameter> allParameters, IImportLogger logger)
+      {
+         // No parameters to check from, return as IS
+         if (allParameters == null)
```
msevestre: This is the case from R. In PKSim, the list will be given and as  such, the parameters will be striped from units

src/OSPSuite.R/Services/PopulationImporter.cs
```diff
@@ -0,0 +1,41 @@
...
+      public IndividualValuesCache ImportPopulation(string fileFullPath)
+      {
+         var importLogger = new ImportLogger();
+         var parameterValuesCache = _individualValuesCacheImporter.ImportFrom(fileFullPath, importLogger);
+         if (importLogger.Status.Is(NotificationType.Error))
+            throw new OSPSuiteException(importLogger.LogString);
+
+         if (importLogger.Status.Is(NotificationType.Warning))
+         {
+            importLogger.Log.Each(Console.WriteLine);
```
msevestre: Not sure how we will communicate with R. This will be part of the Logging task

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1370,6 +1366,63 @@ public static string LastDataRowLessThanFirstDataRow(int dataEndRow, int dataSta
 
       public const string QualificationReportConfigurationFileNotDefined = "Qualification report configuration file not defined.";
 
+      public static string ParameterValuesDoNotHaveTheExpectedCount(string parameterPath, int expectedCount, int actualCount)
```
msevestre: All of that comes from PKSim. Import of results is now in CORE

src/OSPSuite.Infrastructure.Import/Services/IndividualResultsImporter.cs
```diff
@@ -0,0 +1,165 @@
...
+using OSPSuite.Core.Extensions;
+using OSPSuite.Core.Services;
+using OSPSuite.Infrastructure.Import.Extensions;
+using OSPSuite.Utility.Collections;
+using OSPSuite.Utility.Exceptions;
+using OSPSuite.Utility.Extensions;
+
+namespace OSPSuite.Infrastructure.Import.Services
+{
+   public interface IIndividualResultsImporter
```
msevestre: 100% copy out of PKSim

src/OSPSuite.Infrastructure.Import/Services/SimulationResultsImporter.cs
```diff
@@ -0,0 +1,209 @@
...
+using System.Threading.Tasks;
+using OSPSuite.Assets;
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Data;
+using OSPSuite.Core.Domain.Services;
+using OSPSuite.Core.Services;
+using OSPSuite.Utility.Events;
+using OSPSuite.Utility.Extensions;
+
+namespace OSPSuite.Infrastructure.Import.Services
```
msevestre: ALso 100% comes from PKSim with the exception of one check that has been moved  out and will need to be done in PKsim. (actually not sure if this is required). I have created an issue for htat

src/OSPSuite.Core/Domain/Populations/ParameterValuesCache.cs
Yuri05: 2 concerns regarding the `internalKEy`-function:

1. Following situation:
 - Cache contains paths `"Container|P1"` and `"Container|P [ml]"` and `AllParameterPaths()` retrieves them in this order
 - we call `internalKeyFor("Container|P")`
 - `index = Array.FindIndex(allSupportedPath, x => x.StartsWith(parameterPath))` will return the index of `"Container|P1"`
 - `string.Equals(allSupportedPath[index].StripUnit(), parameterPath))` will return false and thus no internal key will be found. But it should be `"Container|P [ml]"`

 ➡️ I think we should use e.g. `FindAll` instead of `FindIndex` and apply unit removal to all such keys.

2. I still not quite sure if I fully understand the use case and may be the example below will never occur. Following situation:
 - cache contains `"Container|P"`
 - we call `internalKeyFor("Container|P [ml]")`
 - `Array.Find(allSupportedPath, x => x.StartsWith(parameterPath))` will not find anything and thus no internal key will be found. But it should be `"Container|P"`.

 ➡️ If this use case is valid, we should do something like
```
Array.Find(allSupportedPath, x => x.StartsWith(parameterPath) || parameterPath.StartsWith(x)
...
if (... && (string.Equals(allSupportedPath[index].StripUnit(), parameterPath) || string.Equals(allSupportedPath[index], parameterPath.StripUnit()))
...
```

msevestre: >I think we should use e.g. FindAll instead of FindIndex and apply unit removal to all such keys.

Fair enough. I'll implement


>I still not quite sure if I fully understand the use case and may be the example below will never occur. Following situation:

I don't think it can occur. This whole change is about making sure we can retrieve output by path even though they were loaded with unit (from file)

in PK-Sim:  PATH WILL NEVER BE CACHED WITH UNIT. We strip the unit when we load
in R:PATH may be cached with unit. If the cache contains unit, then the user will be able to retrtrieve with unit and without unit. This makes sense because user uses quantity and should not care about the unit

HOWEVER, there is absolutely no reason for a cache that does not contains unit to be acccess with unit. I don't see a problem here and thus believe this use case in invalid

Yuri05: >HOWEVER, there is absolutely no reason for a cache that does not contains unit to be acccess with unit. I don't see a problem here and thus believe this use case in invalid

OK

msevestre: Done. 

this is the code now
```
 var matchingPath = Array.FindAll(AllParameterPaths(), x => x.StartsWith(parameterPath))
            .FirstOrDefault(p => string.Equals(p.StripUnit(), parameterPath));
```

I think this cover all tests cases. I have also added more test cases to support the scenario above

src/OSPSuite.R/MinimalImplementations/ProgressManager.cs
```diff
@@ -0,0 +1,12 @@
+﻿using OSPSuite.Utility.Events;
+
+namespace OSPSuite.R.MinimalImplementations
+{
+   public class ProgressManager : IProgressManager
```
msevestre: @Yuri05 Theoretically, that's where we could implement some progress stuff for R. Version 2 but possible

tests/OSPSuite.Core.Tests/Domain/ParameterValuesCacheSpecs.cs
```diff
@@ -0,0 +1,351 @@
...
+   public class When_adding_a_parameter_value_with_similar_path_to_a_parameter_value_cache : concern_for_ParameterValuesCache
+   {
+      private ParameterValue _parameterValue1;
+      private ParameterValue _parameterValue2;
+
+      protected override void Context()
+      {
+         base.Context();
+         _parameterValue1 = new ParameterValue("PATH_P1", 11, 0.1);
+         _parameterValue2 = new ParameterValue("PATH_P [mg]", 21, 0.2);
```
msevestre: @Yuri05 This is your use case. Did not work at the beginning. Worked after the fix

PR #615, opened 2019-09-25
Work on retintegration of simmodel on pop

msevestre:
WIP Juri. 
As far as I can tell, it does not work. I am only getting NaN when I simulate the tests (all tests are failing)


src/OSPSuite.Core/Domain/Services/PopulationDataSplitter.cs
```diff
@@ -1,167 +1,166 @@
...
+            // if no aging data found, parameter value is stored in the population data table
+            // (single value)
+            if (!parameterProperties.TablePoints.Any())
+               parameterProperties.Value = nonTableParameterValues[parameterProperties.Path].ConvertedTo<double>();
+         }
+
+         var initialValues = initialValuesDataRowFromIndividualId(individualId);
+         if (initialValues == null)
+            return;
+         //TODO
```
msevestre: This is not available yet

tests/OSPSuite.Core.Tests/Services/PopulationDataSplitterSpecs.cs
```diff
@@ -102,37 +101,37 @@
...
 //      protected override void Context()
 //      {
 //         base.Context();
 //
-//         _variableParameters = new List<IParameterProperties>();
+//         _variableParameters = new List<ParameterProperties>();
 //
-//         var parameterInfo1 = A.Fake<IParameterProperties>();
-//         A.CallTo(()=>parameterInfo1.IsTable).Returns(false);
+//         var parameterInfo1 = A.Fake<ParameterProperties>();
```
msevestre: Because we have no interface, I won't be able to fake those guys, 
I will rewrite thsoe tests completely to be interactiontests. But first, the PopRunner needs to work

tests/OSPSuite.Core.Tests/Services/PopulationDataSplitterSpecs.cs
```diff
@@ -183,77 +183,80 @@
...
+////         var tableParameter1 = _variableParameters.First(x => x.IsTable && x.Path.Equals(_agingPath1));
+////         var tablePoints1 = tableParameter1.TablePoints;
+////
+////         tablePoints1.Count().ShouldBeEqualTo(2);
+////         tablePoints1.ElementAt(0).X.ShouldBeEqualTo(1);
+////         tablePoints1.ElementAt(0).Y.ShouldBeEqualTo(1);
+////         tablePoints1.ElementAt(1).X.ShouldBeEqualTo(2);
+////         tablePoints1.ElementAt(1).Y.ShouldBeEqualTo(2);
+////
+////         var tableParameter2 = _variableParameters.First(x => x.IsTable && x.Path.Equals(_agingPath2));
```
msevestre: @Yuri05 IsTable is gone from the interface. A bunch of stuff does not work

msevestre:
@Yuri05 There are no more TODO here as far as I can tell. I am going to merge this and if you agree, I'll merge the .NET Standard package into develop (there is one more .NET Standard DevExpress dll to extract as package but this can be done later)

msevestre:
@Yuri05 Running out of idea here
https://ci.appveyor.com/project/open-systems-pharmacology-ci/ospsuite-core/builds/28209411 

for some reason, the freaking Native dll is not found. The message is probably wrong

msevestre:
well I had to use VS 2019! WTF?

Yuri05:
>There are no more TODO here as far as I can tell. I am going to merge this and if you agree, I'll merge the .NET Standard package into develop (there is one more .NET Standard DevExpress dll to extract as package but this can be done later)

Fine with that :)

>well I had to use VS 2019! WTF?

Well, all C++ DLLs were compiled with VS2019, maybe VS2017 image does not have 2019 C++ Runtime or so?

msevestre:
Meaning we will have to update to the new runtime in setup. Good to know :)

Yuri05:
>Meaning we will have to update to the new runtime in setup. Good to know :)

Yep :)

msevestre:
IS there a reason for this? It now breaks the R build too
https://ci.appveyor.com/project/open-systems-pharmacology-ci/ospsuite-r/builds/28216680

and I am not sure that I can specify an image . let me try but this is really painful

Yuri05:
>IS there a reason for this?

Apart from the fact that VS2017 was the worst VS version ever: VS2019 has some useful new features.
E.g. under Win10 you can compile/run/debug C++ and .NET core applications under WSL (Windows Subsystem for Linux) directly from within VS. In a case where the windows version works and the Linux version does not this could be really helpful.
However I think none of new features in VS2019 is really a MUST. If we will have some problems with it, we can downgrade to VS2017

msevestre:
Managed to get it to work

And VS2005 was shit in comparison ;-)

PR #655, opened 2019-10-18
Fixes #654 update covariate API

### LINKED ISSUE #654: Update IndividualValueCache to support array instead of list for all covariates properties

msevestre:


msevestre:
@Yuri05 Just make sure that we return arrray and not list so that we can use in R

PR #657, opened 2019-10-24
Fixes perf update for R

msevestre:


src/OSPSuite.Core/Domain/Data/IndividualResults.cs
```diff
@@ -53,17 +53,22 @@ public virtual void Add(QuantityValues values)
...
-      public virtual QuantityValues ValuesFor(string quantityPath)
+      public virtual QuantityValues QuantityValuesFor(string quantityPath)
       {
          return AllValues.FirstOrDefault(x => Equals(x.QuantityPath, quantityPath));
       }
 
+      /// <summary>
+      ///    Returns the values defined for the quantity with path <paramref name="quantityPath" /> or null it they do not exist
+      /// </summary>
+      public float[] ValuesFor(string quantityPath) => QuantityValuesFor(quantityPath)?.Values;
```
msevestre: @Yuri05 REturning all values for one path as opposed to having to iterate through all entries  in R

PR #658, opened 2019-10-25
Add presnetation serialization that was missing so far

msevestre:


msevestre:
@Yuri05 I had forgotten to add this class back. Serialization for presentation objects.

PR #673, opened 2019-11-07
Fixes #672 wrong percentile serialization

### LINKED ISSUE #672: Percentile serialization for ParameterValues is wrong

msevestre:


src/OSPSuite.Core/Serialization/Xml/ParameterValuesXmlSerializer.cs
```diff
@@ -40,7 +40,9 @@ protected override XElement TypedSerialize(ParameterValues parameterValues, Seri
             return element;
 
          var doubleSerializer = SerializerRepository.SerializerFor(parameterValues.Percentiles);
-         element.Add(doubleSerializer.Serialize(parameterValues.Percentiles, serializationContext));
+         var percentilesElement = doubleSerializer.Serialize(parameterValues.Percentiles, serializationContext);
+         percentilesElement.Name = Constants.Serialization.PERCENTILES;
```
msevestre: This node did not have the right name => E.g. it was not deserialized properly

Yuri05: @msevestre why is percentile serialized at all? Isn't it always calculated from the distribution and value of the parameter?

Issue #684, opened 2019-11-26
R Implementation: Expose Sensitivy Analysis calculation


msevestre:
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/220

PR #685, opened 2019-11-28
WIP #684 sensitivity calculation for R

msevestre:


msevestre:
@Yuri05 This PR aims at supporting the basic sensitivity use case from R

rakefile.rb
```diff
@@ -1,4 +1,4 @@
-# require_relative 'scripts/coverage'
+require_relative 'scripts/coverage'
```
msevestre: unreladed but it was missing (forgotten to uncomment)

src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -90,6 +90,9 @@ public static class Constants
       public const byte RANGE_AREA_OPACITY = 55;
       public const byte RANGE_AREA_TRANSPARENCY = 255 - RANGE_AREA_OPACITY;
       public const int FEEDBACK_REFRESH_TIME = 1000; //refresh time in ms
+      public const int DEFAULT_SENSITIVITY_NUMBER_OF_STEPS = 2;
```
msevestre: I have defined those values as constant

src/OSPSuite.Core/Domain/ISimulation.cs
```diff
@@ -6,12 +6,5 @@ namespace OSPSuite.Core.Domain
    public interface ISimulation : ILazyLoadable, IAnalysable, IUsesObservedData, IModelCoreSimulation
    {
       IEnumerable<CurveChart> Charts { get; }
-
-      /// <summary>
-      ///    Name of all compounds used in the simulation
-      /// </summary>
-      IReadOnlyList<string> CompoundNames { get; }
```
msevestre: I am moving as much as possible to ModelCoreSimulation so that I don't have to repeat myself all the time.
ISimulation is now basicaly a IModelCoreSimulation with analysis and observed data. Nothing more

src/OSPSuite.Core/Domain/ModelCoreSimulation.cs
```diff
@@ -27,13 +27,20 @@ public interface IModelCoreSimulation : IObjectBase, IWithCreationMetaData, IWit
...
+
+      IEnumerable<T> All<T>() where T : class, IEntity;
    }
 
    public class ModelCoreSimulation : ObjectBase, IModelCoreSimulation
    {
       public IModel Model { get; set; }
+
       public IBuildConfiguration BuildConfiguration { get; set; }
-      public virtual DataRepository Results { get; set; }
```
msevestre: Not sure why this is here of this is required. I think for MoBi but then I'll definie it in MoBi directly

src/OSPSuite.Core/Domain/ParameterExtensions.cs
```diff
@@ -37,5 +37,14 @@ public static bool IsOfType(this IParameter parameter, PKSimBuildingBlockType bu
       {
          return parameter.BuildingBlockType.Is(buildingBlockType);
       }
+
+      /// <summary>
+      ///    Returns true if the value of the parameter was set by the user or if the formula is a constant formula or a
+      ///    distributed formula (E.g. an actual constant value that does not depend on other parameter or time)
+      /// </summary>
+      public static bool IsConstantParameter(this IParameter parameter)
```
msevestre: @Yuri05 Do you agree that this is the right definition?
In the presenter (AddAllConstant) we  do fixed and !formula.IsExplicit. This works because we filter other parameter before but this is not really ergnomic

I think constant parameters in our case are
FixedValue
Or
ConstantFormula
or DistributedFormula

src/OSPSuite.Core/Domain/RunOptions.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace OSPSuite.Core.Domain
+{
+   public class RunOptions
+   {
+      /// <summary>
+      ///    (Maximal) number of cores to be used (1 per default)
+      /// </summary>
+      public int NumberOfCoresToUse { get; set; } = 1;
```
msevestre: This is the most basic option now that needs to be passed along to all our runner. We did not have that for sensitivity

src/OSPSuite.Core/Domain/SensitivityAnalyses/SensitivityParameter.cs
```diff
@@ -84,41 +84,26 @@ public static IEnumerable<IBusinessRule> All
...
-         private static IBusinessRule nameNotEmpty
-         {
-            get { return GenericRules.NonEmptyRule<SensitivityParameter>(x => x.Name); }
-         }
+         private static IBusinessRule nameNotEmpty { get; } = GenericRules.NonEmptyRule<SensitivityParameter>(x => x.Name);
 
-         private static IBusinessRule nameUnique
-         {
-            get
+         private static IBusinessRule nameUnique { get; } = CreateRule.For<SensitivityParameter>()
```
msevestre: small optimization to create one instance instead of one every time it's asked for

src/OSPSuite.Core/Domain/Services/SensitivityAnalyses/ISensitivityAnalysisTask.cs
```diff
@@ -11,24 +11,29 @@ public interface ISensitivityAnalysisTask
...
-      /// Asks the user if he really wants to delete the <paramref name="sensitivityAnalyses"/>. Returns <c>true</c> if the <paramref name="sensitivityAnalyses"/>
-      /// are deleted from the current project otherwise <c>false</c>
+      ///    Asks the user if he really wants to delete the <paramref name="sensitivityAnalyses" />. Returns <c>true</c> if the
+      ///    <paramref name="sensitivityAnalyses" />
+      ///    are deleted from the current project otherwise <c>false</c>
       /// </summary>
       bool Delete(IReadOnlyList<SensitivityAnalysis> sensitivityAnalyses);
 
       void SwapSimulations(SensitivityAnalysis sensitivityAnalysis, ISimulation oldSimulation, ISimulation newSimulation);
-      bool ValidateSwap(SensitivityAnalysis sensitivityAnalysis, ISimulation oldSimulation, ISimulation newSimulation);
```
msevestre: This task was using some presenter stuff. I had to remove it outside of usage so that I can reuse. This is better now

src/OSPSuite.Core/Domain/Services/SensitivityAnalyses/SensitivityAnalysisEngine.cs
```diff
@@ -9,48 +9,49 @@ namespace OSPSuite.Core.Domain.Services.SensitivityAnalyses
...
+         IEventPublisher eventPublisher, 
+         ISensitivityAnalysisVariationDataCreator sensitivityAnalysisVariationDataCreator,
+         IPopulationRunner populationRunner, 
+         ISimulationToModelCoreSimulationMapper modelCoreSimulationMapper,
+         ISensitivityAnalysisRunResultCalculator runResultCalculator, 
+         ISimulationPersistableUpdater simulationPersistableUpdater)
       {
          _eventPublisher = eventPublisher;
-         _sensitivyAnalysisVariationDataCreator = sensitivyAnalysisVariationDataCreator;
+         _sensitivityAnalysisVariationDataCreator = sensitivityAnalysisVariationDataCreator;
```
msevestre: typo

src/OSPSuite.Core/Domain/Services/SensitivityAnalyses/SensitivityAnalysisRunner.cs
```diff
@@ -46,10 +53,11 @@ public async Task Run(SensitivityAnalysis sensitivityAnalysis)
             if (IsRunning)
                throw new OSPSuiteException(Error.CannotStartTwoConcurrentSensitivityAnalyses);
 
+            var options = runOptions ?? new RunOptions {NumberOfCoresToUse = _coreUserSettings.MaximumNumberOfCoresToUse};
```
msevestre: If it's not specified, we use the user settings (only for PKSim and MoBi)

src/OSPSuite.Core/Domain/Services/SensitivityAnalyses/SensitivityAnalysisTask.cs
```diff
@@ -125,10 +100,7 @@ public bool Delete(IReadOnlyList<SensitivityAnalysis> sensitivityAnalyses)
 
       private void delete(SensitivityAnalysis sensitivityAnalysis)
       {
-         _applicationController.Close(sensitivityAnalysis);
```
msevestre: I remove the usage here so I need to react to the SensitivityAnalysisDeletedEvent somewhere else to close the view

src/OSPSuite.Core/Services/IDialogCreator.cs
```diff
@@ -14,33 +14,26 @@ public interface IDialogCreator
    {
       void MessageBoxError(string message);
 
-      ViewResult MessageBoxYesNoCancel(string message);
-
-      ViewResult MessageBoxYesNoCancel(string message, ViewResult defaultButton);
+      ViewResult MessageBoxYesNoCancel(string message, ViewResult defaultButton = ViewResult.Yes);
```
msevestre: using default value to remove duplicate

src/OSPSuite.Presentation/Core/CloseSubjectPresenterInvokerBase.cs
```diff
@@ -3,7 +3,9 @@
 
 namespace OSPSuite.Presentation.Core
 {
-   public interface ICloseSubjectPresenterInvokerBase : IListener<ObservedDataRemovedEvent>
+   public interface ICloseSubjectPresenterInvokerBase : 
+      IListener<ObservedDataRemovedEvent>,
+      IListener<SensitivityAnalysisDeletedEvent>
```
msevestre: Here. I am listeing to the event so that I can close the presenter

src/OSPSuite.Presentation/Presenters/SensitivityAnalyses/SensitivityAnalysisParameterSelectionPresenter.cs
```diff
@@ -107,12 +117,20 @@ public void ChangeSimulation(ISimulation simulation)
 
       public bool ValidateSimulationChange(ISimulation newSimulation)
       {
-         return _sensitivityAnalysisTask.ValidateSwap(_sensitivityAnalysis, _sensitivityAnalysis.Simulation, newSimulation);
+         var swapValidationResult = _sensitivityAnalysisTask.ValidateSwap(_sensitivityAnalysis, _sensitivityAnalysis.Simulation, newSimulation);
+
+         if (swapValidationResult.ValidationState == ValidationState.Valid)
+            return true;
+
+         using (var validationMessagesPresenter = _applicationController.Start<IValidationMessagesPresenter>())
```
msevestre: move the logic that was in the task in the presenter. That way I can resue the task for R

src/OSPSuite.Presentation/Presenters/SensitivityAnalyses/SensitivityAnalysisParameterSelectionPresenter.cs
```diff
@@ -123,11 +141,7 @@ public int IconFor(ISimulation simulation)
 
       public ISimulation Simulation => _sensitivityAnalysis.Simulation;
 
-      private static bool isNonFormulaParameter(ParameterSelection parameterSelection)
-      {
-         var parameter = parameterSelection.Parameter;
-         return parameter.IsFixedValue || !parameter.Formula.IsExplicit();
```
msevestre: See comment above. This was the same because we are filtering parameters but I think this is better like that

src/OSPSuite.R/Domain/SensitivityAnalysis.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Collections.Generic;
+using OSPSuite.Core.Domain;
+
+namespace OSPSuite.R.Domain
+{
+   public class SensitivityAnalysis
```
msevestre: This will be the object used from R. Ref to simulation and list of parmaeters to vary

src/OSPSuite.R/Domain/Simulation.cs
```diff
@@ -0,0 +1,109 @@
...
+using OSPSuite.Core.Chart;
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Core.Domain.Data;
+using OSPSuite.Core.Domain.Services;
+using OSPSuite.Utility.Visitor;
+
+namespace OSPSuite.R.Domain
+{
+   public class Simulation : ISimulation
```
msevestre: A simulation in R delegates everyhting to the core simulation for now

src/OSPSuite.R/Mapper/SensitivityAnalysisToCoreSensitivityAnalysisMapper.cs
```diff
@@ -0,0 +1,73 @@
...
+
+      private IReadOnlyList<IParameter> parametersToVaryFrom(SensitivityAnalysis sensitivityAnalysis)
+      {
+         var parametersToVary = new List<IParameter>(sensitivityAnalysis.Parameters);
+         if (parametersToVary.Any())
+            return parametersToVary;
+
+         var simulation = sensitivityAnalysis.Simulation;
+         var constantParametersCache = _containerTask.CacheAllChildrenSatisfying<IParameter>(simulation.Model.Root, x => _parameterSelector.CanUseParameter(x) && x.IsConstantParameter());
+         var allUsedParameterPaths = _simulationAnalyzer.AllPathOfParametersUsedInSimulation(sensitivityAnalysis.Simulation);
```
msevestre: This is my try to get all used parameters from SimModel. But I believe this does not work

src/OSPSuite.R/MinimalImplementations/FullPathDisplayResolver.cs
```diff
@@ -0,0 +1,27 @@
...
+   public class FullPathDisplayResolver : IFullPathDisplayResolver
+   {
+      private readonly IEntityPathResolver _entityPathResolver;
+
+      public FullPathDisplayResolver(IEntityPathResolver entityPathResolver)
+      {
+         _entityPathResolver = entityPathResolver;
+      }
+
+      public string FullPathFor(IObjectBase objectBase, bool addSimulationName = false)
```
msevestre: Looks like a real implementation to me that could be moved to core. But good for now

src/OSPSuite.R/MinimalImplementations/RExecutionContext.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System;
+using OSPSuite.Core.Commands;
+using OSPSuite.Core.Commands.Core;
+using OSPSuite.Core.Domain;
+
+namespace OSPSuite.R.MinimalImplementations
+{
+   public class RExecutionContext : IOSPSuiteExecutionContext
```
msevestre: A little bit unclean. But for the SA, we need th execution context that is now quite substantial. Defining a defualt one here

src/OSPSuite.R/MinimalImplementations/RExecutionContext.cs
```diff
@@ -0,0 +1,74 @@
...
+      {
+         //nothing to do
+      }
+
+      public void ProjectChanged()
+      {
+         //nothing to do
+      }
+
+      public IProject Project { get; } = new RProject();
```
msevestre: This is really the one object we need . rest is really never used in R

src/OSPSuite.R/Services/ParameterAnalysableParameterSelector.cs
```diff
@@ -0,0 +1,19 @@
+﻿using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Services;
+
+namespace OSPSuite.R.Services
+{
+   public class ParameterAnalysableParameterSelector : AbstractParameterAnalysableParameterSelector
+   {
+      public override bool CanUseParameter(IParameter parameter)
```
msevestre: Here is where we filter stuff out. 
So we can never vary parameter that are !CanBeVaried or REadOnlyu or table or categorial (e..g we have the info in core already but hard coded)

src/OSPSuite.R/Services/SimulationAnalyzer.cs
```diff
@@ -0,0 +1,29 @@
...
+
+   public class SimulationAnalyzer : SimModelManagerBase, ISimulationAnalyzer
+   {
+      public SimulationAnalyzer(ISimModelExporter simModelExporter, ISimModelSimulationFactory simModelSimulationFactory) : base(simModelExporter, simModelSimulationFactory)
+      {
+      }
+
+      public IReadOnlyList<string> AllPathOfParametersUsedInSimulation(IModelCoreSimulation modelCoreSimulation)
+      {
+         var simulationExport = CreateSimulationExport(modelCoreSimulation, SimModelExportMode.Optimized);
```
msevestre: @Yuri05 I have no idea what to do. This seems to always return everything (e.g. all VF parameters even tho they are not used)

Yuri05: Ok, I know why. Per default identification of used parameters is disabled in SimModel (because this information was not used by PK-Sim/MoBi but only by the Toolboxes).
In order to activate it, you must set [SimModelSimulation.Options.IdentifyUsedParameters](https://github.com/Open-Systems-Pharmacology/OSPSuite.SimModel/blob/2d478758d11d8f9f8df234701a5624108e434784/src/OSPSuite.SimModel/SimulationOptions.cs#L134) to true before loading SimModel-Simulation from XML.
E.g. you could define `identifyUsedParameters `as an optional parameter for the `CreateSimulation `function (default: false) and [set it before loading:](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/06ec0ffe87fbd46cdee8a6c5c3580fad905f6dd1/src/OSPSuite.Core/Domain/Services/SimModelManagerBase.cs#L47)
```
    var simulation = _simModelSimulationFactory.Create();
    simulation.Options.IdentifyUsedParameters = identifyUsedParameters;
    simulation.LoadFromXMLString(simulationExport);
```

Yuri05: Guess one day we should move this functionality from SimModel to OSPSuiteCore (especially if we would like to mark/hide unused parameters in PK-Sim/MoBi)

src/OSPSuite.Core/Domain/ParameterExtensions.cs
Yuri05: Seems correct to me. Until we find a use case where it is not :)

src/OSPSuite.R/MinimalImplementations/ExceptionManager.cs
```diff
@@ -0,0 +1,14 @@
...
+using OSPSuite.Utility.Exceptions;
+using OSPSuite.Utility.Extensions;
+
+namespace OSPSuite.R.MinimalImplementations
+{
+   public class ExceptionManager : ExceptionManagerBase
+   {
+      public override void LogException(Exception ex)
+      {
+         Console.WriteLine(ex.FullMessage());
```
Yuri05: Here _Console_ would be R output window? (or would the error message just disappear?)

src/OSPSuite.R/MinimalImplementations/FullPathDisplayResolver.cs
Yuri05: Don't we have something like this already in core?

src/OSPSuite.R/Services/ParameterAnalysableParameterSelector.cs
```diff
@@ -0,0 +1,19 @@
...
+      {
+         return parameter.CanBeVaried
+                && !parameter.Info.ReadOnly
+                && !ParameterIsTable(parameter)
+                && !ParameterIsCategorial(parameter);
+
```
Yuri05: `IsUsedInSimulation `is checked somewhere else, correct?

src/OSPSuite.R/MinimalImplementations/ExceptionManager.cs
msevestre: yes R output

src/OSPSuite.R/Services/SimulationAnalyzer.cs
msevestre: ok/ What's the difference betweeen variableParmaeters and ParameterProperties?

Yes. This would be useful to set this property after the first run for example. One could even hide those parameters in PKSim or mark them with a grayed out color (e..g unused)

src/OSPSuite.R/MinimalImplementations/FullPathDisplayResolver.cs
msevestre: In fact no. We haev the same implementation in MoBi but PKSim uses something different using the representationInfo repository. If we move that concept out of PKSim and into core, and have MobI simple return name for display name, then we could reuse the same logic. Might do it next

src/OSPSuite.R/Services/SimulationAnalyzer.cs
Yuri05: ParameterProperties are all parameters, VariableParameters are those which can be varied between simulation runs (are set between Load and Finalize)

src/OSPSuite.Core/Domain/Services/SimModelManagerBase.cs
```diff
@@ -44,9 +44,10 @@ protected void FinalizeSimulation(Simulation simModelSimulation)
          simModelSimulation.FinalizeSimulation();
       }
 
-      protected Simulation CreateSimulation(string simulationExport)
+      protected Simulation CreateSimulation(string simulationExport, Action<SimulationOptions> configureSimulationAction = null)
```
msevestre: @Yuri05 I have added a configuration method that can be called. That way, we do not have to add one flag for each possible option. Instead caller can decide what to do

Yuri05: Nice!

msevestre:
@Yuri05 Going to merge here. Please review the changes when you have time. This PR exposes ALL required objects to run a SA and get the sensitivity as in PKSim

We might need to expose lower primitive once we start to use it
I do not know how we can integrate dynamic PK Parameter at that stage since there is no way to callback to R

PR #695, opened 2020-01-11
693 expose data spliiter

msevestre:


msevestre:
@Yuri05 This introduces the concept of IndividualIds stored separately from the rest. So far, we always assumed that ids were always starting at zero when generating population. but this is not quite the case with split files (NEW concept)

src/OSPSuite.Core/Domain/Populations/CovariateValues.cs
```diff
@@ -3,9 +3,9 @@
 
 namespace OSPSuite.Core.Domain.Populations
 {
-   public class CovariateValues : IWithName
+   public class CovariateValues
    {
-      public string Name { get; set; }
+      public string CovariateName { get; set; }
```
msevestre: I have renamed this guy to be clearer. No need for conversion as this class is brand new anyways

src/OSPSuite.Core/Domain/Populations/CovariateValuesCache.cs
```diff
@@ -102,8 +102,8 @@ public virtual CovariateValues GetOrAddCovariateValuesFor(string covariateName)
 
       public void Remove(string covariateName) => _covariateValuesCache.Remove(covariateName);
 
-      public string[] AllCovariateValuesForIndividual(int individualId) => AllCovariateValues.Select(x => x.ValueAt(individualId)).ToArray();
+      public string[] AllCovariateValuesForIndividual(int individualIndex) => AllCovariateValues.Select(x => x.ValueAt(individualIndex)).ToArray();
```
msevestre: This was assuming index == id which is wrong. We do not support retrieval by individualId at the moment, only by index

src/OSPSuite.Core/Domain/Populations/IndividualValues.cs
```diff
@@ -5,6 +5,8 @@ namespace OSPSuite.Core.Domain.Populations
 {
    public class IndividualValues
    {
+      public int? IndividualId { get; set; }
```
msevestre: Null=>id will be set automatically. This may not be required as I believe this is only used in pksim when crfeating a population but better be consistent.

src/OSPSuite.Core/Domain/Populations/IndividualValuesCache.cs
```diff
@@ -1,26 +1,41 @@
...
       public void Add(IndividualValues individualValues)
       {
          CovariateValuesCache.AddIndividualValues(individualValues.Covariates);
          Add(individualValues.ParameterValues);
+         IndividualIds.Add(getIndividualIdFor(individualValues));
+      }
+
+      private int getIndividualIdFor(IndividualValues individualValues)
+      {
+         if (individualValues.IndividualId != null)
```
msevestre: In that case, if the id is null, we are just using a new id.

src/OSPSuite.Core/Domain/Populations/IndividualValuesCache.cs
```diff
@@ -1,26 +1,38 @@
...
          ParameterValuesCache = parameterValuesCache;
          CovariateValuesCache = covariateValuesCache;
+         IndividualIds = individualIds;
       }
 
       public void Add(IndividualValues individualValues)
       {
          CovariateValuesCache.AddIndividualValues(individualValues.Covariates);
          Add(individualValues.ParameterValues);
+         IndividualIds.Add(getNextIndividualId());
```
msevestre: That's coming from PKSim. Ids assigned automatically

src/OSPSuite.Core/Domain/Populations/IndividualValuesCache.cs
```diff
@@ -32,7 +44,7 @@ public void Add(IndividualValues individualValues)
       /// <summary>
       ///    Returns the number of individuals
       /// </summary>
-      public virtual int Count => ParameterValuesCache.Count;
```
msevestre: This was quite confusing. The count here was the not the number of items in the cache but the number of values in the first item of the cache (e.g. number of individuals). Like that, this is much clearner

src/OSPSuite.Infrastructure.Import/Services/IndividualValuesCacheImporter.cs
```diff
@@ -125,12 +125,13 @@ private IndividualValuesCache createIndividualPropertiesFromCSV(CsvReader csv, s
          {
             for (int i = 0; i < fieldCount; i++)
             {
+               var header = headers[i];
                if (i == indexIndividualId)
-                  continue;
+                  individualPropertiesCache.IndividualIds.Add(csv.IntAt(i));
```
msevestre: IndividualIds are now loaded from CSV

src/OSPSuite.R/Services/PopulationTask.cs
```diff
@@ -1,20 +1,30 @@
...
-      DataTable PopulationTableFrom(IndividualValuesCache population, IModelCoreSimulation simulation);
+
+      DataTable PopulationTableFrom(IndividualValuesCache population, IModelCoreSimulation simulation = null);
+
+      /// <summary>
+      /// Loads the population from the <paramref name="populationFile"/> and split the loaded population according to the <paramref name="numberOfCores"/>.
+      /// Resulting files will be exported in the <paramref name="outputFolder"/>. File names will be constructed using the <paramref name="outputFileName"/> concatenated with the node index.
+      /// Returns an array of string containing the full path of the population files created
+      /// </summary>
+      IReadOnlyList<string> SplitPopulation(string populationFile, int numberOfCores, string outputFolder, string outputFileName);
```
msevestre: This method will split a population from a file in multiple files. Exactly what is needed for // in R. No need to implement it twice

tests/OSPSuite.Core.Tests/Domain/IndividualValuesCacheSpecs.cs
```diff
@@ -46,35 +47,87 @@ public class When_merging_an_individual_properties_cache_with_another_one : conc
...
+      public void should_merge_the_ids_together()
+      {
+         sut.IndividualIds.ShouldOnlyContainInOrder(1, 2, 3, 10, 20);
+      }
+
       [Observation]
       public void should_add_the_covariates_from_the_properties_cache()
       {
-         A.CallTo(() => _originalCovariates.Merge(_covariatesToMerge)).MustHaveHappened();
+         sut.CovariateValuesCache.AllCovariateValues.Count.ShouldBeEqualTo(2);
```
msevestre: Real tests instead of useless mocking

src/OSPSuite.R/Services/PopulationTask.cs
```diff
@@ -56,14 +71,42 @@ public DataTable PopulationTableFrom(IndividualValuesCache population, IModelCor
...
-         //add individual ids column
-         individualIds.Each(i => dataTable.Rows.Add(dataTable.NewRow()));
-         addColumnValues(dataTable, Constants.Population.INDIVIDUAL_ID_COLUMN, individualIds);
+         for (int i = 0; i < numberOfCores; i++)
+         {
+            var outputFile = Path.Combine(outputFolder, $"{outputFileName}_{i + 1}{Constants.Filter.CSV_EXTENSION}");
+            var rowIndices = dataSplitter.GetRowIndices(i).ToList();
+
+            //This is potentially empty if the number of individuals in the population is less than the number of cores provided
+            if(!rowIndices.Any())
```
Yuri05: I hope R population runner will not crash if it will not find its input population file :)

src/OSPSuite.R/Services/PopulationTask.cs
```diff
@@ -56,14 +71,42 @@ public DataTable PopulationTableFrom(IndividualValuesCache population, IModelCor
...
+               continue;
```
Yuri05: guess it could be `break `instead of `continue `here. 
Because if `rowIndices(i)` is empty, then `rowIndices(i+1)` etc. will be empty as well, right?

src/OSPSuite.R/Services/PopulationTask.cs
msevestre: This is somewhat of a stupid example. Why woyld anyone run in // when the problem is small

src/OSPSuite.R/Services/PopulationTask.cs
msevestre: yep.

src/OSPSuite.R/Services/PopulationTask.cs
Yuri05: could happen even with bigger pops.
E.g. if you have 960 Individuals and 50 cores we will end up with 20 Individuals per cores and thus the last 2 cores will have nothing todo.

msevestre: hum yeah ok, Whatever then :)

msevestre:
Merging for now. I'll probably have to get back to it once r is integrated

PR #703, opened 2020-01-23
Fixes #702 pi feedback

### LINKED ISSUE #702: PI: Visual Feedback shows identical values for Best and Current
In the Viesual Feedback of a PI, the values of parameters shown in the columns "Best" and "Current" are always equal, though the error might be different.

image

msevestre:


msevestre:
@Yuri05 Going to merge this for now. Please review. THis fixes the issue for me with MC

PR #705, opened 2020-01-27
Fixes export SA Parameters

msevestre:


msevestre:
@Yuri05 I am going to merge this so that I can update OSPSuite-R and have abdullah update his SA code. 
Please review

src/OSPSuite.Infrastructure.Import/Services/PKParameterSensitivitiesImporter.cs
```diff
@@ -0,0 +1,73 @@
...
+            return Array.Empty<PKParameterSensitivity>();
+         }
+         catch (Exception e)
+         {
+            logger.AddError(e.FullMessage());
+            return Array.Empty<PKParameterSensitivity>();
+         }
+      }
+
+      private IReadOnlyList<PKParameterSensitivity> pkParameterSensitivitiesFrom(string fileFullPath, char delimiter)
```
msevestre: Very simple import. Same logic as for importing results

src/OSPSuite.Infrastructure.Import/Services/SensitivityAnalysisRunResultsImportTask.cs
```diff
@@ -0,0 +1,141 @@
...
+using OSPSuite.Core.Domain.Services;
+using OSPSuite.Core.Services;
+using OSPSuite.Utility.Events;
+using OSPSuite.Utility.Extensions;
+
+namespace OSPSuite.Infrastructure.Import.Services
+{
+   public interface ISensitivityAnalysisRunResultsImportTask
+   {
+      Task<SensitivityAnalysisRunResultsImport> ImportResults(IModelCoreSimulation simulation, IReadOnlyCollection<string> files, CancellationToken cancellationToken, bool showImportProgress = true);
```
msevestre: Can import SARunResults from multiple files. This was missing for the SA workflow

src/OSPSuite.Infrastructure.Import/Services/SensitivityAnalysisRunResultsImportTask.cs
```diff
@@ -0,0 +1,141 @@
...
+
+      private void addPKParameterSensitivitiesFromSingleFile(SensitivityAnalysisRunResultsImport sensitivityAnalysisRunResultsImport, PKParameterSensitivitiesImport pkParameterSensitivitiesImport)
+      {
+         var sensitivityAnalysisRunResult = sensitivityAnalysisRunResultsImport.SensitivityAnalysisRunResult;
+         sensitivityAnalysisRunResultsImport.PKParameterSensitivitiesImportFiles.Add(pkParameterSensitivitiesImport.PKParameterSensitivitiesImportFile);
+
+         foreach (var pkParameterSensitivity in pkParameterSensitivitiesImport.PKParameterSensitivities)
+         {
+            if (sensitivityAnalysisRunResult.HasPKParameterSensitivityWithId(pkParameterSensitivity.Id))
+               sensitivityAnalysisRunResultsImport.AddError(Error.DuplicatedPKParameterSensitivityFor(pkParameterSensitivity.Id));
```
msevestre: We error here if there is two entries for the smae QuantityPath, PKParameter, Parameter combo. This should never happen

src/OSPSuite.Infrastructure.Import/Services/SimulationResultsImporter.cs
```diff
@@ -33,44 +32,36 @@ public SimulationResultsImportTask(IEntitiesInSimulationRetriever quantitiesRetr
...
+               tasks.Remove(firstFinishedTask);
+
+               // Await the completed task. 
+               allImportedResults.Add(await firstFinishedTask);
+               progressUpdater.IncrementProgress();
             }
-         }
-         finally
-         {
-            _allQuantities.Clear();
```
msevestre: no need for this guy here global. It's used once, and created where it's used

src/OSPSuite.R/Services/ContainerTask.cs
```diff
@@ -18,6 +18,7 @@ public interface IContainerTask
       IContainer[] AllContainersMatching(IModelCoreSimulation simulation, string path);
       IQuantity[] AllQuantitiesMatching(IModelCoreSimulation simulation, string path);
       IMoleculeAmount[] AllMoleculesMatching(IModelCoreSimulation simulation, string path);
+      IParameter[] AllSensitivityAnalysisParametersMatching(ISimulation simulation, string path);
```
msevestre: I have added this one here that returns all SA parameters with a  wild card
Not useful for reporting engine but probably neat when checking only the influence of parameters withing a specifc organ for example

src/OSPSuite.R/Services/SensitivityAnalysisTask.cs
```diff
@@ -0,0 +1,73 @@
...
+      /// <returns>An array of parameter path</returns>
+      string[] PotentialVariableParameterPathsFor(ISimulation simulation);
+
+      /// <summary>
+      ///    Exports sensitivity analysis run results to the csv file with path <paramref name="csvFile" />
+      /// </summary>
+      /// <param name="sensitivityAnalysisRunResult">Sensitivity analysis run result to export</param>
+      /// <param name="simulation">Simulation used</param>
+      /// <param name="csvFile">Full path of csv file where results should be exported</param>
+      void ExportResultsToCSV(SensitivityAnalysisRunResult sensitivityAnalysisRunResult, ISimulation simulation, string csvFile);
```
msevestre: All potential parameters, export and import nwo available in R

src/OSPSuite.Infrastructure.Import/Services/PKParameterSensitivitiesImportFile.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace OSPSuite.Infrastructure.Import.Services
+{
+   public class PKParameterSensitivitiesImportFile : ImportLogger
+   {
+      
+   }
+}
```
Yuri05: ? What is this class for?

src/OSPSuite.R/Services/ContainerTask.cs
Yuri05: Will this return ALL parameters matching the path, or only meaningful (used in model, non-categorical etc.)?

src/OSPSuite.Infrastructure.Import/Services/PKParameterSensitivitiesImportFile.cs
msevestre: It is used just to be consistent with all other files import where we have a dedicated class. I can remove it if you prefer.

src/OSPSuite.R/Services/ContainerTask.cs
msevestre: yes. I am going to rename it to AllParametersForSensitivityAnalysisMatching. This makes more sense I believe

msevestre: >Will this return ALL parameters matching the pat 

No that would not make sense. This is AllParametersMatching

PR #708, opened 2020-01-28
Fixes #707 add indexes unit

### LINKED ISSUE #707: Support hasUnit and ConvertValue with unit index

msevestre:


msevestre:
@Yuri05 For R, because of this crazy unicode issues, I will validate the unit IN R and then retrieve the index to perform conversion in .NET

src/OSPSuite.R/Extensions/WithDimensionExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+   public static class WithDimensionExtensions
+   {
+      /// <summary>
+      /// Converts the value given in the unit at index <paramref name="unitIndex"/> into base unit
+      /// </summary>
+      /// <param name="withDimension">Object with dimension used for conversion</param>
+      /// <param name="value">Value in unit defined at <paramref name="unitIndex"/></param>
+      /// <param name="unitIndex">1-based index of the unit in the units array</param>
+      /// <returns>The value converted in base unit</returns>
+      public static double ConvertToBaseUnit(this IWithDimension withDimension, double value, int unitIndex)
```
msevestre: These are the two methods specifically for R, hence in the R dll

PR #711, opened 2020-01-29
709 dimension per time

msevestre:


src/OSPSuite.Core/Converter/v5_2/DimensionMapper.cs
```diff
@@ -35,6 +35,8 @@ public interface IDimensionMapper
 
    internal class DimensionMapper : IDimensionMapper
    {
+      private const string RHS_DIMENSION_SUFFIX = " per Time";
```
msevestre: @Yuri05 I would vote to remove all those conversions here. Trying to convert pkml from version 52 becomes ridiculous
I think we should support starting 6.3.1

src/OSPSuite.Core/Converter/v9/Converter730To90.cs
```diff
@@ -0,0 +1,90 @@
...
+            }
+         }
+      }
+
+      private string mapDimensionName(string dimensionName)
+      {
+         if(_dimensionFactory.Has(dimensionName))
+            return dimensionName;
+
+         if (dimensionName.IsOneOf($"{Constants.Dimension.DIMENSIONLESS} per Time", $"{Constants.Dimension.DIMENSIONLESS} per time", "Fraction per time", "Fraction per Time"))
```
msevestre: @pavelbal @Yuri05 
Here, I am trying convert some of the stupid dimensions to the existing one. 
Do we have anything else? in short, `dimensionless per time` and `fraction per time` will be converted to `Inversed time`

src/OSPSuite.Core/Serialization/PKMLVersion.cs
```diff
@@ -13,6 +13,7 @@ public static class PKMLVersion
       public const int V6_3_1 = 11;
```
msevestre: Anything below 631 should be removed in my opinion

src/OSPSuite.UI/Controls/UxTreeView.cs
```diff
@@ -64,6 +64,7 @@ public UxTreeView()
          ToolTipForNode = node => node.ToolTip;
          ToolTipController = new ToolTipController();
          ToolTipController.GetActiveObjectInfo += (o, e) => ShowToolTip(e);
+         OptionsMenu.ShowExpandCollapseItems = false;
```
msevestre: this is a breaking change in new devexpress. This ensures old behavior everywhere

PR #712, opened 2020-01-29
Try moving rhs to core

msevestre:


msevestre:
@Yuri05 I think RHS Dimension should be created dynamically upon retrieval. At the moment, they are stored in the MoBi project and makes transfer to PKSim impossible depending on the dimension

With this first try, the idea is to recreate the RHS upon project load.

src/OSPSuite.Core/Converter/v5_2/SerializationDimensionFactory.cs
```diff
@@ -27,6 +27,11 @@ public SerializationDimensionFactory(IDimensionFactory dimensionFactory, IDimens
          _allDimensions.AddRange(dimensionMapper.DummyDimensionsForConversion);
       }
 
+      public bool TryGetDimension(string dimensionName, out IDimension dimension)
```
msevestre: This whole class should go once we agree that we do not want to support such outdated projects

src/OSPSuite.Core/Domain/UnitSystem/DimensionFactory.cs
```diff
@@ -24,18 +29,37 @@ public void AddDimension(IDimension dimension)
          _dimensions.Add(dimension.Name, dimension);
       }
 
-      public IEnumerable<IDimension> Dimensions => _dimensions;
-
-      public IEnumerable<string> DimensionNames => _dimensions.Keys;
-
       public IDimension Dimension(string name)
```
msevestre: Same behavior as before. It throws an exception if the dimension by name is not found

src/OSPSuite.Core/Domain/UnitSystem/DimensionFactory.cs
```diff
@@ -24,18 +29,37 @@ public void AddDimension(IDimension dimension)
...
       {
          if (!Has(name))
-            throw new KeyNotFoundException("Dimension " + name + " not available in DimensionFactory.");
+            throw new KeyNotFoundException(unknownDimension(name));
 
          return _dimensions[name];
       }
 
+      public bool TryGetDimension(string dimensionName, out IDimension dimension)
```
msevestre: THis is slightly new. Here we return the dimension if it exists by name
If not, this is potentially something like `Hydraulic conductivity per time` for which we have no dimension. THis was created automatically by MoBi. What we do is that we try to get it based on the name. If we succeed, all good, we return this one. AT that moment, the RHS Dimension was added to the lsit of dimension and is available during life time of the project

src/OSPSuite.Core/Domain/UnitSystem/DimensionFactory.cs
```diff
@@ -24,18 +29,37 @@ public void AddDimension(IDimension dimension)
...
+         //It might be a RHS dimension. Let see if we can retrieve it
+
+         try
+         {
+            dimension = getOrAddRHSDimensionForName(dimensionName);
+            return true;
+         }
+         catch (KeyNotFoundException)
+         {
+            dimension = null;
```
msevestre: not found => null and false

src/OSPSuite.Core/Domain/UnitSystem/DimensionFactory.cs
```diff
@@ -55,6 +79,72 @@ public void Clear()
 
       public IDimension NoDimension => _dimensions[Constants.Dimension.DIMENSIONLESS];
 
+      private string rhsDimensionName(IDimension dimension) => $"{dimension.Name}{Constants.Dimension.RHS_DIMENSION_SUFFIX}";
+
+      private string rhsDefaultUnitName(IDimension dimension)
```
msevestre: Code taken from Mobi

src/OSPSuite.Core/Domain/UnitSystem/DimensionFactory.cs
```diff
@@ -55,6 +79,72 @@ public void Clear()
...
+         var equivalentRHSDimension = findFirstEquivalentDimension(rhsDimension, unitName);
+         if (equivalentRHSDimension != null)
+            return equivalentRHSDimension;
+
+         // Equivalent dimensions does nto exist. We add it and return it
+         AddDimension(rhsDimension);
+         return rhsDimension;
+      }
+
+      public IDimension DimensionForUnit(string unitName)
```
msevestre: I have also moves this guy here. Why not

src/OSPSuite.Core/Domain/UnitSystem/DimensionFactory.cs
```diff
@@ -55,6 +79,72 @@ public void Clear()
...
+      private IDimension getOrAddRHSDimensionForName(string rhsDimensionName)
+      {
+         if (string.IsNullOrEmpty(rhsDimensionName))
+            throw new KeyNotFoundException();
+
+         if (!rhsDimensionName.Contains(Constants.Dimension.RHS_DIMENSION_SUFFIX))
+            throw new KeyNotFoundException(unknownDimension(rhsDimensionName));
+
+         var dimensionName = rhsDimensionName.Replace(Constants.Dimension.RHS_DIMENSION_SUFFIX, "");
+         var originDimension = Dimension(dimensionName);
```
msevestre: This will throw if the dimension is not found

src/OSPSuite.Core/Serialization/Xml/SerializationContext.cs
```diff
@@ -87,6 +87,10 @@ public DataColumnCollectionReference AddDataCollectionReference(ICache<Auxiliary
 
       public IDimension DimensionByName(string dimensionName)
       {
+         var found = DimensionFactory.TryGetDimension(dimensionName, out var dimension);
```
msevestre: This is the key difference here. Instead of throwing the exception if not found, we try to get it first

src/OSPSuite.Core/Serialization/Xml/SerializationContext.cs
```diff
@@ -87,6 +87,10 @@ public DataColumnCollectionReference AddDataCollectionReference(ICache<Auxiliary
...
+         if (found)
+            return dimension;
+
          return DimensionFactory.Dimension(dimensionName);
       }
```
msevestre: If not found, same behavior as before (which is going to throw an exception)

Yuri05: The last return statement `return DimensionFactory.Dimension(dimensionName);` will always throw an exception.
Why not 
```
if (found)
  return dimension;
else
 throw new XXXException

```?

msevestre: Because I don't want to know what exception to throw. If I leave the same code as before, nothing will change. Same exception, same message etc..

PR #722, opened 2020-02-05
Fixes allow update of path instance

msevestre:
@Yuri05 REquired for https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1320


src/OSPSuite.Core/Domain/ObjectPath.cs
```diff
@@ -213,6 +220,12 @@ public void RemoveFirst()
          RemoveAt(0);
       }
 
+      public void ReplaceWith(IEnumerable<string> pathEntries)
```
msevestre: need to be able to swap object path in place

PR #724, opened 2020-02-05
723 wrong mapping windows key

msevestre:


msevestre:
@Yuri05 One learns every day. Flag enum can only have 32 entries. So in our case, because I had defined more, F1 was mapped to R!!

I just copied all the values from the widows enum as is

tests/OSPSuite.UI.Tests/KeysToWindowsKeysMapperSpecs.cs
```diff
@@ -20,6 +20,46 @@ public class When_mapping_some_keys_to_windows_keys : concern_for_KeysToWindowsK
       public void should_return_the_expected_keys()
       {
          sut.MapFrom(Keys.A).ShouldBeEqualTo(WindowsKey.A);
+         sut.MapFrom(Keys.B).ShouldBeEqualTo(WindowsKey.B);
```
msevestre: Testing everything here!

PR #729, opened 2020-02-08
Fixes #728 Fix cpp export

### LINKED ISSUE #728: Fix cpp export

Yuri05:


src/OSPSuite.Core/Serialization/SimModel/Services/SimModelExporter.cs
```diff
@@ -52,6 +52,9 @@ private void exportToCode(IModelCoreSimulation modelCoreSimulation, string outpu
          var simModelXml = ExportSimModelXml(modelCoreSimulation, SimModelExportMode.Full);
          var simModelSimulation = new Simulation();
 
+         if (codeExportLanguage == CodeExportLanguage.Cpp)
+            simModelSimulation.Options.KeepXMLNodeAsString = true;
```
Yuri05: For some weird reason cpp export relies on the original XML used for simulation creation. Per default XML string used to setup a SimModel-simulation is thrown away after simulation.Load. With the line above it's kept and can be accessed by the cpp export

tests/OSPSuite.Core.IntegrationTests/SimModelExporterSpecs.cs
```diff
@@ -173,11 +173,7 @@ protected override void Context()
 
       private void checkFiles(string path)
       {
-         //TODO
-         var cppFiles = new string[]
-         {
-            
-         };
+         var cppFiles = new[] {"Standard.cpp"};
```
Yuri05: at the moment exported cpp model is always named "Standard.cpp"

Yuri05:
The test in `When_exporting_a_simulation_as_cpp_code` still fails. For some reason a "mini-PBPK" model used in this test makes trouble for the cpp-Exporter. Trying to export as Values crashes; trying to export as Formulas ends up in a dead loop.

However with the fix in this PR "standard" PBPK models can be exported without problems.
E.g. an IV simulation from the Theophylline example is exported like this: 
[Theophylline_Kaumeier_IV_100_mg.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/4175184/Theophylline_Kaumeier_IV_100_mg.zip)

Yuri05:
Exporting full PBPK model takes quite a while (several minutes).
When we offer this feature in non-dev mode, we should do it the same way as pdf export: message at the beginning "Export started in the background" and message at the end. Will create a separate issues for this

src/OSPSuite.Core/Serialization/SimModel/Services/SimModelExporter.cs
msevestre: I do not understand what this line does, even with your comments

msevestre: I'd suggest to add the explanation in the code because it's just too weird

Yuri05: done

PR #730, opened 2020-02-10
576 remove curve check

msevestre:


Yuri05:
@msevestre I assume this PR is the fix for https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1109 ?

PR #732, opened 2020-02-14
Fixes #731 explicit Dispose for Simmodel-simulation

### LINKED ISSUE #731: explicit Dispose for Simmodel-simulation

Yuri05:


src/OSPSuite.Core/Domain/Services/SimModelManager.cs
```diff
@@ -59,6 +59,7 @@ public SimulationRunResults RunSimulation(IModelCoreSimulation simulation, Simul
          }
          finally
          {
+            _simModelSimulation?.Dispose();
```
Yuri05: eliminates memory leaks for Individual simulations

src/OSPSuite.Core/Domain/Services/SimModelBatch.cs
```diff
@@ -88,6 +88,7 @@ public double[] SensitivityValuesFor(string outputPath, string parameterPath)
       public void Clear()
       {
          _modelCoreSimulation = null;
+         _simModelSimulation?.Dispose();
```
Yuri05: @msevestre memory leaks for Pop simulations still present. Is `SimModelBatch.Clear()` called in PK-sim after pop calculations?

msevestre:
@Yuri05 Can you create a branch here as opposed to on your account? I'd like to add the missing pieces for population simulaiton. It is not called indeed because we do not use the SimModelBatch. A Simulation is created and discarted. Therefor disposed is never called

PR #745, opened 2020-03-03
Fixes #739 Remove FuncParserException

### LINKED ISSUE #739: Remove FuncParserException
and replace all usages with `OSPSuiteException`
This together with https://github.com/Open-Systems-Pharmacology/OSPSuite.FuncParser/issues/37 will fix https://github.com/Open-Systems-Pharmacology/MoBi/issues/453

@msevestre Or do you see any reason why we would need `FuncParserException`?

Yuri05:


src/OSPSuite.Core/Services/ExplicitFormulaParser.cs
```diff
@@ -46,13 +46,13 @@ public double Compute(double[] variableValues, double[] parameterValues)
             _parserFunc.SetParameterValues(parameterValues);
             return _parserFunc.CalcExpression(variableValues);
          }
-         catch (FuncParserException)
+         catch (OSPSuiteException ex)
          {
-            throw;
+            throw new FuncParserException(ex.Message,ex);
```
Yuri05: ok, keeping FuncParserException for now. But the FuncParser itself throws OSPSuiteException in case of (handled) errors - so it must be catched explicitly.

PR #749, opened 2020-03-14
Fixes #748 user defined pk parameters

### LINKED ISSUE #748: Allow for better definitiion of UserDEfined PK Parmaeters

msevestre:


src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -350,14 +353,14 @@ public static class Distribution
 
       public static class PKParameters
       {
-         internal const string FirstIntervalSuffix = "_t1_t2";
-         internal const string LastIntervalSuffix = "_tLast_tEnd";
+         internal const string FirstIntervalSuffix = "_tD1_tD2";
```
msevestre: based on rename from the parameter file

src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -328,8 +329,10 @@ public static class SimulationResults
          public const string PARAMETER = "Parameter";
          public const string VALUE = "Value";
          public const string UNIT = "Unit";
+         public const string DISPLAY = "Display";
```
msevestre: new exported value so that RE can simply read the column

src/OSPSuite.Core/Domain/PKAnalyses/PKParameter.cs
```diff
@@ -31,5 +32,11 @@ public virtual string DisplayName
       public virtual IDimension Dimension { get; set; }
 
       public virtual PKParameterMode Mode { get; set; }
+
+      public virtual Unit DisplayUnit
```
msevestre: THis will only be set from R. Same behavior as before otherwise by using the default unit of a dimension

src/OSPSuite.Core/Domain/SensitivityAnalyses/SensitivityAnalysis.cs
```diff
@@ -73,22 +70,6 @@ public virtual void RemoveSensitivityParameter(SensitivityParameter sensitivityP
          HasChanged = true;
       }
 
-      public virtual void AddDynamicPKParameter(DynamicPKParameter dynamicPKParameter)
```
msevestre: UserDefinedPKPArameter are now added globally and not per SA.. Much more ergnomic

src/OSPSuite.R/Domain/RLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.Extensions.Logging;
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Services;
+using OSPSuite.Infrastructure.Import.Services;
+using OSPSuite.Utility.Extensions;
+using ILogger = OSPSuite.Core.Services.ILogger;
+
+namespace OSPSuite.R.Domain
+{
+   public class RLogger : ILogger
```
msevestre: I have also added a basic logger that we will extend later on

src/OSPSuite.R/Services/DimensionTask.cs
```diff
@@ -73,6 +85,44 @@ public double[] ConvertToUnit(string dimensionName, string targetUnit, double va
 
       public string[] AllAvailableDimensionNames() => _dimensionFactory.DimensionNames.OrderBy(x => x).ToArray();
 
+      public IDimension CreateUserDefinedDimension(string unit)
+      {
+         return new Dimension(Constants.Dimension.NO_DIMENSION.BaseRepresentation, unit, unit);
+      }
+
+      public IDimension DimensionForStandardPKParameter(int standardPKParameterValue)
```
msevestre: The idea here is to be able to retrieve the deafult dimenson for a user defined pk parameter just based on the base paraemter used. e.g. AUC t4 t5 wil have the same dim as AUC automatically

Yuri05: How should we deal with normalized parameters (XXX/Dose and XXX/Dose/BW)?
Theoretically in this case the dimension must be set by user along with the normalization factor?

msevestre: I think we need to add a _norm so that the default dimension is retrieve automatically
Otherwise the user can specify a unit. THe core will try to find the dimension that matches the unit. If this is not successful, then a dimension will be created with only ONE unit. I am not sure what cases we will hvae but this should cover most of them... Let see during our tests

PR #754, opened 2020-03-24
Dimension convrter

msevestre:


src/OSPSuite.Core/Converters/v9/Converter730To90.cs
```diff
@@ -90,5 +100,51 @@ private string mapDimensionName(string dimensionName)
...
+      {
+         sensitivityAnalysisRunResult?.AllPKParameterSensitivities.Each(x => { x.PKParameterName = ConvertPKParameterName(x.PKParameterName); });
+      }
+
+      public string ConvertPKParameterName(string pkParameterName)
+      {
+         return _pkParameterNameMapping.Contains(pkParameterName) ? _pkParameterNameMapping[pkParameterName] : pkParameterName;
+      }
+
+      private static Cache<string, string> createPKParameterNameMapping()
```
msevestre: @Yuri05 This will need to be called by PKSim and MoBi. For PKSim, also for Population results. That's why I have a public method available in the converter

PR #758, opened 2020-03-27
756 dyamically retrieve mw

msevestre:


msevestre:
@Yuri05 There are many places where we have to consolidate. I don't think those places where much in use so we are probably good

1/ Import of PK Analysis Data. We need to ensure that the data that were exported (potentially in mass unit are converted back to our core unit). This may not be 100% required but it can only lead to issue if we have mixed dimensions and values


2/ Export of PK Analysis to csv. WE need to ensure that we are exporting using a merge dimension

3/ Add capability to convert Single PK Parameter from one unit to the other

src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -294,7 +294,7 @@ public static class Dimension
          public const string MOLAR_CONCENTRATION = "Concentration (molar)";
          public const string MASS_CONCENTRATION = "Concentration (mass)";
          public const string DIMENSIONLESS = "Dimensionless";
-         public const string AMOUNT = "Amount";
+         public const string MOLAR_AMOUNT = "Amount";
```
msevestre: THis is nicer when matching MASS To MOLAR. All molar starts with Molar, all MASS start with MASS

src/OSPSuite.Core/Domain/Model.cs
```diff
@@ -59,6 +48,12 @@ public IContainer Neighborhoods
          return molWeightParameter?.Value;
       }
 
+      public double? MolWeightFor(string quantityPath)
```
msevestre: Load a molweight for a given quantity by path or return null

src/OSPSuite.Core/Domain/QuantityPKParameterContext.cs
```diff
@@ -0,0 +1,22 @@
+﻿using OSPSuite.Core.Domain.UnitSystem;
+
+namespace OSPSuite.Core.Domain
+{
+   public class QuantityPKParameterContext: IWithDimension
```
msevestre: The context is a trick we use to be able to create a converter for the QuantityPKPArameter

src/OSPSuite.Core/Domain/UnitSystem/DimensionFactory.cs
```diff
@@ -124,6 +124,11 @@ public IDimension DimensionForUnit(string unitName)
          return matches.FirstOrDefault(x => x.Units.Any(unit => string.Equals(unitName, unit.Name, StringComparison.Ordinal))) ?? matches.First();
       }
 
+      public IDimension CreateUserDefinedDimension(string dimensionName, string unit)
```
msevestre: Theoretically, we will export PK PArameter in dimension that are unknown. Therefore I moved this definition in the dimensionFactory so that we can import object in an unknown dimension. This should only be used from R. But this might show up in PKSim when importing results

src/OSPSuite.Core/Serialization/Xml/QuantityPKParameterXmlSerializer.cs
```diff
@@ -0,0 +1,15 @@
+﻿using OSPSuite.Core.Domain;
+
+namespace OSPSuite.Core.Serialization.Xml
+{
+   public class QuantityPKParameterXmlSerializer : OSPSuiteXmlSerializer<QuantityPKParameter>
```
msevestre: This was still in PKSim

src/OSPSuite.Infrastructure.Import/Services/SimulationPKAnalysesImporter.cs
```diff
@@ -36,60 +38,70 @@ public SimulationPKAnalysesImporter(IDimensionFactory dimensionFactory)
...
+      public IEnumerable<QuantityPKParameter> ImportPKParameters(string fileFullPath, IModelCoreSimulation modelCoreSimulation, IImportLogger logger)
       {
          try
          {
+            _logger = logger;
             _importedPK = new Cache<string, QuantityPKParameter>(x => x.Id);
-            //cache containing a list of tupe<individual Id, value in core unit>
-            _valuesCache = new Cache<QuantityPKParameter, List<Tuple<int, float>>>();
+            //cache containing a list of tuple<individual Id, value in core unit>
+            _valuesCache = new Cache<QuantityPKParameter, List<(int individualId, float value)>>();
```
msevestre: just a nicer way to write named tuples

src/OSPSuite.Infrastructure.Import/Services/SimulationPKAnalysesImporter.cs
```diff
@@ -101,34 +113,56 @@ private double convertValueToCoreValue(IDimension dimension, double valueInUnit,
...
-         {
-            var dimension = findDimensionFor(csv[UNIT]);
-            var pkParameter = new QuantityPKParameter {Name = parameterName, QuantityPath = quantityPath, Dimension = dimension};
-            _importedPK.Add(pkParameter);
-         }
+         if (_importedPK.Contains(id))
+            return _importedPK[id];
+
+
+         var dimension = findDimensionFor(parameterName, csv[UNIT]);
```
msevestre: This dimension will be retrieved based on the unit. So something like mg => MASS
But we need to save the value in Amount. 
So we create a merge DImension "AMOUNT, MASS" so that the value will be converted from mg to umol

src/OSPSuite.Infrastructure.Import/Services/SimulationPKAnalysesImporter.cs
```diff
@@ -101,34 +113,56 @@ private double convertValueToCoreValue(IDimension dimension, double valueInUnit,
...
 
-         if (dimension != null)
-            return dimension;
+         if (dimension == null)
+         {
+            _logger.AddWarning(Error.CouldNotFindDimensionWithUnit(unit));
+            return _dimensionFactory.CreateUserDefinedDimension(parameterName, unit);
+         }
 
+         // for a mass dimension, we will return the Molar dimension to ensure that we return the value of the PK-Parameter in the base supported unit
```
msevestre: We always want the SOURCE DIMENSION of a merge dimension to be molar. Otherwise the base unit would be mg, thus mixing and matching

src/OSPSuite.Core/Domain/PKAnalyses/PKParameter.cs
```diff
@@ -33,10 +33,10 @@ public virtual string DisplayName
 
       public virtual PKParameterMode Mode { get; set; }
 
-      public virtual Unit DisplayUnit
+      public virtual string DisplayUnitName
```
msevestre: In order to support any possible units, we need to make sure we have a string here and not a unit. If this does not exist, we will export values in their default unit

src/OSPSuite.Core/Domain/Services/PopulationSimulationPKAnalysesToDataTableConverter.cs
```diff
@@ -0,0 +1,85 @@
...
+            var quantityPath = quantityPKParameter.QuantityPath;
+            var molWeight = simulation.MolWeightFor(quantityPath);
+            var pkParameter = _pkParameterRepository.FindByName(quantityPKParameter.Name);
+            var quantityPKParameterContext = new QuantityPKParameterContext(quantityPKParameter, molWeight);
+            var mergedDimension = _dimensionFactory.MergedDimensionFor(quantityPKParameterContext);
+            var unit = mergedDimension.UnitOrDefault(pkParameter.DisplayUnitName);
+            quantityPKParameter.Values.Each((value, index) =>
+            {
+               var row = dataTable.NewRow();
+               row[INDIVIDUAL_ID] = index;
```
msevestre: @Yuri05 This might be wrong here. We are using the index instead of individualId. This was always like this so this is not new but we should think about what it means

msevestre:
Going to merge for now. I may have a follow up PR once I integrate it with R if something is missing

PR #760, opened 2020-04-01
add Clear RHS Formula to parameter

msevestre:


src/OSPSuite.Core/Domain/Parameter.cs
```diff
@@ -87,6 +87,11 @@ public interface IParameter : IQuantity, IWithDefaultState
       ///    value of the parameter are taken
       /// </summary>
       double RandomDeviateIn(RandomGenerator randomGenerator, double? min = null, double? max = null);
+
+      /// <summary>
+      /// Sets the RHS Formula to NULL. This is required for R-Only in order to be able to set the RHS formula to NULL
+      /// </summary>
+      void ClearRHSFormula();
```
msevestre, line 94: @Yuri05 Required because I cannot set a value to NULL from R. (e.g. it cannot cast a NULL from R to a NULL from .NET)

PR #761, opened 2020-04-06
Fixes pk analysis calculation bug for AUC_tD1_tD2

msevestre:


src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -391,7 +391,7 @@ internal static string Create(string param, string suffix = null)
          public static readonly string AUC_inf_tD1 = Create(AUC_inf, "_tD1");
          public static readonly string AUC_inf_tD1_norm = NormalizedName(AUC_inf_tD1);
 
-         public static readonly string AUC_tD1_tD2 = Create(AUC_tEnd, FirstIntervalSuffix);
+         public static readonly string AUC_tD1_tD2 = Create("AUC", FirstIntervalSuffix);
```
msevestre, line 394: This was creating AUC_tEnd_tD1_tD2

Now I am testing that the names are also in the PKParameter.xml files!

PR #766, opened 2020-04-17
WIP #765 ensure that we can support NaN in population table

msevestre:


src/OSPSuite.Core/Domain/Services/PopulationDataSplitter.cs
```diff
@@ -115,8 +116,15 @@ private void fillParameterAndInitialValuesFor(int individualId, IReadOnlyList<Pa
...
             // if no aging data found, parameter value is stored in the population data table
             // (single value)
-            if (!parameterProperties.TablePoints.Any())
-               parameterProperties.Value = nonTableParameterValues[parameterProperties.Path].ConvertedTo<double>();
+            if (parameterProperties.TablePoints.Any())
+               continue;
+            
+            var parameterValue = nonTableParameterValues[parameterProperties.Path].ConvertedTo<double>();
+            //reset the parameter to its default value if the value in the table is NaN
+            if (double.IsNaN(parameterValue))
```
msevestre, line 124: If the value is NaN, using the default value instead

msevestre:
@Yuri05 Going to merge to update R so that @pchelle can start working

PR #771, opened 2020-05-04
Fixes #770 copy clipboard issue

### LINKED ISSUE #770: Copy to clipboard of graph broken on screen with 150% scaling
see https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1473

msevestre:
@Yuri05 This is a fix for the copy to clipboard issue

+: It works on all screen scaling 
+: It does not use that hacky code to copy stuff to clipboard

-: It is not emf so if someone needs to make the picture bigger or smaller, it might look not as good. But we have the Export options anyways so in my eyes ok

### DIFF
--- a/src/OSPSuite.UI/Controls/UxChartControl.cs
+++ b/src/OSPSuite.UI/Controls/UxChartControl.cs
@@ -2,12 +2,12 @@ using System;
 using System.Drawing;
 using System.Drawing.Imaging;
 using System.IO;
+using System.Windows.Forms;
 using DevExpress.Utils;
 using DevExpress.XtraBars;
 using DevExpress.XtraCharts;
 using OSPSuite.Assets;
 using OSPSuite.Core.Services;
-using OSPSuite.UI.Services;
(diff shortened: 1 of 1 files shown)


src/OSPSuite.UI/Controls/UxChartControl.cs
```diff
@@ -84,10 +83,11 @@ public virtual void CopyChartToClipboard(ChartControl chartControl)
       {
          using (var ms = new MemoryStream())
          {
-            chartControl.ExportToImage(ms, ImageFormat.Emf);
+            chartControl.ExportToImage(ms, ImageFormat.Png);
             ms.Seek(0, SeekOrigin.Begin);
 
-            _clipboardTask.PutEnhMetafileOnClipboard(this, new Metafile(ms));
```
msevestre, line 90: This throws a bunch of excpetion too (GDI stuff etc..). I would get rid off it completely but it's used in the wj export..lol

PR #774, opened 2020-05-08
WIP: 773 add designer back to UI. Still lots to do!

msevestre:
This does not work 100% unfortunately

Following
https://github.com/dotnet/winforms/blob/master/Documentation/winforms-designer.md

Option 1 seems easier but there are issues at the moment

Going to https://github.com/dotnet/winforms/blob/master/Documentation/winforms-designer.md#option-2 seems like a massive pain


msevestre:
After installing vs2019 preview...everything is working agaon

image

SO no need for this

PR #776, opened 2020-05-18
Fixes #371 Update dimension task to also support conversion to base unit

msevestre:


msevestre:
This to support the other conversion
However I don't think this is very useful. Without the quantity, you have no idea in which base dimension you want to convert (e.g. if the quantity is in amount but the unit you have is mass, you would convert the value in base unit from mass and not from amount which would be wrong)

This makes sense however for completion

PR #793, opened 2020-05-29
792 group info lost

msevestre:


src/OSPSuite.R/Bootstrap/ApplicationStartup.cs
```diff
@@ -105,13 +104,6 @@ private static void initializeDimensions(IContainer container)
          dimensionFactory.AddMergingInformation(new SimpleDimensionMergingInformation(aucMassDimension, aucMolarDimension));
       }
 
-      private static void initializeGroups(IContainer container)
-      {
-         var groupRepository = container.Resolve<IGroupRepository>();
-         var undefinedGroup = new Group {Name = Constants.Groups.UNDEFINED, Id = "0"};
```
msevestre, line 111: We only had one group before. that's why we only add 0

src/OSPSuite.R/MinimalImplementations/RGroupRepository.cs
```diff
@@ -0,0 +1,25 @@
...
+
+      public IEnumerable<IGroup> All() => _allGroups.Values;
+
+      public IGroup GroupByName(string groupName) => getOrAdd(groupName);
+
+      public IGroup GroupById(string groupId) => getOrAdd(groupId);
+
+      private IGroup getOrAdd(string groupIdOrName)
+      {
+         return _allGroups.GetOrAdd(groupIdOrName, x => new Group {Id = groupIdOrName, Name = groupIdOrName});
```
msevestre, line 19: Now I always have groups with same name and id. That way, all is consistent within the R run (id is what is used for serialization but name is what is used in the object model).

PR #802, opened 2020-06-12
Fixes #801 expose application parameters

### LINKED ISSUE #801: Expose Applicaton Parameters for the simulation
For now 
  public IParameter StartTime { get; }
         public IParameter InfusionTime { get; }
         public IParameter DrugMass { get; }

msevestre:


msevestre:
@Yuri05 This is the exact same code as before only moved to an extension method so that it can be exposed to R.

src/OSPSuite.Core/Domain/ApplicationParameters.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace OSPSuite.Core.Domain
+{
+   public class ApplicationParameters
```
msevestre, line 3: This is what we will return. StartTime may not be null but rest might be.
Obviously, application create by user that do not fit our schema won't be found here. But unlikely

src/OSPSuite.Core/Domain/ModelCoreSimulationExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+            return new List<IContainer>();
+
+         var allApplications = applicationEventGroup.SelectMany(x => x.GetAllChildren<IContainer>(c => c.ContainerType == ContainerType.Application))
+            .ToList();
+
+         return getApplicationsForAppliedAncestorMolecule(simulation.Reactions, moleculeName, allApplications);
+      }
+
+      private static IReadOnlyList<IContainer> getApplicationsForAppliedAncestorMolecule(IEnumerable<IReactionBuilder> reactions, string moleculeName,
+         IReadOnlyList<IContainer> allApplications)
```
Yuri05, line 35: ok, it's the same code as before and maybe I never looked at it... but I don't understand this function at all. 
* If my simulation has no reactions at all - it will return empty list and the calling function`allApplicationsForMolecule(IModelCoreSimulation simulation, string moleculeName)` will also return empty list - even if there are applications of `moleculeName`
* If I have a reaction A==>B then `allApplicationsForMolecule(simulation, "B") `will return Applications of A?

msevestre: @Yuri05 What are you talking about :)

The first call of the one and only public method is 

````
    var applicationsForAppliedAncestorMolecule =
            allApplications.Where(c => c.GetSingleChildByName<IMoleculeAmount>(moleculeName) != null).ToList();

         // If there are any applications of this molecule, use them
         if (applicationsForAppliedAncestorMolecule.Any())
            return applicationsForAppliedAncestorMolecule;
```

so if we found an application, we use it

Arguably, the test with reactions can be done later but a building block should NEVER be null. At most it should be empty

msevestre: If I have a reaction A==>B then allApplicationsForMolecule(simulation, "B") will return Applications of A?

NO only if there are no application producing B

So in short
* NO
* NO

Yuri05:
>building block should NEVER be null

ok, if `reactions `is never null, then it's fine.

>If I have a reaction A==>B then allApplicationsForMolecule(simulation, "B") will return Applications of A?
NO only if there are no application producing B

OK. Still no idea however what is the rationale behind and what the check for "only 1 educt" is for.

msevestre:
I am not quite sure where this is used. But the idea is. What application resulted in this molecule being created

It allows us to calculate those pk parameters also for metabolite.
If there is more than one educt, then it's a more complex reaction and it's not clear what application resulted indirectly in the metabolite being created. 

Edge case of course. And I didn't invent this so we probably specified it like that ;-))

PR #804, opened 2020-06-12
803 also retrieve quantity path

msevestre:


src/OSPSuite.Core/Domain/Model.cs
```diff
@@ -22,6 +19,17 @@ public interface IModel : IObjectBase, IMolWeightFinder
       ///    Returns the total drug mass parameter defined in the model for the <paramref name="moleculeName"/> if available or null otherwise
       /// </summary>
       IParameter TotalDrugMassFor(string moleculeName);
+
+      /// <summary>
+      /// Returns the molecule name associated with a quantity with path <paramref name="quantityPath"/>
+      /// </summary>
+      string MoleculeNameFor(string quantityPath);
```
msevestre, line 26: I am exposing these guys because it is quite useful to retrieve the molecule name for a quantity.
Especially I am not sure how the applications parameters will be retrieve and if a moleculename will be available

That way, it is symetrical to MolWeight

src/OSPSuite.Core/Domain/ModelCoreSimulationExtensions.cs
```diff
@@ -34,17 +45,18 @@ private static IReadOnlyList<IContainer> allApplicationsForMolecule(IModelCoreSi
...
-
-         var reactionsList = reactions.ToList();
          var applicationsForAppliedAncestorMolecule =
             allApplications.Where(c => c.GetSingleChildByName<IMoleculeAmount>(moleculeName) != null).ToList();
 
          // If there are any applications of this molecule, use them
          if (applicationsForAppliedAncestorMolecule.Any())
             return applicationsForAppliedAncestorMolecule;
 
+         if (reactions == null)
```
msevestre, line 55: @Yuri05 I moved the check with reactions below the first if. This is probably clearer even though this is equivalent as null reaction building block in a simulation is a problem

Yuri05: yep, that's more clear :)

src/OSPSuite.Core/Domain/Model.cs
```diff
@@ -66,6 +69,22 @@ public IContainer Neighborhoods
...
+      }
+
+      public string MoleculeNameFor(IQuantity quantity)
+      {
+         if (quantity == null)
+            return string.Empty;
+
+         return quantity.IsAnImplementationOf<IMoleculeAmount>() ?
+            quantity.Name :
+            quantity.ParentContainer?.Name;
```
Yuri05, line 85: if `quantity `is not a molecule amount: shouldn't we check that `quantity.ParentContainer` is a molecule amount before returning `quantity.ParentContainer?.Name`? 
If the function was called for a quantity which is e.g. not molecule related (plottable parameter for example), then parent container would be not a molecule.

Yuri05: This would be also interesting test for RE: add a `plottable parameter `as output and check what happens 😄

msevestre:
Yeah I agree. This would be a breaking change so I didn't want to do it
And in fact the container might be a molecule container which is not a molecule amount. Yet we want to retrieve the molecule name.


Therefore....good enough for now

Yuri05:
ok, created an issue for later: #805

PR #809, opened 2020-06-18
1465 remove warning

abdelr:


msevestre:
@abdelr I do not understand why this code was removed from CORE as this is a service that should also be used in MoBi

abdelr:
> @abdelr I do not understand why this code was removed from CORE as this is a service that should also be used in MoBi

As discussed in https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/1507 the code was duplicated in PKSim and we did not see any reference to it on MoBi. If we would like to use this in MoBi also then we should indeed move it to Core and remove it from PKSim. I would wait for the release though so we could build a new version of our nuget and use the changes in PKSim without any code duplication to avoid forgetting it again.

msevestre:
@abdelr  now that the release is behing us, please add everything back to core as we discussed 
(also renaming ILogger to IOSPLogger as suggested) .
The version will be 10.x.

Don't forget to rebase on develop

abdelr:
@msevestre Thanks for letting me know. I will start with the changes. One quick question. Did you come to an agreement about the FileLogger? Should we totally replace it by a third party implementation or should we fix the StreamWriter issue?

msevestre:
@abdelr @georgeDaskalakis 
As proposed last Friday, I'd like you to open the discussion tomorrow during our meeting to explain why we should move away from the current implementation. Adding 5+ new dependencies to replace a system that is under 50 lines of code including comments needs to be really thought through.  More specifically, you need to explain what we would be able to do that we cannot do now with the very naive and simple implementation of the FileLogger.

abdelr:
@msevestre Core should be in its final version now. Please take a look and merge it to develop if you agree with all changes. Also, we need to remove Serilog references and LoggingBuilderExtensions from PKSim since this has been moved to Core. The problem is that the nuget version is a bit strange yet with the naming. Last version is currently 10.0.5-cuhhvhhx and we would need it to be just 10.0.0.5. We need to refine the process now so we could reset the last digit on the version so we could easily update our references in PKSim.

msevestre:
Sorry I had reviewed but forgot to submit my comments. 
Just one required change because coding standard are not respected. Aside from that good to go

abdelr:
@msevestre Thanks! I have just committed the changes you requested.

msevestre:
Well done

msevestre:
Don't forget to reset the appveyor flags as discussed

PR #814, opened 2020-07-22
Importer branch

abdelr:


OSPSuite.Presentation.Importer/Core/DataFormat/ColumnsDataFormat.cs
```diff
@@ -0,0 +1,153 @@
...
+using System.Linq;
+using System.Reflection;
+using NPOI.SS.Util;
+using System;
+
+namespace OSPSuite.Presentation.Importer.Core.DataFormat
+{
+   public class ColumnsDataFormat : IDataFormat
+   {
+      public string Name => "ColumnsFormat";
```
msevestre, line 13: Define a property readonly get instead of a method that is evaluated everytime

OSPSuite.Presentation.Importer/Core/DataFormat/ColumnsDataFormat.cs
```diff
@@ -0,0 +1,153 @@
...
+
+      public string Description => "Simple format with data identified by column data...";
+
+      public IList<DataFormatParameter> Parameters { get; private set; }
+
+      public bool CheckFile(IUnformattedData data)
+      {
+         if (data.Headers.Where(h => h.Value.Level == ColumnDescription.MeasurementLevel.NUMERIC).Count() < 2)
```
msevestre, line 21: Explain what 2 means.. It's ok to use magic numbers but comments are required so that we understand what that means in 6 months

OSPSuite.Presentation.Importer/Core/DataFormat/ColumnsDataFormat.cs
```diff
@@ -0,0 +1,153 @@
...
+      }
+
+      private void SetParameters(IUnformattedData data)
+      {
+         var keys = data.Headers.Keys.ToList();
+         Parameters = new List<DataFormatParameter>();
+
+         var missingKeys = new List<string>();
+
+         extractQualifiedHeadings(data.Headers.Keys.ToList(), missingKeys);
```
msevestre, line 34: you already have` keys` above. use it

OSPSuite.Presentation.Importer/Core/DataSourceFileReaders/ExcelDataSourceFile.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System;
+using System.Collections.Generic;
+using OSPSuite.Core.Services;
+using OSPSuite.Infrastructure.Import.Services;
+using NPOI.SS.UserModel;
+using System.IO;
+using NPOI.HSSF.UserModel;
+using OSPSuite.Presentation.Importer.Infrastructure;
+
+namespace OSPSuite.Presentation.Importer.Core.DataSourceFileReaders
```
msevestre, line 10: Why are all those classes in the Presentation namespace? Aren't they the business logic and not related with presentation at all?

OSPSuite.Presentation.Importer/Core/IDataFormat.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace OSPSuite.Presentation.Importer.Core
+{
+
+   public interface IDataFormat
+   {
+      string Name { get; }
+      string Description { get; }
+      bool CheckFile(IUnformattedData rawData);
+      IList<DataFormatParameter> Parameters { get; }
+      IList<Dictionary<Column, IList<double>>> Parse(IUnformattedData data);
```
msevestre, line 14: Not found of this return value. It looks very much low level. 
Also exposing IList assumes that the consumer can modify the content. Why not exposing IReadOnlyList?

OSPSuite.Presentation.Importer/Core/IDataSet.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Collections.Generic;
+
+namespace OSPSuite.Presentation.Importer.Core
+{
+   /// <summary>
+   /// Data from a single experiment
+   /// </summary>
+   public interface IDataSet
```
msevestre, line 8: Maybe we do not need an interface for every single domain object we are creating?

OSPSuite.Presentation.Importer/Core/IDataSheet.cs
```diff
@@ -0,0 +1,39 @@
...
+      }
+
+      public ColumnNames Name { get; set; }
+      public string Unit { get; set; }
+   }
+
+   /// <summary>
+   /// e.g. a sheet in an excel file
+   /// </summary>
+   public interface IDataSheet
```
msevestre, line 23: Same here. Why do we have an interface? What's the purpose of the abstraction?

OSPSuite.Presentation.Importer/Core/IDataSourceFileParser.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace OSPSuite.Presentation.Importer.Core
+{
+   public interface IDataSourceFileParser
+   {
+      IDataSourceFile For(string path);
+   }
+
+   public class DataSourceFileParser : IDataSourceFileParser
+   {
+      private readonly ICsvDataSourceFile csvDataSourceFile;
```
msevestre, line 18: This does not follow our coding conventions. private members are _xxx

OSPSuite.Presentation.Importer/Infrastructure/ExcelReader.cs
```diff
@@ -0,0 +1,145 @@
...
+         }
+
+         System.Collections.IEnumerator excelRows = CurrentSheet.GetRowEnumerator();
+         excelRows.MoveNext();
+
+         var currentExcelRow = getCurrentExcelRow(excelRows);
+
+         int tableStart = 0;
+
+         for (int j = 0; j < currentExcelRow.LastCellNum; j++)
```
msevestre, line 119: LEt's start with i :)

OSPSuite.Presentation.Importer/Services/IImporter.cs
```diff
@@ -0,0 +1,51 @@
...
+
+   public class Importer : IImporter
+   {
+      private readonly IoC container;
+      private readonly IDataSourceFileParser parser;
+
+      public Importer(IoC container)
+      {
+         this.container = container;
+         this.parser = container.Resolve<IDataSourceFileParser>();
```
msevestre, line 23: Why not have the parser as dependencies?

OSPSuite.Presentation.Importer/Services/IImporter.cs
```diff
@@ -0,0 +1,51 @@
...
+      }
+
+      public IList<IDataFormat> AvailableFormats(IUnformattedData data)
+      {
+         return GetType().Assembly.GetTypes()
```
msevestre, line 28: containerhas a resolveAll or something like this that you can use
You can even use the IRepository<IDataFormat> in constructor which will inject a generic service doing exactly this

OSPSuite.Presentation.Importer/packages.config
```diff
@@ -0,0 +1,9 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="LumenWorksCsvReader" version="4.0.0" targetFramework="net472" />
+  <package id="NPOI" version="2.5.1" targetFramework="net472" />
+  <package id="OSPSuite.DevExpress" version="19.2.4.1" targetFramework="net472" />
+  <package id="OSPSuite.Utility" version="4.0.0.4" targetFramework="net472" />
+  <package id="Portable.BouncyCastle" version="1.8.6" targetFramework="net472" />
```
msevestre, line 7: What is this?

georgeDaskalakis: BouncyCastle is used by npoi. Without it npoi does not run.

OSPSuite.Presentation.Importer/Core/DataFormat/ColumnsDataFormat.cs
abdelr: At least a mapping for the time and the observation... Keep in mind that this is some kind of very simple format just to have the whole picture but until we agree on the formats this is not going to be thoroughly implemented

OSPSuite.Presentation.Importer/Services/IImporter.cs
abdelr: We already have the container as a dependency which enable us to resolve it... but yes, it will be more flexible adding it as a dependency

OSPSuite.Presentation.Importer/Infrastructure/ExcelReader.cs
georgeDaskalakis: corrected :)

OSPSuite.Presentation.Importer/Services/IImporter.cs
georgeDaskalakis: Done :)

OSPSuite.Presentation.Importer/OSPSuite.Presentation.Importer.csproj
```diff
@@ -0,0 +1,219 @@
...
+    <Compile Include="Core\IDataSheet.cs" />
+    <Compile Include="Core\IDataFormat.cs" />
+    <Compile Include="Core\IUnformattedData.cs" />
+    <Compile Include="Core\UnsupportedFormatException.cs" />
+    <Compile Include="Infrastructure\ExcelReader.cs" />
+    <Compile Include="PresentationImporterRegister.cs" />
+    <Compile Include="Presenters\ColumnMappingPresenter.cs" />
+    <Compile Include="Presenters\DataViewingPresenter.cs" />
+    <Compile Include="Presenters\IDataViewingPresenter.cs" />
+    <Compile Include="Presenters\IColumnMappingPresenter.cs" />
```
msevestre, line 161: The only reason why Presenter Views etc where all in one project was because it was going to be redesigned

Presneter=> in presenter project
Views=> in view project 
IView => in presenter project
etc. etc.

Ok if you rename OSPSuite.Presneteation.Omporter t- OSPsuite.Importer and have core and presentation logic together
But view should be in another project

OSPSuite.Presentation.Importer/Core/IDataSourceFile.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+
+namespace OSPSuite.Presentation.Importer.Core
+{
+   /// <summary>
+   /// Single file containing the data, e.g. excel file or csv file
+   /// </summary>
+   public interface IDataSourceFile
+   {
+      Dictionary<string, IDataSheet> DataSheets { get; set; }
```
msevestre, line 11: No get here. 
I think when definining contracts it makes sense to decide what is readonly and what not
Default as everything {get;set} makes it hard for the 'outsider' to guess how a entity is to be used

OSPSuite.Presentation.Importer/Core/IDataSourceFile.cs
```diff
@@ -0,0 +1,40 @@
...
+         get; 
+         set; 
+      }
+
+
+      private string _path;
+      public string Path 
+      {
+         get  => _path; 
+         set { _path = value; DataSheets = LoadFromFile(value); }
```
msevestre, line 35: Unexpected side effects of setting the path of a data source. I would call it explicitely

msevestre:
[ObservedData.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/4968189/ObservedData.xlsx)

OSPSuite.Presentation.Importer/Core/IUnformattedData.cs
```diff
@@ -0,0 +1,118 @@
...
+      bool AddRow(IEnumerable<string> row);
+      void AddColumn(string columnName, int columnIndex);
+      DataTable GetSheetAsDataTable();
+   }
+
+   public class UnformattedData : IUnformattedData
+   {
+      private readonly List<List<string>> _rawDataTable = new List<List<string>>();
+
+      public Dictionary<string, ColumnDescription> Headers { get; set; } =
```
msevestre, line 27: This should be readonly
Also we have a superclass called Cache that has a few more properties than the Dictionary. Maybe this could be used instead

OSPSuite.Presentation.Importer/Core/IUnformattedData.cs
```diff
@@ -0,0 +1,118 @@
...
+
+      public void AddColumn(string columnName, int columnIndex) //it seems to me there is little sense in adding column after column
+         //the list of headers is somehow the definition of the table
+      {
+         Headers.Add(columnName, new ColumnDescription(columnIndex));
+      }
+
+      public void CalculateColumnDescription(List<ColumnDescription.MeasurementLevel> levels)
+      {
+         Headers.ForEach(header =>
```
msevestre, line 38: This is using a ForEach method from DeExpress.
Another reason why UI component should not be referenced at all in the presentation project. 
Our own method is called `Each`

OSPSuite.Presentation.Importer/Core/IUnformattedData.cs
```diff
@@ -0,0 +1,118 @@
...
+      public Dictionary<string, ColumnDescription> Headers { get; set; } =
+         new Dictionary<string, ColumnDescription>(); //we have to ensure headers and RawData sizes match
+
+      public void AddColumn(string columnName, int columnIndex) //it seems to me there is little sense in adding column after column
+         //the list of headers is somehow the definition of the table
+      {
+         Headers.Add(columnName, new ColumnDescription(columnIndex));
+      }
+
+      public void CalculateColumnDescription(List<ColumnDescription.MeasurementLevel> levels)
```
msevestre, line 36: I don't understand the naming. What is the description of the column?

OSPSuite.Presentation.Importer/Core/IUnformattedData.cs
```diff
@@ -0,0 +1,118 @@
...
+         var resultList = new List<string>();
+
+         //change this
+         for (var i = 0; i < _rawDataTable.Count; i++)     //should have at least one row
+            resultList.Add(_rawDataTable[i][columnIndex]); //make sure the 2 indexes are correctly positioned here
+
+         return resultList;
+      }
+
+      public IEnumerable<string> GetColumn(string columnName)
```
msevestre, line 78: Why is this the same code as the GetCOlumn by Index?!?

OSPSuite.Presentation.Importer/Core/IUnformattedData.cs
```diff
@@ -0,0 +1,118 @@
...
+
+         var resultColumn = new List<string>();
+
+         for (var i = 0; i < _rawDataTable.Count; i++)
+            resultColumn.Add(_rawDataTable[i][columnIndex]);
+
+         return resultColumn;
+      }
+
+      public IEnumerable<string> GetRow(int index)
```
msevestre, line 90: I don't understnadthe resturned value .GetRow returns and IEnumerable<string> so does  GetColumn
Are we storing everything as string in the UnformatedTable?

should not we store the data per column and not per row? Pros and cons?
Or will we store the values by column later on?

OSPSuite.Presentation.Importer/Core/IUnformattedData.cs
```diff
@@ -0,0 +1,118 @@
...
+      public void CalculateColumnDescription(List<ColumnDescription.MeasurementLevel> levels)
+      {
+         Headers.ForEach(header =>
+         {
+            if (header.Value.Level == ColumnDescription.MeasurementLevel.NotSet) //hell, we could even not check here
+               header.Value.Level = levels[header.Value.Index];
+         });
+      }
+
+      public bool AddRow( IEnumerable<string> row)
```
msevestre, line 45: What does the boolean does as returned value.
In fact, what does that mean if it's false? Should not we throw an exception instead?

OSPSuite.Presentation.Importer/Core/DataSourceFileReaders/ExcelDataSourceFile.cs
```diff
@@ -0,0 +1,62 @@
...
+               }
+
+               loadedData.Add(sheetName, dataSheet);
+            }
+
+            return loadedData;
+         }
+         catch (Exception ex)
+         {
+            _logger.AddError(ex.ToString());
```
msevestre, line 57: AddException. Here you lose the stacktrace etc..

msevestre: but I am not quite sure that we should not rethrow...

OSPSuite.Presentation.Importer/Core/IUnformattedData.cs
```diff
@@ -0,0 +1,118 @@
...
+using System.Collections.Generic;
+using System.Data;
+using System.Linq;
+using DevExpress.DataProcessing;
+
+namespace OSPSuite.Presentation.Importer.Core
+{
+   public interface IUnformattedData
+   {
+      IEnumerable<string> GetRow(int index);
```
msevestre, line 11: Should we have more explicit names ? Such as ValuesAt or RowValuesAt etc, FilterRows,

OSPSuite.Presentation.Importer/Core/IUnformattedData.cs
```diff
@@ -0,0 +1,118 @@
...
+      IEnumerable<string> GetColumn(int columnIndex);
+      IEnumerable<string> GetColumn(string columnName);
+      Dictionary<string, ColumnDescription> Headers { get; }
+
+      IEnumerable<IEnumerable<string>> GetRows(Func<IEnumerable<string>, bool> filter);
+
+      bool AddRow(IEnumerable<string> row);
+      void AddColumn(string columnName, int columnIndex);
+      DataTable GetSheetAsDataTable();
```
msevestre, line 20: AsDataTable().

OSPSuite.Presentation.Importer/Core/IUnformattedData.cs
```diff
@@ -0,0 +1,118 @@
...
+      }
+
+      public DataTable GetSheetAsDataTable()
+      {
+         var resultTable = new DataTable();
+
+         // Add columns.
+         for (var i = 0; i < Headers.Count; i++)
+         {
+            resultTable.Columns.Add(Headers.ElementAt(i).Key, typeof(string));
```
msevestre, line 107: We have a method called AddColumn that allows you to simplify this quite a bit

OSPSuite.Presentation.Importer/Core/DataSourceFileReaders/CsvDataSourceFile.cs
```diff
@@ -0,0 +1,78 @@
...
+               }
+
+               var loadedData = new Dictionary<string, IDataSheet>();
+               loadedData.Add("", dataSheet);
+               return loadedData;
+            }
+         }
+         catch (Exception e)
+         {
+            _logger.AddError(e.ToString());
```
msevestre, line 48: same here. AddException

OSPSuite.Presentation.Importer/Core/DataFormat/ColumnsDataFormat.cs
```diff
@@ -0,0 +1,150 @@
...
+         var missingKeys = new List<string>();
+
+         extractQualifiedHeadings(keys, missingKeys);
+         extractNonQualifiedHeadings(keys, missingKeys, data);
+         extractGeneralParameters(keys, data);
+      }
+
+      private string extractUnits(string description)
+      {
+         var units = Regex.Match(description, @"\[.+\]").Value;
```
msevestre, line 38: Please implement directly support for multiple units def such as [] or ()

OSPSuite.Presentation.Importer/Core/DataFormat/ColumnsDataFormat.cs
```diff
@@ -0,0 +1,150 @@
...
+      private void extractQualifiedHeadings(List<string> keys, List<string> missingKeys)
+      {
+         foreach (var header in Enum.GetNames(typeof(Column.ColumnNames)))
+         {
+            var headerKey = keys.FirstOrDefault(h => h.ToUpper().Contains(header.ToUpper()));
+            if (headerKey != null)
+            {
+               keys.Remove(headerKey);
+               Column.ColumnNames columnName;
+               Enum.TryParse(header, out columnName);
```
msevestre, line 51: EnumHelper.ParseValue<>()

OSPSuite.Presentation.Importer/Core/DataFormat/ColumnsDataFormat.cs
```diff
@@ -0,0 +1,150 @@
...
+
+      private string extractUnits(string description)
+      {
+         var units = Regex.Match(description, @"\[.+\]").Value;
+         return units.Substring(1, units.Length - 2).Trim();
+      }
+
+      private void extractQualifiedHeadings(List<string> keys, List<string> missingKeys)
+      {
+         foreach (var header in Enum.GetNames(typeof(Column.ColumnNames)))
```
msevestre, line 44: EnumHelper.AllValuesFor<>()

OSPSuite.Presentation.Importer/Core/DataFormat/ColumnsDataFormat.cs
```diff
@@ -0,0 +1,150 @@
...
+         }
+      }
+
+      private void extractNonQualifiedHeadings(List<string> keys, List<string> missingKeys, IUnformattedData data)
+      {
+         foreach (var header in missingKeys)
+         {
+            var headerKey = keys.FirstOrDefault
+               (h => 
+                  data.Headers[h].Level == ColumnDescription.MeasurementLevel.Numeric &&
```
msevestre, line 67: This is super hard to read. Let's simplify this. Muliple Where and select ... I am too dumb for this

OSPSuite.Presentation.Importer/Core/DataFormat/ColumnsDataFormat.cs
```diff
@@ -0,0 +1,150 @@
...
+
+      private void buildDataSet(IUnformattedData data, IEnumerable<(string ColumnName, IList<string> ExistingValues)> parameters, Stack<int> indexes, List<Dictionary<Column, IList<double>>> dataSets)
+      {
+         if (indexes.Count() == parameters.Count())
+         {
+            var rawDataSet = data.GetRows(row =>
+            {
+               var check = true;
+               var i = 0;
+               while (check
```
msevestre, line 111: oh my. Can we simplify this also. What does this do?

OSPSuite.Presentation.Importer/Core/DataFormat/ColumnsDataFormat.cs
abdelr: Done

OSPSuite.Presentation.Importer/Core/DataSourceFileReaders/ExcelDataSourceFile.cs
abdelr: We started by putting everything together. Truth be told we did not understand the separation on presentation/UI that existed before so we were planning to discuss with you guys and then move the classes to their final projects

OSPSuite.Presentation.Importer/Core/IDataFormat.cs
abdelr: Maybe we could create a data-structure for this, but I personally do not see a problem in allowing edition from the outside. The point of this method is to create the different datasets, which are actually a copy of the imported unformatted data. What the caller do with the with the data is out of the scope of the format implementation.

OSPSuite.Presentation.Importer/Core/IDataSourceFileParser.cs
abdelr: Done

OSPSuite.Presentation.Importer/Infrastructure/ExcelReader.cs
abdelr: Done

OSPSuite.Presentation.Importer/Services/IImporter.cs
abdelr: Done

OSPSuite.Presentation.Importer/Core/DataFormat/DataFormatParameter.cs
```diff
@@ -0,0 +1,46 @@
...
+   public enum DataFormatParameterType
+   {
+      Mapping,
+      GroupBy,
+      MetaData
+   }
+
+   public abstract class DataFormatParameter
+   {
+      public string ColumnName { get; set; }
```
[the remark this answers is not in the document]
abdelr: I made the set protected so it cannot be set from the outside. The class is abstract and the constructors are below

OSPSuite.Presentation.Importer/Core/IDataSourceFile.cs
abdelr: I think the logic should remain together. Otherwise undesired behavior could show. Imagine you load a file and then change the path (but do not execute LoadFromFile explicitetly). Then you save everything for reloading next time. Next time you open it you will be opening a completely new file, maybe not even compatible with the configuration you have saved. It will not be able to reload the file definitely. Why would we want to change the path without reloading the file?

OSPSuite.Presentation.Importer/Core/IUnformattedData.cs
abdelr: Done

OSPSuite.Presentation.Importer/Core/DataFormat/ColumnsDataFormat.cs
abdelr: Again, this is a very basic class just to test the whole logic but I would not go into any more details until the real formats have been agreed upon

OSPSuite.Presentation.Importer/Core/DataFormat/ColumnsDataFormat.cs
abdelr: Done

OSPSuite.Presentation.Importer/Core/DataFormat/ColumnsDataFormat.cs
abdelr: Done

OSPSuite.Presentation.Importer/Core/DataFormat/ColumnsDataFormat.cs
abdelr: I hope it makes more sense now?

abdelr:
> It feels like this could be a generic type here no? especially the way it's being used later on. You are always casting to a know type (string, column) etc.

Maybe it is possible to do it like that, but I have not succeeded on it. I initially thought the same but I got complicated when having the concrete collection. A generic type is useful when you want to encapsulate a logic regardless the datatype but here we actually need to merge different datatypes. So imagine I would say I have ParameterConfiguration<T> and then MetaDataFormatParameter : DataFormatParameter<string> and MappingDataFormatParameter : DataFormatParameter<Column>, later I would need to have IEnumerable<ParameterConfiguration<?>> what datatype should I use here. I could not even say the collection is also generic since the collection would have different datatypes. I could maybe say it is IEnumerable<ParameterConfiguration<object>> but then I lose the generic add on since I would need to have casts all the time again. @msevestre, @georgeDaskalakis  Any suggestion here is most welcome. I have been struggling with this as well but although it looks like generics to me at the very beginning I end up moving into generalization/specialization at the end.

tests/OSPSuite.Presentation.Tests/Importer/Presenters/ColumnMappingPresenterSpecs.cs
```diff
@@ -55,14 +55,44 @@ public void identify_basic_format()
...
+         if (self == null)
+            return other == null;
+         if (self.ColumnName != other.ColumnName || self.Description != other.Description)
+            return false;
+         switch (self.Source)
+         {
+            case MappingDataFormatParameter mp:
+               if (!(other.Source is MappingDataFormatParameter))
+                  return false;
+               var mmp = other.Source as MappingDataFormatParameter;
```
msevestre, line 80: It feels that this method should be implemented in EACH Parameter class instead of delegating that outside?

abdelr: > It feels that this method should be implemented in EACH Parameter class instead of delegating that outside?

Yes, you are right, but it is only needed for testing purposes so I would not put it on the production code and since it is only intended for one test, should we spend much more time on it? Maybe just a function instead of an extension method? I am open to suggestions here... I will do it as you recommend...

msevestre: oh I did not see that it was in the test project. sorry

PR #819, opened 2020-08-19
Generic proposal again

msevestre:


msevestre:
@abdelr @georgeDaskalakis 
Same PR adapted to update code

There are quite a few issues that are found by resharper. I would strongly advise you to use resharper for consistency purposes.
For example, all Equals checks are flawed because the object could be null. Resharper creates the methods for you and you do not have to think about it
Also a bunch of code using lync could be simplified

msevestre:
@abdelr @georgeDaskalakis Note that the code does not compile because the tests are using an outdated version of setFormat (with two parameters instead of the required 3)

msevestre:
example of issue with equals pointed with resharper
image
if other is null, the code will throw an exception

Note that if you override equal, you NEED to override GetHashCode. Calling base.GetHashCode does not make sense I believe.

msevestre:
@abdelr @georgeDaskalakis Please commment here to make sure you have received this PR as I don't want to work on it a 3rd time

abdelr (approved):

georgeDaskalakis (approved):

abdelr:
Do not merge!!!

abdelr:
I was still not completely comfortable with this solution since the generics is a bit strange to me. I think we are using generics only because an error in the code design. So we are trying to solve with generics what should have been solved with generalization/specialization. There is no need for the field Configuration on DataFormatParameter (that is why we even have null unnecessary objects in Ignored or GroupBy cases). I removed it and only kept the typed properties on the derived classes (MetaDataId in MetaDataFormatParameter and MappedColumn in MappingDataFormatParameter). You could see the changes in commit "52c1a79". I also removed the Equals overwritten methods since it was only needed for testing purposes and we should not modify production code only for testing purposes. Please take a look on it and feedback with your thoughts. It reads cleaner to me now. I will really appreciate any criticism and we could come back to this solution if you think it is the best.

msevestre:
sounds good. The generic was trying to solve a problem that is not there anymore.

PR #821, opened 2020-08-24
Possible fix for refresh issue plus suggestion for more optimistic code

msevestre:


msevestre:
@abdelr The reason why it does not refresh automatically is because you are changing the content of the bound list. e.g. the binder cannot find the value in the list as it does not exist anymore. The only way at the moment is to use Rebind. (e.g. the list of stuff bound to is changing dynamically invalidating the selected value)

OSPSuite.Presentation.Importer/Presenters/ColumnMappingPresenter.cs
```diff
@@ -38,11 +39,11 @@ IApplicationController applicationController
          {
             var format = _availableFormats.First(f => f.Name == formatName);
             SetDataFormat(format, _availableFormats);
-            OnFormatChanged?.Invoke(format);
+            OnFormatChanged(format);
          });
       }
 
-      public event FormatChangedHandler OnFormatChanged;
+      public event FormatChangedHandler OnFormatChanged = delegate { };
```
msevestre, line 44: Always initialize an event handler to empty delegate. that way you do not have to always check for null
Also I'd suggest to use the generic construct EventHandler<T> instead of defining your own EventHandler type. 
So instead of FormatChangedHandler, specify EventHandler<Format> or whatever eventargs you have

OSPSuite.Presentation.Importer/Presenters/ColumnMappingPresenter.cs
```diff
@@ -101,10 +102,8 @@ public void ChangeUnitsOnRow(ColumnMappingViewModel model)
             {
                column.Unit = unitsEditorPresenter.SelectedUnit;
                model.Description = ColumnMappingFormatter.Stringify(model.Source);
-               OnDataFormatParametersChanged?.Invoke(this, new DataFormatParametersChangedArgs()
-               {
-                  Parameters = new List<DataFormatParameter>() { activeRow.Source }
```
msevestre, line 106: IMO event parameters should only be readonly. This also allows to simplify the code by using a constructor

OSPSuite.Presentation.Importer/Presenters/ColumnMappingPresenter.cs
```diff
@@ -101,10 +102,8 @@ public void ChangeUnitsOnRow(ColumnMappingViewModel model)
...
-               });
+               _view.RefreshList();
```
msevestre: THis is the one method that is needed. This is defacto very similar to what you were doing before with the begin update and endupdate. We need to rebing to the lists.

OSPSuite.Presentation.Importer/Views/ColumnMappingControl.cs
```diff
@@ -83,13 +85,13 @@ public override void InitializeBinding()
             .WithShowButton(ShowButtonModeEnum.ShowAlways);
 
          _gridViewBinder.AddUnboundColumn()
-            .WithCaption("...")
+            .WithCaption(UIConstants.EMPTY_COLUMN)
```
msevestre, line 88: Using are default for empty columns for consistency purposes

OSPSuite.Presentation.Importer/Views/ColumnMappingControl.cs
```diff
@@ -136,6 +139,11 @@ private void onFormatChanged(object sender, EventArgs e)
 
       public event FormatChangedHandler OnFormatChanged;
 
+      public void Rebind()
+      {
+         _gridViewBinder.Rebind();
```
msevestre, line 144: Only way at the moment to let the binder knows that the underlying list of possibility for the combo box has changed

abdelr:
Nice, I would then remove the public methods BeginUpdate and EndUpdate from the view interface. Many thanks for the explaination.

PR #823, opened 2020-09-04
Exclude LoggerCreator from scanner for later registration as singleton

abdelr:


msevestre:
The build was failing because of outdated SmartXLS references . I have just updated those

PR #833, opened 2020-10-22
Fixes #832 issue with molecule replacement

### LINKED ISSUE #832: Molecule keyword replacement does not work for global parmaeter referencing other global molecule parameters
Trying to define a global parameter such as 

fract_exp_baso = 1 - fract_exp_api - fract_exp_end does not work when all references are global

msevestre:


src/OSPSuite.Core/Domain/Services/KeywordReplacerTask.cs
```diff
@@ -271,11 +279,13 @@ private void replaceInUsingFormula(IUsingFormula usingFormula, IContainer rootCo
...
       {
-         var moleculeAmount = parameter.ParentContainer as IMoleculeAmount;
-         if (moleculeAmount == null)
-            replaceIn(parameter, rootContainer, string.Empty);
-         else
-            ReplaceIn(parameter, rootContainer, moleculeAmount.Name);
+         var parameterContainer = parameter.ParentContainer;
+
+         //Global molecule container or local molecule amount container
+         var isInMolecule = parameterContainer.IsAnImplementationOf<IMoleculeAmount>() ||
```
msevestre, line 285: @Yuri05 This is the issue here being fixed. The rest is noise.
In short, global molecule parameters referencing other global molecule parameters using keyword would not work. Replacement was not being done for MOLECULE in that case

Yuri05 (approved):

PR #834, opened 2020-10-23
831 reverse command

msevestre:


src/OSPSuite.Core/Commands/Core/Command.cs
```diff
@@ -93,7 +93,7 @@ public interface IReversibleCommand<in TExecutionContext> : ICommand<TExecutionC
       /// <summary>
       ///   Create an inverse for the current command
       /// </summary>
-      IReversibleCommand<TExecutionContext> InverseCommand(TExecutionContext context);
+      ICommand<TExecutionContext> InverseCommand(TExecutionContext context);
```
msevestre, line 96: @Yuri05 I am changing this to not require the undo command to also be undoable. That way we can have the best of both worlds

src/OSPSuite.Core/Commands/Core/CommandExtensions.cs
```diff
@@ -25,7 +25,7 @@ public static bool IsEmpty(this ICommand command)
       public static bool IsEmptyMacro(this ICommand command)
       {
          var macroCommand = command as IMacroCommand;
-         return macroCommand != null && macroCommand.IsEmtpy;
```
msevestre, line 28: We've had this typo since the beginning!!

PR #835, opened 2020-10-27
WIP select on click: Some improvements

msevestre:


msevestre:
@abdelr and @georgeDaskalakis Let's review tomorrow. There is some code in the controller that should be changed I believe.
I am getting some error at the moment and I do not know why. I can explain the concept tomorrow

src/OSPSuite.UI/Views/Importer/ColumnMappingControl.cs
```diff
@@ -52,13 +56,14 @@ public ColumnMappingControl(IImageListRetriever imageListRetriever)
...
+         columnMappingGridView.MultiSelect = false;
+         columnMappingGridView.CellValueChanged += (o, e) => OnEvent(_presenter.ValidateMapping);
 
          columnMappingGridView.OptionsView.ShowButtonMode = ShowButtonModeEnum.ShowOnlyInEditor;
-         columnMappingGridView.OptionsBehavior.EditorShowMode = EditorShowMode.MouseUp;
-         columnMappingGridView.MouseDown += onMouseDown;
-         columnMappingGrid.ToolTipController = new ToolTipController();
-         columnMappingGrid.ToolTipController.GetActiveObjectInfo += (o, e) => this.DoWithinExceptionHandler(() => onGetActiveObjectInfo(o, e));
+         columnMappingGridView.OptionsBehavior.EditorShowMode = EditorShowMode.MouseDown;
+         columnMappingGridView.MouseDown += (o, e) => OnEvent(onMouseDown, o, e);
```
msevestre, line 62: EVery single event in the UI should be in OnEvent. Otherwise, crash would lead to app disappearing completely. Major

src/OSPSuite.UI/Views/Importer/ColumnMappingControl.cs
```diff
@@ -68,33 +73,33 @@ public void AttachPresenter(IColumnMappingPresenter presenter)
...
+         var descriptionRepository = new UxRepositoryItemComboBox(columnMappingGridView)
          {
             AutoComplete = true,
             AllowNullInput = DefaultBoolean.True,
             CloseUpKey = new KeyShortcut(Keys.Enter)
          };
-         fillComboBoxItems(repo, _presenter.GetAvailableOptionsFor(model));
-         return repo;
+//         descriptionRepository.FillImageComboBoxRepositoryWith(_presenter.GetAvailableOptionsFor(model), x => x.IconIndex);
+         fillComboBoxItems(descriptionRepository, _presenter.GetAvailableOptionsFor(model));
```
msevestre, line 83: we have some predefined function fro all of this. However I am not quite sure what the extra stuff is doing. If we want the UI to behave like the rest of the suite, let see if this is needed or not

src/OSPSuite.UI/Views/Importer/ColumnMappingControl.cs
```diff
@@ -104,7 +109,7 @@ public override void InitializeBinding()
          base.InitializeBinding();
          _gridViewBinder.AddUnboundColumn()
             .WithCaption(UIConstants.EMPTY_COLUMN)
-            .WithShowButton(ShowButtonModeEnum.ShowOnlyInEditor)
+            .WithShowButton(ShowButtonModeEnum.ShowAlways)
```
msevestre, line 110: Button should always be visible

src/OSPSuite.UI/Views/Importer/ColumnMappingControl.cs
```diff
@@ -194,12 +205,14 @@ private RepositoryItem lloqRepository(ColumnMappingDTO model)
...
+
          return _disabledLloqButtonRepository;
       }
 
       public void Rebind()
-      { 
-         _gridViewBinder.Rebind();
+      {
+         columnMappingGridView.RefreshData();
+         // _gridViewBinder.Rebind();
```
msevestre, line 213: We need to discuss here. Rebind is called quite often. It's required when adding or deleting rows. But when modifying bound date, it is not

src/OSPSuite.UI/Views/Importer/ColumnMappingControl.Designer.cs
```diff
@@ -19,6 +19,7 @@ protected override void Dispose(bool disposing)
          {
             components.Dispose();
          }
+         _gridViewBinder.Dispose();
```
msevestre, line 22: @abdelr @georgeDaskalakis It is key to dispose the binders. Otherwise you will get memory leaks

src/OSPSuite.UI/Views/Importer/ColumnMappingControl.cs
msevestre: in fact this is only required for combo box. Otherwise buttons are visible by default

src/OSPSuite.UI/Views/Importer/ColumnMappingControl.cs
```diff
@@ -115,31 +118,32 @@ public override void InitializeBinding()
...
             .WithCaption(UIConstants.EMPTY_COLUMN)
-            .WithShowButton(ShowButtonModeEnum.ShowOnlyInEditor)
+            .WithShowButton(ShowButtonModeEnum.ShowAlways)
             .WithRepository(lloqRepository)
             .WithFixedWidth(UIConstants.Size.BUTTON_WIDTH);
 
          _removeButtonRepository.ButtonClick += (o, e) =>
          {
             columnMappingGridView.ActiveEditor.EditValue = ColumnMappingFormatter.Ignored();
+            columnMappingGridView.CloseEditor();
```
msevestre, line 146: This ensures that the editor is released. I am not sure why this is required in the first place, probably because of how the objects are updated?  not sure

src/OSPSuite.UI/Views/Importer/ColumnMappingControl.cs
georgeDaskalakis: done overall for the events in the Importer/Views

PR #842, opened 2020-11-03
Fixes #841 improve entity path

### LINKED ISSUE #841: Improve EntityPath not to return path beyond root container

msevestre:


src/OSPSuite.Core/Domain/EntityExtensions.cs
```diff
@@ -31,7 +33,7 @@ private static List<string> entityPathAsListFor(IEntity entity)
          while (parent != null)
          {
             list.Insert(0, parent.Name);
-            parent = parent.ParentContainer;
+            parent = parent.IsAnImplementationOf<IRootContainer>() ? null : parent.ParentContainer;
```
msevestre, line 36: @Yuri05 This was a bug. The path should not go beyond the root container

PR #845, opened 2020-11-05
843 remove nan parameters

msevestre:


src/OSPSuite.Core/Domain/Container.cs
```diff
@@ -22,7 +22,7 @@ public interface IContainer : IEntity, IEnumerable<IEntity>
       /// <summary>
       ///    Returns the children of the Container
       /// </summary>
-      IEnumerable<IEntity> Children { get; }
+      IReadOnlyList<IEntity> Children { get; }
```
msevestre, line 25: THis will make life easier. Not sure why it was still IEnumerable

src/OSPSuite.Core/Domain/Services/ModelConstructor.cs
```diff
@@ -112,6 +113,20 @@ private void finalizeModel(IModel model)
 
          //now we should be able to resolve all references
          _referencesResolver.ResolveReferencesIn(model);
+
+
+         //This should be done after reference were resolved to ensure that we do not remove formula parameter that could not be evaluated
+         removeUndefinedInstanceParametersIn(model);
```
msevestre, line 119: Once everything is created, I am removing explicitly parameter instances with NaN values

src/OSPSuite.Core/Domain/Services/ModelConstructor.cs
```diff
@@ -112,6 +113,20 @@ private void finalizeModel(IModel model)
...
+      }
+
+      private void removeUndefinedInstanceParametersIn(IModel model)
+      {
+         var allNaNParametersFromMolecules = model.Root.GetAllChildren<IContainer>()
+            .Where(c => c.ContainerType == ContainerType.Molecule)
```
msevestre, line 125: @Yuri05  Right now only for Molecule. But..should we do this for all parameters in a simulation

@PavelBal Can you think of a use case where it makes sense to have parameter in a simulation with a NaN value?

tests/OSPSuite.Core.IntegrationTests/ModelConstructorIntegrationTests.cs
```diff
@@ -406,6 +406,16 @@ public void should_have_update_the_scale_factor()
          bone_cell_E.ScaleDivisor.ShouldBeEqualTo(2.5);
       }
 
+
+      [Observation] 
+      public void should_have_removed_local_molecule_parameters_with_a_NAN_value()
+      {
+         var bone_cell_E = _model.ModelOrganCompartmentMolecule(ConstantsForSpecs.Bone, ConstantsForSpecs.Cell, "E");
+         bone_cell_E.Parameter("NaNParam").ShouldBeNull();
+         bone_cell_E.Parameter("OtherNaNParam").ShouldNotBeNull();
```
msevestre, line 415: testing this too

src/OSPSuite.Core/Domain/Services/ModelConstructor.cs
PavelBal: @msevestre No, actually not. Get rid of them :)

Yuri05: >But..should we do this for all parameters in a simulation?

No. Even though I do not have any use case in mind where we would need NaN-Parameter: they do not harm. So please keep them.

Yuri05 (approved):

PR #847, opened 2020-11-13
Fixes #846 as well as Fixes #843

### LINKED ISSUE #843: Do not generate parameter instance in simulation when resulting value is NaN
This is a new use case requried with the update to rel exp. We have local parameters defined at the molecule level such as fraction instertial or fraction intracellular but the instances int he simulation should only be where the resulting value is not NAN

### LINKED ISSUE #846: EntityValidator does not validate parameter in molecules that are not compound
Now that we have formula parmaeters etc.. I think we should also do it for molecules and potentially remove parameters that are NaN for instance from undefined molecules etc...

msevestre:


src/OSPSuite.Core/Domain/Services/EntityValidator.cs
```diff
@@ -78,24 +76,13 @@ public void Visit(IEntity entity)
 
       public void Visit(IParameter parameter)
       {
-         //only validate visible parmaeter
+         //only validate visible parameter
          if (!parameter.Visible)
             return;
 
-         //do not validate parameter defined in molecule amount for other molecule than drug (value cannot be set by user and some values are invalid)
-         if (parentIsAMoleculeButNotDrug(parameter.ParentContainer))
```
msevestre, line 86: @Yuri05 Becasue we are removing local nan parameters now, I don't think this is required anymore. This will also ensure that we validate all parameters of a enzyme that may be invalid

src/OSPSuite.Core/Domain/Services/ModelConstructor.cs
```diff
@@ -116,14 +116,15 @@ private void finalizeModel(IModel model)
...
+         removeUndefinedLocalMoleculeParametersIn(model);
       }
 
-      private void removeUndefinedInstanceParametersIn(IModel model)
+      private void removeUndefinedLocalMoleculeParametersIn(IModel model)
       {
          var allNaNParametersFromMolecules = model.Root.GetAllChildren<IContainer>()
             .Where(c => c.ContainerType == ContainerType.Molecule)
             .SelectMany(c => c.AllParameters(p => double.IsNaN(p.Value)))
+            .Where(x => x.BuildMode == ParameterBuildMode.Local)
```
msevestre, line 127: WE CANNOT REMOVE global parameters (see bug with weird SimModel error).

src/OSPSuite.Infrastructure/Services/LoggerCreator.cs
```diff
@@ -26,15 +27,11 @@ public ILogger GetOrCreateLogger(string categoryName)
 
       private ILogger setupLogger(string categoryName)
       {
-         using (var loggerFactory = LoggerFactory.Create(
-             builder =>
```
msevestre, line 30: Just for me to be able to read this

PR #852, opened 2020-11-30
Fixes #851 reset id

### LINKED ISSUE #851: ResetId Upon loading simulation
relates to https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/446

msevestre:


msevestre:
@Yuri05 This is a request that makes sense. I am adding it to the v10 branch so that it does not break anything but this should be fairly safe

PR #866, opened 2021-01-11
Importer GUI Redesign

georgeDaskalakis:


src/OSPSuite.Assets.Images/ApplicationIcons.cs
```diff
@@ -137,7 +137,7 @@ public static class ApplicationIcons
       public static readonly ApplicationIcon Human = AddNamedIcon("Human", "Human");
       public static readonly ApplicationIcon HumanGreen = AddNamedIcon("HumanGreen", "HumanGreen");
       public static readonly ApplicationIcon HumanRed = AddNamedIcon("HumanRed", "HumanRed");
-      public static readonly ApplicationIcon Import = AddNamedIcon("ObservedData", "Import");
+      public static readonly ApplicationIcon Import = AddNamedIcon("ObservedData", "LoadCurrentSheet");
```
msevestre, line 140: This needs to be reverted

src/OSPSuite.Assets.Images/ApplicationIcons.cs
```diff
@@ -441,7 +441,7 @@ public static class ApplicationIcons
       public static readonly ApplicationIcon Tree = AddNamedIcon("Tree", "Tree");
       public static readonly ApplicationIcon Diagram = AddNamedIcon("Diagram", "Diagram");
       public static readonly ApplicationIcon ContainerObserver = AddNamedIcon("ContainerObserver", "ContainerObserver");
-      public static readonly ApplicationIcon ImportAll = AddNamedIcon("ImportAll", "ImportAll");
```
msevestre, line 444: This needs to be reverted

src/OSPSuite.Presentation/Presenters/Importer/ImporterDataPresenter.cs
```diff
@@ -0,0 +1,163 @@
...
+   {
+      private readonly IImporter _importer;
+      private IDataSourceFile _dataSourceFile;
+      private IReadOnlyList<ColumnInfo> _columnInfos;
+      private IReadOnlyList<MetaDataCategory> _metaDataCategories;
+      private readonly IDataSetToDataRepositoryMapper _dataRepositoryMapper;
+      public Cache<string, IDataSheet> Sheets { get; set; }
+
+      public event EventHandler<FormatChangedEventArgs> OnFormatChanged = delegate { };
+      public event EventHandler<TabChangedEventArgs> OnTabChanged;
```
msevestre, line 23: Initialize event to empty delegate

src/OSPSuite.UI/Views/Importer/ModalImporterView.cs
```diff
@@ -0,0 +1,175 @@
...
+
+      public void AttachImporterPresenter(IImporterPresenter presenter)
+      {
+         presenter.OnTriggerImport += (s, d) =>
+         {
+            DialogResult = System.Windows.Forms.DialogResult.OK;
+         };
+      }
+
+      private void InitializeComponent()
```
msevestre, line 46: Why is this code in the main file and not in a designer file?

src/OSPSuite.UI/Views/Importer/ImporterDataView.resx~merged
```diff
@@ -0,0 +1,129 @@
...
+    Version 2.0
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
+    associated with the data types.
+    
+    Example:
+    
+    ... ado.net/XML headers & schema ...
```
msevestre, line 15: Removve this file?

src/OSPSuite.UI/Views/Importer/ImporterDataView.cs
```diff
@@ -0,0 +1,223 @@
...
+using OSPSuite.UI.Extensions;
+
+namespace OSPSuite.UI.Views.Importer
+{
+   public partial class ImporterDataView : BaseUserControl, IImporterDataView
+   {
+      private IImporterDataPresenter _dataPresenter;
+
+      private string _contextMenuSelectedTab;
+      private bool sheetImportedFlag;
```
msevestre, line 21: private member should be _xxx

msevestre (changes requested):
Looking good

src/OSPSuite.Infrastructure.Import/Core/DimensionHelper.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System.Collections.Generic;
 
-namespace OSPSuite.Core.Importer
+namespace OSPSuite.Infrastructure.Import.Core
 {
    public static class DimensionHelper
```
msevestre, line 5: I think we need to review usage of all those DImension specific importer stuff. It could be that we need a bunch of new methods (e.g. a DimensionHelper or whatever) but I don't think we need all those extra stuff

src/OSPSuite.Infrastructure.Import/Core/ExtendedColumn.cs
```diff
@@ -0,0 +1,21 @@
+﻿using OSPSuite.Core.Import;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace OSPSuite.Infrastructure.Import.Core
+{
+   public class ExtendedColumn
+   { 
+      public Column Column { get; set; } //WE END UP HAVING COLUMN WAY TOO MANY TIMES LIKE THIS
```
msevestre, line 10: What does the comment mean?

src/OSPSuite.Infrastructure.Import/Core/IColumnDescription.cs
```diff
@@ -0,0 +1,33 @@
...
+      //not sure we are going to be needing this kind of constructor
+      public ColumnDescription(int index, List<string> existingValues, MeasurementLevel columnDataType)
+      {
+         Index = index;
+         ExistingValues = existingValues;
+         Level = columnDataType;
+      }
+
+      public int Index { get; private set; }
+      public IList<string> ExistingValues { get; set; }
```
msevestre, line 30: HEre again encapsulation of values. Let's make sure we do not expose (in general, of course there are exceptions) remember as IList. They should be IReadonlyList<T> without get

src/OSPSuite.Infrastructure.Import/Core/IDataSet.cs
```diff
@@ -0,0 +1,62 @@
...
+      public int Id { get; set; }
+      public string Value { get; set; }
+   }
+
+   /// <summary>
+   /// Data from a single experiment
+   /// </summary>
+   public interface IDataSet
+   {
+      IEnumerable<ParsedDataSet> Data { get; set; }
```
msevestre, line 20: IREadonly and no set

src/OSPSuite.Infrastructure.Import/Core/IDataSet.cs
```diff
@@ -0,0 +1,62 @@
...
+            throw new NanException();
+      }
+
+      public void ClearNan(double indicator)
+      {
+         for (var dataSetIndex = 0; dataSetIndex < Data.Count(); dataSetIndex++)
+         {
+            var dataSet = Data.ElementAt(dataSetIndex);
+            var mainColumns = dataSet.Data.Where(pair => pair.Key.ColumnInfo.IsMandatory).Select(pair => pair.Value).ToList();
+            for (var i = 0; i < mainColumns.Count(); i++)
```
msevestre, line 41: Why don't you foreach?

src/OSPSuite.Infrastructure.Import/Core/IDataSet.cs
```diff
@@ -0,0 +1,62 @@
...
+
+      public void ThrowsOnNan(double indicator)
+      {
+         if (Data.Any(dataSet => dataSet.Data.Any(pair => pair.Key.ColumnInfo.IsMandatory && pair.Value.Any(point => point.Value == indicator || double.IsNaN(point.Value)))))
+            throw new NanException();
+      }
+
+      public void ClearNan(double indicator)
+      {
+         for (var dataSetIndex = 0; dataSetIndex < Data.Count(); dataSetIndex++)
```
msevestre, line 37: Why not foreach?

src/OSPSuite.Infrastructure.Import/Core/IDataSet.cs
```diff
@@ -0,0 +1,62 @@
...
+            }
+         }
+      }
+
+      private void removeRowsContainingElements(IList<SimulationPoint> elements, IList<SimulationPoint> column, ParsedDataSet dataSet)
+      {
+         foreach (var element in elements)
+         {
+            var index = column.IndexOf(element);
+            for (var valueIndex = 0; valueIndex < dataSet.Data.Values.Count(); valueIndex++)
```
msevestre, line 54: foreach

src/OSPSuite.Infrastructure.Import/Core/IDataSheet.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace OSPSuite.Infrastructure.Import.Core
+{
+   /// <summary>
+   ///    e.g. a sheet in an excel file
+   /// </summary>
+   public interface IDataSheet
```
msevestre, line 10: do you need that interface?

src/OSPSuite.Infrastructure.Import/Core/IDataSource.cs
```diff
@@ -0,0 +1,119 @@
...
+   {
+      private IImporter _importer;
+      private ImporterConfiguration _configuration;
+
+      public DataSource(IImporter importer)
+      {
+         _importer = importer;
+         _configuration = new ImporterConfiguration();
+
+         var type = typeof(DynamicQueryable).Assembly.GetType("System.Linq.Dynamic.ExpressionParser");
```
msevestre, line 39: What is this?

src/OSPSuite.Infrastructure.Import/Core/IDataSource.cs
```diff
@@ -0,0 +1,119 @@
...
+         return _configuration;
+      }
+
+      public IEnumerable<MetaDataMappingConverter> GetMappings()
+      {
+         return _mappings;
+      }
+
+      private IEnumerable<MetaDataMappingConverter> _mappings;
+      public Cache<string, IDataSet> DataSets { get; } = new Cache<string, IDataSet>();
```
msevestre, line 112: move all those members up. Feels weird to have member definition at the end of the file

src/OSPSuite.Infrastructure.Import/Core/IDataSourceFileParser.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+   public interface IDataSourceFileParser
+   {
+      IDataSourceFile For(string path);
+   }
+
+   public class DataSourceFileParser : IDataSourceFileParser
+   {
+      readonly string[] _csvExtensions = { ".csv" };
+      readonly string[] _excelExtensions = { ".xls", ".xlsx" };
```
msevestre, line 14: private
and use existing constants

msevestre: name of file should be name of class not of interface

src/OSPSuite.Infrastructure.Import/Core/IDataSourceFileParser.cs
```diff
@@ -0,0 +1,41 @@
...
+      }
+
+      public IDataSourceFile For(string path)
+      {
+         if (_csvExtensions.Any(path.EndsWith))
+         {
+            _csvDataSourceFile.Path = path;
+            return _csvDataSourceFile;
+         }
+         else if (_excelExtensions.Any(path.EndsWith))
```
msevestre, line 32: else not needed

src/OSPSuite.Infrastructure.Import/Core/ImporterExceptions.cs
```diff
@@ -0,0 +1,18 @@
+﻿using OSPSuite.Utility.Exceptions;
+
+namespace OSPSuite.Infrastructure.Import.Core
+{
+   public class EmptyNamingConventionsException : OSPSuiteException
+   {
+      public EmptyNamingConventionsException() : base("Column naming conventions cannot be empty.")
```
msevestre, line 7: move all those strings to captions=>errors

src/OSPSuite.Infrastructure.Import/Core/ImporterExceptions.cs
```diff
@@ -0,0 +1,18 @@
...
+      {
+      }
+   }
+
+   public class NullNamingConventionsException : OSPSuiteException
+   {
+      public NullNamingConventionsException() : base("Column naming conventions cannot be null.")
```
msevestre, line 14: move all those strings to captions=>errors

src/OSPSuite.Infrastructure.Import/Core/InvalidFileException.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+
+namespace OSPSuite.Infrastructure.Import.Core
+{
+   public class InvalidFileException : Exception
+   {
+      private const string MESSAGE = "An error occurred while reading the file. Please check the content";
```
msevestre, line 7: move all those strings to captions=>errors

src/OSPSuite.Infrastructure.Import/Core/Mappers/IDataSetToDataRepositoryMapper.cs
```diff
@@ -0,0 +1,203 @@
...
+
+         if (column.Key.ColumnInfo.IsBase())
+            dataColumn = new BaseGrid(column.Key.ColumnInfo.Name, dimension);
+         else
+         {
+            
+            if (!containsColumnByName(dataRepository.Columns, column.Key.ColumnInfo.BaseGridName))
+            {
+               //convertParsedDataColumn(dataRepository, importDataTable.Columns.ItemByName(colInfo.BaseGridName));
+               throw new Exception();
```
msevestre, line 75: please, no Exception without any context

src/OSPSuite.Infrastructure.Import/Core/Mappers/IDataSetToDataRepositoryMapper.cs
```diff
@@ -0,0 +1,203 @@
...
+      private static bool containsColumnByName(IEnumerable<DataColumn> columns, string name)
+      {
+         return columns.Any(col => col.Name == name);
+      }
+
+      private ParsedDataSet getDataSet(IDataSource dataSource, int dataSetIndex, out int sheetIndex)
+      {
+         sheetIndex = 0;
+         var sheet = dataSource.DataSets.GetEnumerator();
+         while (sheet.MoveNext() && dataSetIndex >= 0)
```
msevestre, line 189: What is going on here? 
Why is there a return value AND and out parameter?

src/OSPSuite.Infrastructure.Import/Core/NanException.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+
+namespace OSPSuite.Infrastructure.Import.Core
+{
+   public class NanException : Exception
+   {
+      private const string MESSAGE = "Data contains NaN values at imported columns. Select a different action for NaN values or clean your data.";
```
msevestre, line 7: move message to constants

src/OSPSuite.Infrastructure.Import/Core/UnformattedDataRow.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using OSPSuite.Utility.Collections;
+
+namespace OSPSuite.Infrastructure.Import.Core
+{
+   public class UnformattedDataRow :IEnumerable
```
msevestre, line 8: Usage of IEnumerable without generic feels so .NET 1. Don't you know the type?

src/OSPSuite.Infrastructure.Import/Core/MetaDataRowCollection.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System.Data;
 
-namespace OSPSuite.Core.Importer
+namespace OSPSuite.Infrastructure.Import.Core
 {
    public class MetaDataRowCollection : System.Collections.IEnumerable
```
msevestre, line 5: Usage of IEnumerable without generic feels so .NET 1. Don't you know the type?

src/OSPSuite.Infrastructure.Import/Core/MetaDataTable.cs
```diff
@@ -2,9 +2,8 @@
 using System.Collections.Generic;
 using System.Data;
 using System.Linq;
-using OSPSuite.Assets;
 
-namespace OSPSuite.Core.Importer
+namespace OSPSuite.Infrastructure.Import.Core
 {
    public class MetaDataTable : DataTable, System.Collections.IEnumerable
```
msevestre, line 8: Usage of IEnumerable without generic feels so .NET 1. Don't you know the type?

src/OSPSuite.Infrastructure.Import/Core/UnsupportedFormatException.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+
+namespace OSPSuite.Infrastructure.Import.Core
+{
+   public class UnsupportedFormatException : Exception
+   {
+      private const string MESSAGE = "The file format is not supported";
```
msevestre, line 7: constant

src/OSPSuite.Infrastructure.Import/Services/IDataImporter.cs
```diff
@@ -0,0 +1,31 @@
...
+
+   public interface IDataImporter
+   {
+      /// <summary>
+      ///    This function retrieves a list of imported DataRepositories.
+      /// </summary>
+      /// <param name="metaDataCategories">Specification of meta data of the table.</param>
+      /// <param name="columnInfos">Specification of columns including specification of meta data.</param>
+      /// <param name="dataImporterSettings">Settings used to initialize the view</param>
+      (IEnumerable<DataRepository> DataRepositories, ImporterConfiguration Configuration) ImportDataSets(
```
msevestre, line 17: Can we return IReadonly?

src/OSPSuite.Infrastructure.Import/Services/IDataImporter.cs
```diff
@@ -0,0 +1,31 @@
...
+         IReadOnlyList<MetaDataCategory> metaDataCategories, 
+         IReadOnlyList<ColumnInfo> columnInfos, 
+         DataImporterSettings dataImporterSettings
+      );
+
+      IEnumerable<DataRepository> ImportFromConfiguration(
```
msevestre, line 23: Can we return IReadonly?

src/OSPSuite.Infrastructure.Import/Services/IImporter.cs
```diff
@@ -0,0 +1,159 @@
...
+            .Where(p => p.Item2 > 0)
+            .OrderByDescending(p => p.Item2)
+            .Select(p => p.x);
+      }
+
+      public void AddFromFile(IDataFormat format, Cache<string, IDataSheet> dataSheets, IReadOnlyList<ColumnInfo> columnInfos, IDataSource alreadyExisting)
+      {
+         var dataSets = new Cache<string, IDataSet>();
+
+         foreach (var s in dataSheets.KeyValues)
```
msevestre, line 56: what is s? name your parameter

src/OSPSuite.Presentation/Presenters/Importer/ImporterDataPresenter.cs
msevestre: this is still missing

msevestre (changes requested):
Lots to change. Please make sure we address the concerns here so that we can MERGE this monster PR and start incrementally adding features.

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1477,6 +1562,16 @@ public static string WildCardRecursiveCannotBePartOfPath(string wildCardRec, str
       }
 
       public static string IndividualWithIdNotFound(int individualId) => $"Individual with id `{individualId}` not found.";
+
+      public static string TypeNotSupported(Type t)
+      {
+         return $"{t} is not currently been handled";
```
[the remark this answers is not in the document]
abdelr: @msevestre, This is used when switching on types for the DataFormatParameter. So, in some places we have different logic depending on the parameter type. Consider creating the options for a given row on the grid when clicking on the excel column cell. The options are all of the same type (i.e., if the row is a mapping, all options are mappings, if meta data all options are also meta data...). The way to create these options depends on the source object of the row so we switch logic based on the type of the source. Currently we have a default option that should not be hit since all possible options for the source type are listed in the cases. In the future there could be a new type but we could forget to add handling such a type here in the switch so the default will be hit showing the error type not supported. This is the context where the TypeNotSupported method is used.
I agree it could expect just the name of the type. Passing a string allows for any string though so the caller could pass the string he/she builds with some error and still this method will show the string passed. This is why it was a type, to ensure it is properly used. I changed it to string and rename it to "typeName" but I do not completely like it like that since again, it allows for non-intentional wrong use but I could live with it since it is more an error targeting developers than final users.

src/OSPSuite.Core/Import/DataFormatParameter.cs
```diff
@@ -0,0 +1,93 @@
...
+
+      public MetaDataFormatParameter(string columnName, string metaDataId, bool isColumn = true) : base(columnName)
+      {
+         MetaDataId = metaDataId;
+         IsColumn = isColumn;
+      }
+
+      public string MetaDataId { get; set; }
+
+      public bool IsColumn { get; set; }
```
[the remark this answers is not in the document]
abdelr: When the metadata information (ColumnName) changes it could allow for values coming from the project and not from the excel column. It means that if you modify an existing parameter to point the metadata to a value it also needs to change the IsColumn value. I think it would be better to have as a setter function that would do both together so you cannot break it from the outside, meaning when you set the value you specify if it comes from a column or a value and the setter sets both. The problem is that ColumnName is comming from the base class and do not have a setter method that will take two parameters so we could think this deeply or leave it like it is for now and later iterate on it.

src/OSPSuite.Infrastructure.Import/Core/Dimension.cs
```diff
@@ -1,7 +1,7 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
 
-namespace OSPSuite.Core.Importer
+namespace OSPSuite.Infrastructure.Import.Core
 {
    public class Dimension
```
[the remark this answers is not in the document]
abdelr: We don't. It was left behind when we removed everything from the old importer

src/OSPSuite.Infrastructure.Import/Core/DimensionHelper.cs
abdelr: We don't. It was left behind when we removed everything from the old importer

src/OSPSuite.Infrastructure.Import/Core/IDataSet.cs
abdelr: It will modify the data, so you cannot foreach on it

abdelr: It will modify the data, so you cannot foreach on it

src/OSPSuite.Infrastructure.Import/Core/IDataSet.cs
abdelr: It will modify the data, so you cannot foreach on it

src/OSPSuite.Infrastructure.Import/Core/IDataSet.cs
abdelr: It will modify the data, so you cannot foreach on it

src/OSPSuite.Infrastructure.Import/Core/IDataSource.cs
abdelr: Old attempt to start with the filters... already removed

src/OSPSuite.Infrastructure.Import/Core/IDataSet.cs
msevestre: I don't understand this
It's only an issue if you delete elements in the list. To which case, even the for will crash 
Please explain

src/OSPSuite.Assets/UIConstants.cs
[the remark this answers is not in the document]
msevestre: Type.ToString() will return the name with the namespace
For instance `t="OSPSuite.Core.Domain.UnitSystem.Unit"`. I don't think this is a message we want to show to the user
So my suggestion is to have one method call the other passing type.name as parameter
And then correct the grammar

src/OSPSuite.Infrastructure.Import/Core/DataSourceFileReaders/ExcelReader.cs
```diff
@@ -0,0 +1,182 @@
...
+      private IWorkbook _book;
+      private IEnumerator<ISheet> _sheetEnumerator;
+      private IEnumerator _rowEnumerator;
+      private bool _columnOffsetOn;
+      private int _columnOffset;
+      private bool _isTypeXlsx;
+      private IFormulaEvaluator _formulaEvaluator;
+
+      public ISheet CurrentSheet { get; set; } //why is this public even????
+      public List<string> CurrentRow { get; set; } = new List<string>();
```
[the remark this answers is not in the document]
georgeDaskalakis: I changed the setters to be private - since the setting is only being done within the class. CurrentSheet is public because the name of the sheet needs to be read from the client - I have removed the comment.

src/OSPSuite.Infrastructure.Import/Core/ExtendedColumn.cs
abdelr: Added a TODO on code

src/OSPSuite.Infrastructure.Import/Core/Mappers/IDataSetToDataRepositoryMapper.cs
georgeDaskalakis: out parameter changed to become second return value

src/OSPSuite.Infrastructure.Import/Core/IDataSet.cs
georgeDaskalakis: transformed into foreach loop, works fine

src/OSPSuite.Infrastructure.Import/Core/IDataSet.cs
georgeDaskalakis: transformed into foreach loop, works fine

src/OSPSuite.Core/Serialization/Xml/ColumnXmlSerializer.cs
```diff
@@ -0,0 +1,25 @@
...
+         Map(x => x.Name);
+         Map(x => x.Unit);
+      }
+
+      protected override void TypedDeserialize(Column objectToDeserialize, XElement outputToDeserialize, SerializationContext context)
+      {
+         base.TypedDeserialize(objectToDeserialize, outputToDeserialize, context);
+         
+         if (objectToDeserialize.ErrorStdDev == Constants.STD_DEV_GEOMETRIC)
+            objectToDeserialize.Unit = new UnitDescription("?");
```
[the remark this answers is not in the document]
georgeDaskalakis: constructor with no arguments used instead, it has the same result exactly

src/OSPSuite.Core/Import/DataFormatParameter.cs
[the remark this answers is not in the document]
abdelr: Check on AbstractColumnsDataFormat.Paramters

appveyor.yml
```diff
@@ -33,10 +33,10 @@ build:
   use_snupkg_format: true
   project: OSPSuite.Core.sln
 
-skip_branch_with_pr: true
+skip_branch_with_pr: false
```
msevestre, line 36: @abdelr @georgeDaskalakis  can we reset that to the default please?

appveyor.yml
```diff
@@ -33,10 +33,10 @@ build:
   use_snupkg_format: true
   project: OSPSuite.Core.sln
 
-skip_branch_with_pr: true
+skip_branch_with_pr: false
 
 pull_requests: 
-  do_not_increment_build_number: true     
+  do_not_increment_build_number: false
```
msevestre, line 39: @abdelr @georgeDaskalakis  can we reset that to the default please?

msevestre (changes requested):
One last change and we are good to merge

appveyor.yml
abdelr: @msevestre Done

appveyor.yml
abdelr: @msevestre Done

msevestre:
YEAH!!

PR #874, opened 2021-01-12
Fix

abdelr:


Yuri05:
still a lot of funny things happen in Unit/LLOQ-settings window. At least it seems not to crash anymore

PR #889, opened 2021-03-02
888 disable mouse scroll combo box

msevestre:


src/OSPSuite.UI/Controls/UxComboBoxEdit.cs
```diff
@@ -0,0 +1,12 @@
+﻿using DevExpress.XtraEditors;
+
+namespace OSPSuite.UI.Controls
+{
+   public class UxComboBoxEdit : ComboBoxEdit
```
msevestre, line 5: @abdelr @Yuri05 @georgeDaskalakis 
Please make sure that we never directly use ComboBoxEdit in the code. Instead we should use UxComboBoxEdit which ensure that MouseWheel behavior is disabled by defau;t

Yuri05 (approved):

PR #894, opened 2021-03-03
890 synonyms

msevestre:


src/OSPSuite.Core/Domain/UnitSystem/Dimension.cs
```diff
@@ -166,14 +165,15 @@ public Unit DefaultUnit
...
-      public IEnumerable<string> GetUnitNames()
-      {
-         return _units.Keys;
-      }
+      public IEnumerable<string> GetUnitNames() => _units.Keys;
 
       public Unit Unit(string name)
       {
-         return _units[name];
+         if (_units.Contains(name))
```
msevestre, line 172: @Yuri05 Unit by name. Check first if there is a unit with name (cheap) and return
Otherwise lookup by synonym
I wanted to cache this but decided otherwise because one can add synonym after the fact programatically

src/OSPSuite.Core/Domain/UnitSystem/Dimension.cs
```diff
@@ -199,10 +199,7 @@ public double UnitValueToBaseUnitValue(Unit unit, double valueInUnit)
          return (valueInUnit + unit.Offset) * unit.Factor;
       }
 
-      public Unit AddUnit(string unitName, double factor, double offset)
```
msevestre, line 202: not require anymore with optional parameters

src/OSPSuite.Core/Domain/UnitSystem/Dimension.cs
```diff
@@ -231,20 +228,13 @@ public void AddUnit(Unit unit)
...
-      {
-         return _units.Contains(unitName);
-      }
+      private bool hasSynonym(string unitName) =>  _units.Any(x => x.HasSynonym(unitName));
 
-      public bool HasUnit(Unit unit)
-      {
-         return unit != null && HasUnit(unit.Name);
-      }
+      public bool HasUnit(string unitName) => _units.Contains(unitName) || hasSynonym(unitName);
```
msevestre: Has unit: True if has unit or has one with synonym

msevestre (commented):
@Yuri05 I think this is enough from the infrastructure point of view

msevestre:
@Yuri05 Little bit more work to do here for R

src/OSPSuite.Core/Domain/UnitSystem/Unit.cs
```diff
@@ -21,14 +26,43 @@ public class Unit
...
+         internal set
+         {
+            _name = value;
+            addDefaultUnitSynonyms();
+         }
+      }
+
+      private void addDefaultUnitSynonyms()
+      {
+         if(!_name.Contains("�"))
```
msevestre, line 41: @Yuri05 I think this is better like this for our µ issue. 
We still support this for other use cases (e..g loading from XML) but i don't think we should have all those options directly in the XML. It's just too redundant

msevestre:
@Yuri05 This is ready for review

src/OSPSuite.Core/Domain/UnitSystem/Unit.cs
Yuri05: yep, good idea

Yuri05 (approved):

src/OSPSuite.Core/Domain/UnitSystem/Unit.cs
```diff
@@ -21,14 +26,57 @@ public class Unit
...
+         AddUnitSynonym(Name.Replace("�", "u"));
+         AddUnitSynonym(Name.Replace("�", "mc"));
+      }
+
+      private void addExponentSynonyms()
+      {
+         if (!(Name.Contains("�") || Name.Contains("�")))
+            return;
+
+         AddUnitSynonym(Name.Replace("�", "2").Replace("�", "3"));
```
msevestre, line 59: @Yuri05 I have added this too.
Theroreticallly, there are all the combination possible with u as well etc. and that's going to become quite crazy. so I think we should try to keep it simple and offer some variations but not all possible combinations

msevestre:
@Yuri05  MErging this for now. We can always add more rules later

PR #898, opened 2021-03-04
Fixes #897 export time profile changes unit

### LINKED ISSUE #897: Export time profile analysis change the units
see https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1640

msevestre:


src/OSPSuite.Core/Serialization/Chart/CurveChartXmlSerializer.cs
```diff
@@ -18,6 +19,14 @@ public override void PerformMapping()
          MapEnumerable(x => x.Axes, x => x.AddAxis);
          MapEnumerable(x => x.Curves, x => x.AddCurveIfConsistent);
       }
+
+      protected override void TypedDeserialize(TCurveChart curveChart, XElement outputToDeserialize, SerializationContext context)
+      {
+         base.TypedDeserialize(curveChart, outputToDeserialize, context);
+
+         //once deserialize, we need to make sure that the curves are aware of the axis dimensions
+         curveChart.SynchronizeDataDisplayUnit();
```
msevestre, line 28: @Yuri05 This was a fairly simple stuff.  Once we have deserialize the whole chart, we need to make sure we synchronize display unit with the axis. The data is created on the fly and the value is not set properly (only on reload)

PR #905, opened 2021-03-05
904 cannot calculate pk

msevestre:
Idea: Instead of throwing an error, we make sure that we resize the Array of PK-Parameters using the max `IndividualId` in the population

so id = 0,1,2,3 => we return 4
so id = 0,1,5=> we return 6 and 2 3 4 will be fill with NaN

### DIFF
--- a/src/OSPSuite.Core/Domain/Data/SimulationResults.cs
+++ b/src/OSPSuite.Core/Domain/Data/SimulationResults.cs
@@ -93,7 +93,7 @@ namespace OSPSuite.Core.Domain.Data
             foreach (var id in ids)
             {
                var individualResults = ResultsFor(id);
-               values.AddRange(individualResults?.ValuesFor(quantityPath)?? new float[numberOfValuesPerIndividual].InitializeWith(float.NaN));
+               values.AddRange(individualResults?.ValuesFor(quantityPath) ?? new float[numberOfValuesPerIndividual].InitializeWith(float.NaN));
             }
 
             return values.ToArray();
@@ -129,6 +129,17 @@ namespace OSPSuite.Core.Domain.Data
          lock (_locker)
          {
             return AllIndividualResults.Select(x => x.IndividualId).ToArray();
+         }
+      }
+
+      public virtual int NumberOfIndividuals
+      {
+         get
+         {
+            var allIndividualsIds = AllIndividualIds();
+            //+1 because individual ids are typically 0 based
+            var maxNumberOfIndividuals = allIndividualsIds.Any() ? allIndividualsIds.Max() + 1 : 0;
+            return Math.Max(Count, maxNumberOfIndividuals);
          }
       }
 
--- a/tests/OSPSuite.R.Tests/Services/PKAnalysisTaskSpecs.cs
+++ b/tests/OSPSuite.R.Tests/Services/PKAnalysisTaskSpecs.cs
@@ -5,6 +5,7 @@ using OSPSuite.BDDHelper.Extensions;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Data;
 using OSPSuite.Core.Domain.PKAnalyses;
+using OSPSuite.Core.Domain.Populations;
 using OSPSuite.R.Domain;
 
 namespace OSPSuite.R.Services
@@ -55,9 +56,7 @@ namespace OSPSuite.R.Services
          var c_max = _result.PKParameterFor(_outputPath, "C_max");
          c_max.Dimension.Name.ShouldBeEqualTo(Constants.Dimension.MOLAR_CONCENTRATION);
       }
-
    }
-
 
    public class When_exporting_the_pk_analysis_to_data_frame_with_value_in_another_unit : concern_for_PKAnalysisTask
    {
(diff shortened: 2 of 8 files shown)


src/OSPSuite.Core/Domain/Data/SimulationResults.cs
```diff
@@ -132,6 +132,17 @@ public virtual IReadOnlyList<int> AllIndividualIds()
...
       }
 
+      public virtual int NumberOfIndividuals
+      {
+         get
+         {
+            var allIndividualsIds = AllIndividualIds();
+            //+1 because individual ids are typically 0 based
+            var maxNumberOfIndividuals = allIndividualsIds.Any() ? allIndividualsIds.Max() + 1 : 0;
+            return Math.Max(Count, maxNumberOfIndividuals);
```
msevestre, line 142: Just in case the last items did not work, count could be >maxNumberOfIndividuals . we take the max to be sure

tests/OSPSuite.R.Tests/Services/PKAnalysisTaskSpecs.cs
```diff
@@ -126,4 +126,37 @@ public void should_be_able_to_calculate_the_normalized_values()
...
+         _pkAnalysis = sut.CalculateFor(new CalculatePKAnalysisArgs {Simulation = _simulation, SimulationResults = _result});
+      }
+
+      [Observation]
+      public void should_be_able_to_calculate_the_pk_parameters_for_the_successful_simulations()
+      {
+         _pkAnalysis.ShouldNotBeNull();
+         var values = _pkAnalysis.PKParameterFor(_outputPath, "C_max");
+         values.Count.ShouldBeEqualTo(5);
+         float.IsNaN(values.Values[3]).ShouldBeTrue();
```
msevestre, line 159: BOOM :NaN

msevestre:
@Yuri05 merging to add to R.
Please review

Yuri05:
@msevestre 2 questions:
1. What will happen if e.g. I have 5 Individuals (individual-IDs 0,1,2,3,4) and results were calculated for the individuals 0-3 (individual with Individuals id = 4 failed)? Will I still get 5 individuals (and the results for the last one with id=4 will be filled with NaNs)? Or will I just get 4 Individuals (0..3)?
2. What will happen if the individual-IDs are NOT zero based? This could be the case if e.g. the population is loaded from a population file where the IDs are not 0,1,... N-1.

msevestre:
@Yuri05 Good questions
1- because the results only have 4 individuals and the max entry in the array is 3, you will get an array of 4 entries (0..3) 
2- I am assuming it will fill up the 0 index with NaN.  

Are you saying that exporting population to csv creates id starting at 1?

msevestre:
If we want to go away from the index like implementation, we need to convert the float array to a hash structure of some kind. Right now, this is the most optimal way of accessing getting the data.
I am ok with having an extra row of NaN if the pop starts at 1 but I am surprised this would be the case
Here is an export from PK-Sim

image

Yuri05:
>Are you saying that exporting population to csv creates id starting at 1?

No. But user could open this population file, remove some individuals and load it then (so that Ids could be even something like 2, 20, 201,202). Or e.g. "manually" create a population via the R-toolbox with arbitrary individual ids.

>1- because the results only have 4 individuals and the max entry in the array is 3, you will get an array of 4 entries (0..3)

Hmm. So if the LAST individual of the population fails, it still will be missing in the results...
Isn't it better just to **set** the NumberOfIndividuals by caller (who "knows" all the individuals in the population)?

msevestre:
@Yuri05  we don't need to support people doing arbitrary stuff. 
This fixes ensures that when on result crashes, then the whole process does not crash. What you are suggesting above would have never worked before in the first place 

>So if the LAST individual of the population fails, it still will be missing in the results...

This is the way the RESULTS is implemented. Here we are dealing with what is happening AFTER the results are calculated

>Isn't it better just to set the NumberOfIndividuals by caller (who "knows" all the individuals in the population)?

Why?

msevestre:
>Isn't it better just to set the NumberOfIndividuals by caller

And in fact, this does not work
In your exampp,e,  20., 201, 202
The number of individual is 3. But because we are finding pk values by individual id, we **NEED** create an array of 202 entries

Maybe we can revisit this logic here but for now, just setting the NumberOfIndividuals would not work

Yuri05:
>just setting the NumberOfIndividuals would not work

Are you sure? Looking at the pkanalysistask: individual ids are taken from the simulation results, there is no Loop `0..numberOfIndividuals -1` or so. Or maybe I misunderstand the logic of PK parameters filling...
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/d587d61e47104e236f70770917e47b0e95bc9dd5/src/OSPSuite.Core/Domain/Services/PKAnalysesTask.cs#L44-L111

msevestre:
Yes I am sure :) 

>https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/d587d61e47104e236f70770917e47b0e95bc9dd5/src/OSPSuite.Core/Domain/Services/PKAnalysesTask.cs#L44-L111

and 

https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/d587d61e47104e236f70770917e47b0e95bc9dd5/src/OSPSuite.Core/Domain/Services/PKAnalysesTask.cs#L83

and especially 
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/d587d61e47104e236f70770917e47b0e95bc9dd5/src/OSPSuite.Core/Domain/QuantityPKParameter.cs#L47

msevestre:
We can bypass all of this is QauntityPkParameter does not sue a float array. Let's discuss this live

msevestre:
I am going to add a few more tests in 
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/pull/908

to convert scenario 1 and 2 and see what is happening

PR #908, opened 2021-03-08
Fixes #907 repdoruce the bug first

### LINKED ISSUE #907: Container criteria "in container" only applies for the children of the container
@PavelBal commented on [Thu Nov 26 2020](https://github.com/Open-Systems-Pharmacology/MoBi/issues/523)

I would expect that setting the container criteria for e.g. a reaction or an event to "Intracellular" would create the reaction/event in the container itself and its descendants. Instead, only children container become the new entity.

Example: I have created a reaction "InContainer" with container criteria "In container: Intracellular" and a reaction "MatchTag" with container criteria "tagged with: Intracellular". I also did the same for the events.

In the simulation, only the reaction "MatchTag" is created in the comparments "Intracellular", but not the reaction "InContainer". The same is valid for the events - event "MatchTagEvent" is created in the container "Intracellular", event "InContainerEvent" is created in the Molecule-container under "Intracellular.

image
[InContainerBug.zip](https://github.com/Open-Systems-Pharmacology/MoBi/files/5602700/InContainerBug.zip)

msevestre:


src/OSPSuite.Core/Domain/Descriptors/NotInContainerCondition.cs
```diff
@@ -17,7 +17,7 @@ public NotInContainerCondition(string tag) : base(tag, Constants.NOT_IN_CONTAINE
 
       public override bool IsSatisfiedBy(EntityDescriptor entityDescriptor)
       {
-         return !entityDescriptor.ParentContainerTags.Contains(Tag);
+         return !entityDescriptor.Tags.Contains(Tag) && !entityDescriptor.ParentContainerTags.Contains(Tag);
```
msevestre, line 20: same with the NotInContainer

tests/OSPSuite.Core.Tests/Domain/DescriptorCriteriaSpecs.cs
```diff
@@ -122,9 +122,9 @@ protected override void Context()
       }
 
       [Observation]
-      public void should_return_that_the_entity_does_not_satisfy_the_criteria()
+      public void should_return_that_the_entity_does_satisfy_the_criteria()
```
msevestre, line 125: Change of specified behavior

msevestre:
@Yuri05 Please review

PR #910, opened 2021-03-08
Fixes #909 aging data for calculation

### LINKED ISSUE #909: Population calculations and aging
@Yuri05 commented on [Wed Feb 05 2020](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/295)

Following the discussion in https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1406: should aging data be supported by the R-Toolbox?

At the moment population simulation without aging can be executed:
`runSimulation(sim, population)`
Should it be possible to incorporate aging data, e.g.
`runSimulation(sim, population, agingData)`?

Aging data would contain {Time, Value}-(table)-profiles of some simulation parameters.

---

@msevestre commented on [Wed Feb 05 2020](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/295#issuecomment-582386106)

Not for this version. It becomes very tricky with splitting on different core etc .

---

@msevestre commented on [Fri Mar 05 2021](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/295#issuecomment-791718727)

@Yuri05 I see that you put this as HIGH. Does it need to be implemented for this release?

---

@msevestre commented on [Fri Mar 05 2021](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/295#issuecomment-791721696)

table should have 4 columns

```
         dt.Columns.Add(Constants.Population.INDIVIDUAL_ID_COLUMN);
         dt.Columns.Add(Constants.Population.PARAMETER_PATH_COLUMN);
         dt.Columns.Add(Constants.Population.TIME_COLUMN);
         dt.Columns.Add(Constants.Population.VALUE_COLUMN);
```

msevestre:


src/OSPSuite.R/Domain/AgingData.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System.Data;
+using OSPSuite.Core.Domain;
+using OSPSuite.Utility.Extensions;
+
+namespace OSPSuite.R.Domain
+{
+   public class AgingData
```
msevestre, line 7: @Yuri05 It's very hard to create DataTable and have them usable in R
So I created this very stupid class that will need to be instantiated in R (via a wrapper)

src/OSPSuite.Core/Extensions/StringExtensions.cs
```diff
@@ -108,5 +108,14 @@ public static string StripUnit(this string stringToStrip)
...
       }
+
+      /// <summary>
+      /// Returns the give text within quotes. 
+      /// </summary>
+      /// <example>
+      /// InQuotes(test) => "test"
+      /// </example>
+      /// <param name="text">Text to encapsulate in quotes</param>
+      public static string InQuotes(this string text) => $"\"{text}\"";
```
Yuri05, line 119: Should we check here that `text` does not contains (double) quotes? 
At least **in MoBi** user can create parameters with double quotes in name; if such a parameter will be varied in a population, CSV export will be invalid then
Just created an iissue for that: https://github.com/Open-Systems-Pharmacology/MoBi/issues/530

msevestre: "" should be forbidden from parameter names. Why is this allowed in the first place..no idea

PR #915, opened 2021-03-11
860 mol weight

abdelr:
Modifies the editor so it accepts molecules and molWeights (by modifying the settings injected to the importer, any metadata could also act as the molecule). MetaData names for Molecule and MoleWeight are injected also so in case we have different names on MoBi and PKSim, it will be ok. At the import process, it checks the consistency of the data (Molecule / MolWeight) and assign the MolWeight to the DataRepository.


abdelr:
@msevestre @Yuri05 I think the mol weight logic is complete on Core now. We still need to use the proper settings from PKSim and MoBi but should not take long. A question here from our side would be: "Should we create new issues for the other repositories (PKSim, MoBi) cloning this issue?"

msevestre:
923 commit. This is where rebasing would be required IMO. Let's discuss about this tomorrow
And yeah, absolutely. One issue in each repo

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
```diff
@@ -38,6 +53,31 @@ public ModalImporterPresenter(IModalImporterView view, IDataSetToDataRepositoryM
...
+                  {
+                     if (string.IsNullOrEmpty(molecularWeightDescription))
+                     {
+                        molecularWeightDescription = moleculeDescription;
+                        dataRepo.ExtendedProperties.Add(new ExtendedProperty<string>() { Name = molWeightName, Value = moleculeDescription });
+                     }
+                     else
+                     {
+                        if (moleculeDescription != molecularWeightDescription)
+                           throw new InconsistenMoleculeAndMoleWeightException();
```
Yuri05, line 70: Lines 69-70 are **string** comparison, right?
I wonder if it will not make some trouble (e.g. same value but different notations (e.g. `420 g/mol <=> 4.2E+2 g/mol`) or minimal numercal differences (e.g. `420 g/mol <=> 419,9999999` g/mol).
I think **numerical** comparison with some predefined **comparison tolerance** would be better

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1210,6 +1211,7 @@ public static class Error
       public static readonly string InavlidErrorDimension = "The dimension of the error units must be the same as the dimension of the measurement units.";
       public static readonly string NaNOnData = "Data contains NaN values at imported columns. Select a different action for NaN values or clean your data.";
       public static readonly string UnsupportedFileFormat = "The file format is not supported";
+      public static readonly string InconsistenMoleculeAndMoleWeightException = "Molecule and Molecular Weight does not match. Please either edit your Molecule-Molecular Weight in your project or remove the Molecular Weight from your mappings";
```
msevestre, line 1214: DO NOT MATCH

What does `Molecule-Molecule Weight` means?

src/OSPSuite.Infrastructure.Import/Services/IDataImporter.cs
```diff
@@ -17,15 +17,19 @@ public interface IDataImporter
...
+         string molWeightName
       );
 
       IReadOnlyList<DataRepository> ImportFromConfiguration(
          ImporterConfiguration configuration,
          bool promptForConfirmation,
          IReadOnlyList<MetaDataCategory> metaDataCategories,
          IReadOnlyList<ColumnInfo> columnInfos,
-         DataImporterSettings dataImporterSettings
+         DataImporterSettings dataImporterSettings,
```
msevestre, line 30: I think it's time to create an configuration object where. with all those parameters aggregated together. We are repeating the same list of parameters twice already. Also I think it should be an cache<moleculeName, molweight> and not just a single value

src/OSPSuite.Infrastructure.Import/Services/IImporter.cs
```diff
@@ -119,6 +120,14 @@ IEnumerable<MetaDataMappingConverter> mappings
          })).ToList();
       }
 
+      public IReadOnlyList<IReadOnlyDictionary<string, string>> EnumerateMetaData(Cache<string, IDataSet> dataSets, IEnumerable<MetaDataMappingConverter> mappings)
```
msevestre, line 123: I don't understand this return structure. Why is this so complicated?

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -23,17 +23,21 @@ public class ColumnMappingPresenter : AbstractPresenter<IColumnMappingView, ICol
...
-         IMappingParameterEditorPresenter mappingParameterEditorPresenter
+         IMappingParameterEditorPresenter mappingParameterEditorPresenter,
+         IMetaDataParameterEditorPresenter metaDataParameterEditorPresenter
       ) : base(view)
       {
          _importer = importer;
          _mappingParameterEditorPresenter = mappingParameterEditorPresenter;
-         View.FillSubView(_mappingParameterEditorPresenter.BaseView);
+         _metaDataParameterEditorPresenter = metaDataParameterEditorPresenter;
+         View.FillMappingSubView(_mappingParameterEditorPresenter.BaseView);
```
msevestre, line 39: I do not think we use the term `SubView` anywhere else. Can we stick to View.
e.g. AddMappingView, AddMetaDataView etc... just for consistency sake

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -134,10 +139,26 @@ public void SetRawData(UnformattedData rawData)
 
       public void SetDescriptionForRow(ColumnMappingDTO model)
       {
-         var values = _metaDataCategories.FirstOrDefault(md => md.Name == model.MappingName)?.ListOfValues.Select(v => v.ToString());
+         var values = _metaDataCategories.FirstOrDefault(md => md.Name == model.MappingName)?.ListOfValues.Keys.Select(v => v);
```
msevestre, line 142: `select(v=>v)`` ???

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -134,10 +139,26 @@ public void SetRawData(UnformattedData rawData)
...
          _setDescriptionForRow(model, values != null && values.All(v => v != model.ExcelColumn));
       }
 
+      public void UpdateMetaDataForModel()
+      {
+         if (_currentModel == null)
+            return;
+         var metaData = _currentModel.Source as MetaDataFormatParameter;
```
msevestre, line 150: new line between 149 and 150

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -134,10 +139,26 @@ public void SetRawData(UnformattedData rawData)
...
+         if (metaData == null)
+         {
```
msevestre, line 152: remove curly braces since 2 lines before you are not using it

src/OSPSuite.UI/Views/Importer/MetaDataParameterEditorView.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace OSPSuite.UI.Views.Importer
+{
+   public partial class MetaDataParameterEditorView : BaseUserControl, IMetaDataParameterEditorView
+   {
+      private IMetaDataParameterEditorPresenter _presenter;
+
+      public MetaDataParameterEditorView()
+      {
+         InitializeComponent();
+         this.manualnputLayoutControlItem.Text = Captions.Importer.ManualInput;
```
msevestre, line 15: DO this in the InitializeResources Method

src/OSPSuite.UI/Views/Importer/ColumnMappingView.cs
```diff
@@ -74,29 +87,59 @@ private void queryDisplayText(QueryDisplayTextEventArgs e)
...
-
-         _presenter.SetSubEditorSettings(model);
-         return _repositoryItemPopupContainerEdit;
+         switch (model.Source)
+         {
+            case MappingDataFormatParameter _:
+               _presenter.SetSubEditorSettingsForMapping(model);
+               return _repositoryMappingPopupContainerEdit;
+            case MetaDataFormatParameter md:
+               var metaDataCategory = _metaDataCategories.FirstOrDefault(x => x.Name == model.MappingName);
```
msevestre, line 109: THis logic should be in presenter, not in view

src/OSPSuite.Presentation/Presenters/Importer/MetaDataParameterEditorPresenter.cs
```diff
@@ -0,0 +1,28 @@
...
+   public interface IMetaDataParameterEditorPresenter : IDisposablePresenter
+   {
+      string Input { get; }
+   }
+
+   public class MetaDataParameterEditorPresenter : AbstractDisposablePresenter<IMetaDataParameterEditorView, IMetaDataParameterEditorPresenter>, IMetaDataParameterEditorPresenter
+   {
+      public string Input => View.Input;
+
+      public void InitView()
```
msevestre, line 17: What is this empty method?

src/OSPSuite.Presentation/Presenters/Importer/InconsistenMoleculeAndMoleWeightException.cs
```diff
@@ -0,0 +1,10 @@
+﻿using OSPSuite.Assets;
+using System;
+
+namespace OSPSuite.Presentation.Presenters.Importer
+{
+   public class InconsistenMoleculeAndMoleWeightException : Exception
+   {
+      public InconsistenMoleculeAndMoleWeightException() : base(Error.InconsistenMoleculeAndMoleWeightException) { }
```
msevestre, line 8: You do not need to define your own exception all the time, especially if you are not catching them specifically

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
```diff
@@ -126,7 +124,14 @@ public void SetDataSource(string dataSourceFileName)
 
       public void ImportData(object sender, EventArgs e)
       {
-         OnTriggerImport.Invoke(this, new ImportTriggeredEventArgs { DataSource = _dataSource });
+         try
+         {
+            OnTriggerImport.Invoke(this, new ImportTriggeredEventArgs { DataSource = _dataSource });
+         }
+         catch (Exception exception)
```
msevestre, line 131: Catch all Exception? why?

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
```diff
@@ -38,6 +53,31 @@ public ModalImporterPresenter(IModalImporterView view, IDataSetToDataRepositoryM
...
+                  if (moleculeDescription != null)
+                  {
+                     if (string.IsNullOrEmpty(molecularWeightDescription))
+                     {
+                        molecularWeightDescription = moleculeDescription;
+                        dataRepo.ExtendedProperties.Add(new ExtendedProperty<string>() { Name = molWeightName, Value = moleculeDescription });
+                     }
+                     else
+                     {
+                        if (moleculeDescription != molecularWeightDescription)
```
msevestre, line 69: what is moleculeDescription? the name of the molecule? 
Why is this compared to the molecularWeight?

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -290,7 +315,11 @@ public IEnumerable<ImageComboBoxOption> GetAvailableRowsFor(ColumnMappingDTO mod
          {
             var metaDataCategory = _metaDataCategories.FirstOrDefault(md => md.Name == model.MappingName);
             if (metaDataCategory != null && metaDataCategory.ShouldListOfValuesBeIncluded)
+            {
+               if (!metaDataCategory.ListOfValues.Values.Contains(model.ExcelColumn))
+                  options.Add(new ImageComboBoxOption() { Description = model.ExcelColumn, ImageIndex = ApplicationIcons.IconIndex(ApplicationIcons.MetaData) });
```
msevestre, line 320: What is ImageComboBoxOption? Is it a devexpress control?

msevestre (changes requested):

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
abdelr: @Yuri05 You are definitely right. I changed it already so now it used the parsed double values.

src/OSPSuite.Assets/UIConstants.cs
abdelr: @msevestre Changed to "Molecule and Molecular Weight do not match. Please either edit your Molecule or your Molecular Weight in your project or remove the Molecular Weight from your mappings"

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
abdelr: @msevestre Right, fixed.

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
abdelr: @msevestre :o) ... Fixed

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
abdelr: @msevestre done

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
abdelr: done

src/OSPSuite.UI/Views/Importer/MetaDataParameterEditorView.cs
abdelr: done

src/OSPSuite.UI/Views/Importer/ColumnMappingView.cs
abdelr: Done

src/OSPSuite.Presentation/Presenters/Importer/MetaDataParameterEditorPresenter.cs
abdelr: Removed

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
abdelr: Fixed

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
abdelr: Description is the value of the molecular weight coming either from molecule or directly from molecularWeight metadata

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
abdelr: It is an entity defined in OSPSuite.Presentation.Views.Importer holding information for creating the image combo box (label, icon)

src/OSPSuite.Presentation/Presenters/Importer/InconsistenMoleculeAndMoleWeightException.cs
abdelr: It is catch now so we do need it.

src/OSPSuite.Infrastructure.Import/Services/IDataImporter.cs
abdelr: Done

src/OSPSuite.Infrastructure.Import/Core/DataImporterSettings.cs
```diff
@@ -16,6 +16,12 @@ public class DataImporterSettings : IWithIcon
       private string _token;
       private string _delimiter;
 
+      public string NameOfMetaDataHoldingMoleculeInformation { get; set; }
+
+      public string NameOfMetaDataHoldingMolecularWeightInformation { get; set; }
```
msevestre, line 21: CLEAR 👍

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
```diff
@@ -29,6 +29,11 @@ public DataRepository ConvertImportDataSet(IDataSource dataSource, int dataSetIn
          var configuration = dataSource.GetImporterConfiguration();
          var dataRepository = new DataRepository { Name = dataSource.NamesFromConvention().ElementAt(dataSetIndex) };
 
+         foreach (var metaDataDescription in dataSource.EnumerateMetaData().ElementAt(dataSetIndex))
```
msevestre, line 32: I don't understand this. DataSetIndex is the index in the EnumeratedMetaData. How does the caller knows about this? I would like to understand why we are passing indexes back and forth instead of object instances

src/OSPSuite.Infrastructure.Import/Services/IImporter.cs
msevestre: comment still applies here. The fact that were are passing such primitive values in and out results in indexes being pass along everywhere. This may work, but it is very complicated to reason about

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -185,8 +204,19 @@ public void UpdateDescriptrionForModel()
          _view.RefreshData();
          _view.CloseEditor();
       }
+      public bool ShouldManualInputOnMetaDataBeEnabled(ColumnMappingDTO model)
+      {
+         var metaDataCategory = _metaDataCategories.FirstOrDefault(x => x.Name == model.MappingName);
```
msevestre, line 209: We have extension methods (e..g FindByName) if the object Implements IWithName

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
msevestre: ok. Unsual to see the name of a user control here. What if it becomes a list or sthg. Typically this is what we call a DTO

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
```diff
@@ -38,6 +49,37 @@ public ModalImporterPresenter(IModalImporterView view, IDataSetToDataRepositoryM
...
+                        if (!string.IsNullOrEmpty(dataImporterSettings.NameOfMetaDataHoldingMolecularWeightInformation))
+                        dataRepo.ExtendedProperties.Add(new ExtendedProperty<string>() { Name = dataImporterSettings.NameOfMetaDataHoldingMolecularWeightInformation, Value = moleculeDescription });
+                     }
+                     else
+                     {
+                        double molWeight;
+                        double moleculeMolWeight;
+                        double.TryParse(moleculeDescription, out moleculeMolWeight);
+                        double.TryParse(molecularWeightDescription, out molWeight);
+                        if (Math.Abs(moleculeMolWeight - molWeight) > Constants.DOUBLE_RELATIVE_EPSILON)
```
msevestre, line 71: This might be too low of a tolerance no?

tests/OSPSuite.Presentation.Tests/Importer/Presenters/ModalImporterPresenterSpecs.cs
```diff
@@ -0,0 +1,138 @@
...
+
+         fixedValuesRetriever?.Invoke(category);
+
+         return category;
+      }
+   }
+
+   public class When_importing_data : ConcernForModalImporterPresenter
+   {
+      [TestCase]
```
msevestre, line 90: USe Obseration if possible unless you are doing RowTesting (for consistency)

msevestre (commented):
There is certainly a reason for the complex in and out structure. Can we simplify it somehow tso that we work with object instead of primitive structure. The goal is not to write the most efficient code in my opinion as this is not a critical path of our app. Passing around indexes in public methods looks suspicious to me. REturnring arrays of collection of string, string also looks suspsicisou to me.  I feel that a small abstraction is missing. I don't know what is it until I understand what we are trying to solve

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
abdelr: You are right, this is just a DTO. I renamed ImageComboBoxOption to RowOptionDTO. 

@msevestre @Yuri05: Constants.DOUBLE_RELATIVE_EPSILON is 1e-5. Four decimal digits might be restrictive indeed but it sounds also like the standard precision that I remember from statistics lessons at school. I guess it will depend on the standards you have defined already. Please, tell me the exact tolerance you have in mind and I will set it to the value proposed.

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
abdelr: Sounds correct... I am doing some refactoring here since indeed feels like there is a missing abstraction.

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
Yuri05: 1e-5 sounds good (is also what I use for comparisons in the most cases)

msevestre: My bad. I thought this was the .NET double tolerance.
WE should have somewhere a method that compares two doubles and returns of they are equal or not (might just be in PKSim). Let me check

abdelr:
@msevestre @Yuri05 I have refactored the whole EnumerateMetaData logic. I hope it is more readable now. It copies data now into a new class and so on but still, performance is far from being a problem as you suggested.

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
```diff
@@ -24,22 +30,55 @@ public ModalImporterPresenter(IModalImporterView view, IDataSetToDataRepositoryM
...
+                     dataRepo.ExtendedProperties.Add(new ExtendedProperty<string>() { Name = dataImporterSettings.NameOfMetaDataHoldingMolecularWeightInformation, Value = moleculeDescription });
+                  }
+                  else
+                  {
+                     double molWeight;
+                     double moleculeMolWeight;
+                     double.TryParse(moleculeDescription, out moleculeMolWeight);
+                     double.TryParse(molecularWeightDescription, out molWeight);
+                     if (Math.Abs(moleculeMolWeight - molWeight) > Constants.DOUBLE_RELATIVE_EPSILON)
+                        throw new InconsistenMoleculeAndMoleWeightException();
```
Yuri05, line 70: It's actually an _absolute_ error check, not _relative_. 
Does not really matter here because MW is >1 anyway (thus if the relative error is above tolerance => absolute error will be above tolerance as well)

@msevestre Don't we have a method `EqualWithinRelativeTolerance `somewhere? Or was it only in the test framework?

abdelr: @Yuri05 maybe this:
OSPSuite.Core.Domain.Services.ValueComparer.AreValuesEqual

I am committing the change next.

Yuri05: >OSPSuite.Core.Domain.Services.ValueComparer.AreValuesEqual

Yep, this one :)

abdelr: > > OSPSuite.Core.Domain.Services.ValueComparer.AreValuesEqual
> 
> Yep, this one :)

Good, I have just committed the fix. Thanks for the tip.

Yuri05: almost there. should be `if NOT AreValuesEqual` :)

abdelr: > almost there. should be `if NOT AreValuesEqual` :)

Sure... I updated with the solution now... I hope ;)

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -2237,9 +2239,17 @@ public static class ObservedData
       public static readonly string ObservedDataOrganDescription = "Organ where the data was measured";
       public static readonly string ObservedDataCompartmentDescription = "Compartment where the data was measured";
       public static readonly string ObservedDataMoleculeDescription = "Molecule for which the data was measured";
-      public static readonly string ORGAN = "Organ";
-      public static readonly string COMPARTMENT = "Compartment";
-      public static readonly string MOLECULE = "Molecule";
+      public static readonly string Organ = "Organ";
```
msevestre, line 2242: @abdelr @georgeDaskalakis I think that constants that are USED as key should not be defined in UIConstants but in Constants. Then they can have different captions in Observed Date. But we do query after Organ and cOmpartment in PKSim explicitly and I don't like that we are breaking our convention

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -105,5 +105,64 @@ public IEnumerable<string> NamesFromConvention()
...
+               var dataSet = sheet.Current.Data.ElementAt(index);
+               return new ImportedDataSet(
+                  _configuration.FileName,
+                  DataSets.Keys.ElementAt(sheetIndex),
+                  dataSet,
+                  NamesFromConvention().ElementAt(accumulatedIndexes + index),
+                  dataSet.EnumerateMetaData(_mappings)
+               );
+            }
+            else
```
msevestre, line 127: no need for else

src/OSPSuite.Infrastructure.Import/Services/IDataImporter.cs
```diff
@@ -1,10 +1,24 @@
...
 using OSPSuite.Infrastructure.Import.Core;
 using System.Collections.Generic;
+using System.Linq;
 using ImporterConfiguration = OSPSuite.Core.Import.ImporterConfiguration;
 
 namespace OSPSuite.Infrastructure.Import.Services
 {
+   public class ReloadDataSets
+   {
+      public IEnumerable<DataRepository> NewDataSets;
```
msevestre, line 11: {get;set} and in fact probably only get and set in constructor.
Always think immutable if possible

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
```diff
@@ -4,15 +4,22 @@
...
+using System.Linq;
 using ImporterConfiguration = OSPSuite.Core.Import.ImporterConfiguration;
 
 namespace OSPSuite.Presentation.Presenters.Importer
 {
    public interface IModalImporterPresenter : IDisposablePresenter
    {
-      (IReadOnlyList<DataRepository> DataRepositories, ImporterConfiguration Configuration) ImportDataSets(IImporterPresenter presenter, IReadOnlyList<MetaDataCategory> metaDataCategories, IReadOnlyList<ColumnInfo> columnInfos, DataImporterSettings dataImporterSettings);
+      (IReadOnlyList<DataRepository> DataRepositories, ImporterConfiguration Configuration) ImportDataSets(
+         IImporterPresenter presenter,
```
msevestre, line 18: Didn't we change this with a config object?

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
```diff
@@ -126,7 +124,14 @@ public void SetDataSource(string dataSourceFileName)
 
       public void ImportData(object sender, EventArgs e)
       {
-         OnTriggerImport.Invoke(this, new ImportTriggeredEventArgs { DataSource = _dataSource });
+         try
+         {
+            OnTriggerImport.Invoke(this, new ImportTriggeredEventArgs { DataSource = _dataSource });
+         }
+         catch (InconsistenMoleculeAndMoleWeightException exception)
```
msevestre, line 131: This is quite suprising here. You raise a event and you know that a listener will raise a specific error...🤔

src/OSPSuite.UI/Views/Importer/ColumnMappingView.cs
```diff
@@ -58,13 +63,21 @@ public ColumnMappingView(IImageListRetriever imageListRetriever)
...
-         _repositoryItemPopupContainerEdit.AllowDropDownWhenReadOnly = DefaultBoolean.True;
+         _repositoryMappingPopupContainerEdit.Buttons[0].Kind = ButtonPredefines.Combo;
+         _repositoryMappingPopupContainerEdit.PopupControl = _mappingPopupControl;
+         _repositoryMappingPopupContainerEdit.CloseOnOuterMouseClick = false;
+         _repositoryMappingPopupContainerEdit.QueryDisplayText += (o, e) => queryDisplayText(e);
+         _repositoryMappingPopupContainerEdit.CloseUp += (o, e) => closeUpMapping(e);
+         _repositoryMappingPopupContainerEdit.CloseUpKey = new KeyShortcut(Keys.Enter);
+         _repositoryMappingPopupContainerEdit.AllowDropDownWhenReadOnly = DefaultBoolean.True;
+
+         _repositoryMetaDataPopupContainerEdit.Buttons[0].Kind = ButtonPredefines.Combo;
```
msevestre, line 74: maybe create a method to avoid code duplication (same code exactly) just different control

msevestre (changes requested):
I think it's looking very good. 
Just the constant stuff needs to be changed really

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
georgeDaskalakis: fixed

src/OSPSuite.Infrastructure.Import/Services/IDataImporter.cs
georgeDaskalakis: fixed

src/OSPSuite.UI/Views/Importer/ColumnMappingView.cs
georgeDaskalakis: packed it in private initializeButton()

src/OSPSuite.Assets/UIConstants.cs
georgeDaskalakis: moved them all (apart from the description ones) to Core.Constants nad renamed them so that they are uppercase.

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
georgeDaskalakis: I am not sure how this should change.....should the listener actually take care of handling the exception right? so they are not coupled together?

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
georgeDaskalakis: yeap. the config object is DataImporterSettings, it was changed to also contain molweight, moleculeName and promptForConfirmation.

georgeDaskalakis (approved):

PR #917, opened 2021-03-12
Fixes #916 for all observers

### LINKED ISSUE #916: Do not create container observer in "ForAll" scenario for non floating molecules
@Yuri05 commented on [Fri Oct 23 2020](https://github.com/Open-Systems-Pharmacology/MoBi/issues/519)

If I export a simulation which contains proteins from PK-Sim to MoBi and recreate in MoBi, unnecessary empty molecule properties containers are created on organ level.
E.g. originally exported simulation looks like this (molecule properties for CYP3A4 only on **compartment** level, which is correct):
grafik
Creating new simulation from existing building blocks in MoBi produces empty CYP3A4-container on organ level:
grafik

If such a simulation is then exported back to PK-Sim: editing of relative expressions crashes
grafik
(I guess due to normalization, which tries to find relExp_Norm parameter in ALL CYP3A4-containers, which crashes then for those wrongly inserted empty containers.

---

@msevestre commented on [Fri Mar 12 2021](https://github.com/Open-Systems-Pharmacology/MoBi/issues/519#issuecomment-797751104)

Well well...it took me some times but it seems to be `by design`
In PKSim, observers are never ForAll. However, when we export to Mobi, some observers are set explicitly to ForAll=True
```
  public static IReadOnlyCollection<string> MoBiForAll => new[]
            {CONCENTRATION_IN_CONTAINER, FRACTION_EXCRETED_TO_URINE, TISSUE, PLASMA_PERIPHERAL_VENOUS_BLOOD, PLASMA_UNBOUND_PERIPHERAL_VENOUS_BLOOD};

```

One of them is Tissue
image

and that's why it is showing up here

I feel that ForAll should not apply to non floating molecule 

@Yuri05 @PavelBal @StephanSchaller Thoughts?

msevestre:


src/OSPSuite.Core/Domain/Services/ObserverBuilderTask.cs
```diff
@@ -68,7 +68,7 @@ public void CreateObservers(IBuildConfiguration buildConfiguration, IModel model
       private IEnumerable<IMoleculeBuilder> moleculeBuildersValidFor(MoleculeList moleculeList, IEnumerable<IMoleculeBuilder> allMolecules)
       {
          if (moleculeList.ForAll)
-            return allMolecules.Where(molecule => !moleculeList.MoleculeNamesToExclude.Contains(molecule.Name));
+            return allMolecules.Where(molecule => molecule.IsFloating && !moleculeList.MoleculeNamesToExclude.Contains(molecule.Name));
```
msevestre, line 71: @Yuri05 Hard for me to tell
But it feels like ForAll does not apply for non floating molecule. IF you want CYP etc,,, I am cool in saying they should be named excplitely

This is definitely how it was intended at the beginning. Wonder why we never saw it before

Yuri05: Hard to say for me if we should exclude non-floating molecules here...

msevestre:
I am going to close this. I don't think this makes sense to have it like that, but I am not sure if it makes sense to remove it now.
We need to think about this a bit more

PR #918, opened 2021-03-15
LLoq is used for threshold on the value

abdelr:
@msevestre @Yuri05 Lloq logic

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
@@ -154,9 +154,14 @@ namespace OSPSuite.Infrastructure.Import.Core.Mappers
 
       private double truncateUsingLLOQ(SimulationPoint value)
       {
-         if (value == null) return double.NaN;
-         if (double.IsNaN(value.Lloq)) return value.Measurement;
-         if (double.IsNaN(value.Measurement) || value.Measurement < value.Lloq) return value.Lloq / 2;
+         if (value == null) 
+            return double.NaN;
+         
+         if (double.IsNaN(value.Lloq)) 
+            return value.Measurement;
+         
+         if (double.IsNaN(value.Measurement) || value.Measurement < value.Lloq) 
+            return value.Lloq / 2;
 
          return value.Measurement;
       }
--- a/src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
@@ -97,14 +97,7 @@ namespace OSPSuite.Infrastructure.Import.Core.Mappers
          //loop over view rows to get the sorted values.
          foreach (var value in column.Value)
          {
-            var adjustedValue =
-               (value == null || (double.IsNaN(value.Value) && double.IsNaN(value.Lloq))) ?
-               double.NaN :
-               (
-                  double.IsNaN(value.Value) || (value.Value < value.Lloq) ?
-                  value.Lloq / 2 :
-                  value.Value
-               );
+            var adjustedValue = truncateUsingLLOQ(value);
             if (double.IsNaN(adjustedValue))
                values[i++] = float.NaN;
             else if (unit != null)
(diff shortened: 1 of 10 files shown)


src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
```diff
@@ -97,12 +97,20 @@ private void convertParsedDataColumn(DataRepository dataRepository, KeyValuePair
...
          {
-            if (value == null || double.IsNaN(value.Value)) //but we actually should not be allowing this at all right?
+            var adjustedValue =
+               (value == null || (double.IsNaN(value.Value) && double.IsNaN(value.Lloq))) ?
+               double.NaN :
+               (
+                  double.IsNaN(value.Value) || (value.Value < value.Lloq) ?
+                  value.Lloq / 2 :
+                  value.Value
+               );
```
Yuri05, line 107: It seems it will crash if `value.Value` is **not NaN** but `value.Lloq `**is NaN**?
Also I would move this into a small function; with all those nested ifs is a bit hard to understand what happens.

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
```diff
@@ -97,12 +97,20 @@ private void convertParsedDataColumn(DataRepository dataRepository, KeyValuePair
          //loop over view rows to get the sorted values.
          foreach (var value in column.Value)
          {
-            if (value == null || double.IsNaN(value.Value)) //but we actually should not be allowing this at all right?
+            var adjustedValue =
```
msevestre, line 100: This is too hard to understand here. I have no idea what is going on

msevestre (changes requested):
The logic should be simplified

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
abdelr: @Yuri05 Should make more sense now

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
abdelr: @msevestre  Should make more sense now

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
```diff
@@ -151,6 +152,15 @@ private void convertParsedDataColumn(DataRepository dataRepository, KeyValuePair
          dataRepository.Add(dataColumn);
       }
 
+      private double truncateUsingLLOQ(SimulationPoint value)
+      {
+         if (value == null) return double.NaN;
```
msevestre, line 157: Let's not reutrn on the same line
I was hoping that Resharper file formatting should catch this.
so 
if(value==null)
  return double.NaN;

tests/OSPSuite.Infrastructure.Tests/Import/concern_for_DataSetToDataRepositoryMapperSpecs.cs
```diff
@@ -0,0 +1,130 @@
...
+         A.CallTo(() => dimensionFactory.DimensionForUnit(A<string>.Ignored)).Returns(Constants.Dimension.NO_DIMENSION);
+         sut = new DataSetToDataRepositoryMapper(dimensionFactory);
+      }
+   }
+
+   public class When_mapping_a_data_repository : concern_for_DataSetToDataRepositoryMapperSpecs
+   {
+      protected override void Because()
+      {
+         base.Because();
```
msevestre, line 119: base.Becasue does not make sense.
You are testing a SINGLE Action in your test. If there is something happening in base.because, it should be in the context instead

msevestre (changes requested):

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
abdelr: I need to check if I still have Resharper working with the file formatting. I know I installed VisualStudio Community (not preview anymore) at some point and not sure if I updated everything. I will check today. Thanks for the heads up

abdelr: Return at new line is done now

tests/OSPSuite.Infrastructure.Tests/Import/concern_for_DataSetToDataRepositoryMapperSpecs.cs
abdelr: Sure, thanks for the catch.

Yuri05 (approved):

msevestre:
@abdelr Let's create an issue first and then reference this issue in the commit for the next one

PR #921, opened 2021-03-17
Fixes #920 load observed data from R API

### LINKED ISSUE #920: Load ObservedData from the R API

msevestre:
This is a first try to export DataRepository (e.g. ObservedData) to R. 
Instead of implementing a parser in R, we can rely on the existing object structure. There is now a method LoadDataRepository(xx) that takes a pkml file and returns a DataRepository

We just need now to define the R interface and we can have it even for this version


msevestre:
@Yuri05 Going to merge this so that I can update R and PKSim/MoBi and run the tests

Yuri05 (commented):
approved

PR #927, opened 2021-03-19
785 importer reload redesign

georgeDaskalakis:


src/OSPSuite.UI/Services/DataImporter.cs
```diff
@@ -121,7 +128,8 @@ DataImporterSettings dataImporterSettings
                importerPresenter.LoadConfiguration(configuration);
                using (var importerModalPresenter = _container.Resolve<IModalImporterPresenter>())
                {
-                  return importerModalPresenter.ImportDataSets(importerPresenter, metaDataCategories, columnInfos, dataImporterSettings).DataRepositories;
+                  return importerModalPresenter.ImportDataSets(importerPresenter, metaDataCategories, columnInfos, dataImporterSettings)
```
Yuri05, line 131: here and in other functions: why are `IModalImporterPresenter`, `IImporterPresenter`, ... created via `_container.Resolve` and not via constructor injection?

src/OSPSuite.Presentation/Presenters/Importer/IImporterReloadPresenter.cs
```diff
@@ -0,0 +1,54 @@
...
+   {
+
+
+      public ImporterReloadPresenter(
+         IImporterReloadView view) : base(view)
+      {
+      }
+
+      public void Dispose()
+      {
```
msevestre, line 25: Do you need to defined a dispose method empty?

src/OSPSuite.UI/Services/DataImporter.cs
```diff
@@ -121,7 +128,8 @@ DataImporterSettings dataImporterSettings
                importerPresenter.LoadConfiguration(configuration);
                using (var importerModalPresenter = _container.Resolve<IModalImporterPresenter>())
```
msevestre, line 129: Use _applicationControl to retrieve modal Presenter and don't use the _contaienr directly (might be from old code)

src/OSPSuite.UI/Services/DataImporter.cs
msevestre: They are modal presenters and should be disposed explicitly. This is correct but one should use _applicationController instead ;. Not sure why we need the ref to _container at all

src/OSPSuite.UI/Services/DataImporter.cs
```diff
@@ -169,5 +177,45 @@ DataImporterSettings dataImporterSettings
...
+         var newDataSets = dataSetsToImport.Where(dataSet => !repositoryExistsInList(existingDataSets, dataSet));
+         result.Add(Constants.ImporterConstants.DATA_SETS_TO_BE_NEWLY_IMPORTED, newDataSets);
+
+         var dataSetsToBeDeleted = existingDataSets.Where(dataSet => !repositoryExistsInList(dataSetsToImport, dataSet));
+         result.Add(Constants.ImporterConstants.DATA_SETS_TO_BE_DELETED, dataSetsToBeDeleted);
+
+         var overwrittenDataSets = dataSetsToImport.Except(newDataSets);
+         result.Add(Constants.ImporterConstants.DATA_SETS_TO_BE_OVERWRITTEN, overwrittenDataSets);
+
+         using (var reloadPresenter = _applicationController.Start<IImporterReloadPresenter>())
```
msevestre, line 195: to use like this

src/OSPSuite.UI/Services/DataImporter.cs
```diff
@@ -169,5 +177,45 @@ DataImporterSettings dataImporterSettings
             return result;
          }
       }
+
+      public Cache<string, IEnumerable<DataRepository>> ReloadFromConfiguration(IEnumerable<DataRepository> dataSetsToImport,
```
msevestre, line 181: Can those input parameters be IReadOnly?

msevestre, line 181: Why don't you return a triple with 3  names parameters instead of relying on a cache with keys?

src/OSPSuite.UI/Services/DataImporter.cs
```diff
@@ -169,5 +177,45 @@ DataImporterSettings dataImporterSettings
...
+               };
+         }
+
+         return result;
+      }
+
+
+      private bool repositoryExistsInList(IEnumerable<DataRepository> dataRepositoryList, DataRepository targetDataRepository)
+      {
+         return dataRepositoryList.Select(dataRepo => targetDataRepository.ExtendedProperties.KeyValues.All(keyValuePair => dataRepo.ExtendedProperties[keyValuePair.Key].ValueAsObject.ToString() == keyValuePair.Value.ValueAsObject.ToString())).Any(result => result);
```
msevestre, line 217: Why the ToString() in the comparison
Why the any(x=>x). I don't understand the construct x=>x
This can be rewritten with dataRepositoryLiust.Any(x=>xxxx)

tests/OSPSuite.UI.Tests/Services/DataImporterSpecs.cs
```diff
@@ -0,0 +1,175 @@
...
+using OSPSuite.Infrastructure.Import.Services;
+using OSPSuite.Presentation.Core;
+using OSPSuite.Presentation.Presenters.Importer;
+using OSPSuite.Utility.Collections;
+
+namespace OSPSuite.UI.Services
+{
+   public abstract class concern_for_DataImporter : ContextSpecification<IDataImporter>
+   {
+      protected Utility.Container.IContainer _container;
```
msevestre, line 21: contaienr probably not required

msevestre (changes requested):

src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -601,6 +601,10 @@ public static class ChartFontOptions
       public static class ImporterConstants
       {
          public static readonly string[] NAMING_PATTERN_SEPARATORS = { ".", ",", "-", "_" };
+
+         public const string DATA_SETS_TO_BE_DELETED = "toDeleteDataSets";
```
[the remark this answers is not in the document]
georgeDaskalakis: Those are constants that are used for the Cache that ReloadFromConfiguration is returning. If we switch back to a triple they will no longer be necessary, otherwise I guess they are in the correct place, right?

src/OSPSuite.Presentation/Presenters/Importer/IImporterReloadPresenter.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System.Collections.Generic;
+using OSPSuite.Presentation.Views.Importer;
+
+namespace OSPSuite.Presentation.Presenters.Importer
+{
+   public interface IImporterReloadPresenter : IDisposablePresenter, IPresenter<IImporterReloadView>
```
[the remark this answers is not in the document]
georgeDaskalakis: done

src/OSPSuite.Presentation/Presenters/Importer/IImporterReloadPresenter.cs
georgeDaskalakis: removed

src/OSPSuite.UI/Services/DataImporter.cs
georgeDaskalakis: To be honest that was the original way I implemented this, but I found the solution with th Cache a bit more readable. As far as I understand your argument is about stability right? I could go back to having the triple...

georgeDaskalakis: done

src/OSPSuite.UI/Services/DataImporter.cs
georgeDaskalakis: substituted with applicationController.Start<>(). container remove since this was the only place it was used

src/OSPSuite.UI/Services/DataImporter.cs
georgeDaskalakis: done

src/OSPSuite.UI/Services/DataImporter.cs
georgeDaskalakis: done

tests/OSPSuite.UI.Tests/Services/DataImporterSpecs.cs
georgeDaskalakis: removed

src/OSPSuite.UI/Services/DataImporter.cs
georgeDaskalakis: corrected.

src/OSPSuite.UI/Services/DataImporter.cs
georgeDaskalakis: done

src/OSPSuite.Core/Domain/Constants.cs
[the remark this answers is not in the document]
georgeDaskalakis: removed, since the return value type has changed

Issue #925, opened 2021-03-18
Always select the first sheet upon confirmation


msevestre:
Otherwise the view looks empty and you have no idea what to do
![image](https://user-images.githubusercontent.com/1041237/111694133-dedb6180-8807-11eb-9203-4ad70d8d1e2a.png)

Please make sure it is like so

![image](https://user-images.githubusercontent.com/1041237/111694170-edc21400-8807-11eb-8e8c-9a621b042e05.png)

/cc @Yuri05

msevestre:
This was fixed too.

PR #928, opened 2021-03-19
#925 Always select the first dataset at confirmation view

abdelr:


src/OSPSuite.UI/Views/Importer/ImportConfirmationView.cs
```diff
@@ -126,7 +127,7 @@ private void onDataSetNameSelected(object sender, EventArgs eventArgs)
 
          if (listBox.SelectedValue == null) return;  //here if null we have to empty the GUI
 
-         _presenter.DataSetSelected(listBox.SelectedValue.ToString(), listBox.SelectedIndex);
+         _presenter.DataSetSelected(listBox.SelectedIndex);
```
msevestre, line 130: Is there always one Selected by default in the listBox?

abdelr: I am not sure I understand the question. This method is the handler for the SelectedIndexChanged event on the ListBox which also is restricted to select a single element. If nothing is selected then line 128 will catch it and prevent the null to propagate.

msevestre: What I mean is that I don't see the code that selects the first item by default. That why I am asking if there is always one selected. 
It looks like the feature was already implemented but. It working? I don't see any new trigger so to speak

abdelr: I see, there are two changes in the commit. One is to add the feature (ImportConfirmationView.cs line 88). The rest is to remove redundant parameters on functions.

msevestre: oh ok gotcha. It's just bacause it was called listBox and not like the name of the control before.
Any reason to cast sender in the onDataSetNameSelected  instead of using the list directly? I understand when you have multiple event subscribing to the same method. in this case..seems I am not sure what is the advantage.

Merging

PR #934, opened 2021-03-25
Fixes #933 dimension task returns units

### LINKED ISSUE #933: Ensure that dimension task also returns units for a dimension

msevestre:


msevestre:
@Yuri05 Trivial. Merging

PR #946, opened 2021-04-08
Concurrent simulation runner

abdelr:


Yuri05:
@abdelr @PavelBal Can you give an example R code how this "concurrent" simulation run will be used in R?

PavelBal:
@Yuri05 here just an example how it could look like (this code is for the older version and the signatures of the methods are different, but the idea from `@examples` should be clear)

```
#' @title  Runs a set of simulations (only individual simulations) and returns a list of \code{SimulationResults}
#'
#' @param simulations A list of \code{Simulation} objects to simulate.
#' @param simulationRunOptions Optional instance of a \code{SimulationRunOptions} used during the simulation run
#'
#' @return A list of \code{SimulationResults} objects in the order corresponding to the order of simulations.
#'
#' @examples
#' simPath <- system.file("extdata", "simple.pkml", package = "ospsuite")
#' sim <- loadSimulation(simPath)
#' sim2 <- loadSimulation(simPath)
#' sim3 <- loadSimulation(simPath)
#' results <- runSimulationsConcurrently(list(sim, sim2, sim3))
#' @export
runSimulationsConcurrently <- function(simulations, simulationRunOptions = NULL) {
  validateIsOfType(simulations, Simulation)
  validateIsOfType(simulationRunOptions, SimulationRunOptions, nullAllowed = TRUE)
  simulationRunner <- getNetTask("SimulationRunner")

  simulations <- c(simulations)

  #Create SimulationRunnerConcurrentOptions and add all simulations
  runConcurrentOptions <- SimulationRunnerConcurrentOptions$new()
  for (i in seq_along(simulations)){
    runConcurrentOptions$addSimulation(simulations[[i]])
  }

  results <- rClr::clrCall(simulationRunner, "RunConcurrently", runConcurrentOptions$ref)

  simResultsList <- vector("list", length(simulations))
  for (i in seq_along(simulations)){
    simResultsList[[i]] <- SimulationResults$new(results[[i]], simulations[[i]])
  }
  return(simResultsList)
}
```

Yuri05:
And should all simulations be executed only **once** or **multiple times with different parameter sets**?
Because in the latter case the current implementation would be very inefficient.

src/OSPSuite.Core/Domain/Services/SimModelBatch.cs
```diff
@@ -20,17 +20,59 @@ public class SimModelBatch : SimModelManagerBase, ISimModelBatch
...
+         _cloneManagerForModel = cloneManagerForModel;
       }
 
       public IReadOnlyList<string> VariableParameterPaths { get; private set; }
 
       public IReadOnlyList<string> VariableMoleculePaths { get; private set; }
 
+      public object Clone()
+      {
+         var xml = _simModelExporter.ExportSimModelXml(_modelCoreSimulation, new SimulationRunOptions().SimModelExportMode);
```
msevestre, line 41: wow.. what's going on here?😳
What are you trying to achieve? If feels like you are  not working at the right abstraction level here

src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
```diff
@@ -0,0 +1,61 @@
...
+      public ConcurrencyManager(ICoreUserSettings coreUserSettings)
+      {
+         _coreUserSettings = coreUserSettings;
+      }
+
+      public async Task<Dictionary<TData, TResult>> RunAsync<TData, TResult>(int numberOfCoresToUse, CancellationToken cancellationToken, ConcurrentQueue<TData> data, Func<int, CancellationToken, TData, Task<TResult>> action)
+      {
+         if (numberOfCoresToUse <= 0)
+            numberOfCoresToUse = _coreUserSettings.MaximumNumberOfCoresToUse;
+         numberOfCoresToUse = Math.Max(numberOfCoresToUse, data.Count);
```
Yuri05, line 38: Min, not Max?

src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
```diff
@@ -0,0 +1,61 @@
...
+         var results = new ConcurrentDictionary<TData, TResult>();
+         //Starts one task per core
+         var tasks = Enumerable.Range(0, numberOfCoresToUse).Select(coreIndex =>
+            Task.Run(async () =>
+            {
+               //While there is data left
+               while (data.TryDequeue(out var datum))
+               {
+                  //Invoke the action on it and store the result
+                  var result = await action.Invoke(coreIndex, cancellationToken, datum);
```
msevestre, line 49: the cancellationToken needs to be checked here I believe. It is the responsibility of the caller to throw if the action has been cancelled

src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
```diff
@@ -0,0 +1,61 @@
...
+                  results.TryAdd(datum, result);
+               }
+            }, cancellationToken)
+         ).ToList();
+
+         await Task.WhenAll(tasks);
+         //all tasks are completed. Can return results
+
+         return results.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
```
msevestre, line 58: why not return results? Just curious

src/OSPSuite.Core/Domain/Services/ISimModelBatch.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace OSPSuite.Core.Domain.Services
 {
-   public interface ISimModelBatch : IDisposable
+   public interface ISimModelBatch : IDisposable, ICloneable
```
msevestre, line 6: Let's remove this clone stuff please

src/OSPSuite.Core/Domain/Services/SimModelBatch.cs
```diff
@@ -20,17 +20,59 @@ public class SimModelBatch : SimModelManagerBase, ISimModelBatch
...
+         var simulation = _simModelSimulationFactory.Create();
```
msevestre, line 42: Remove this code

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -0,0 +1,74 @@
...
+      }
+
+      public void Clear()
+      {
+         _simulations.Clear();
+      }
+
+      public SimulationResults[] RunConcurrently()
+      {
+         var results = _concurrentManager.RunAsync<IModelCoreSimulation, SimulationResults>(
```
msevestre, line 64: should not we create the instance of ConcurrentMAnager here? why is it injected?

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -0,0 +1,74 @@
...
+            NumberOfCores,
+            new System.Threading.CancellationToken(false),
```
msevestre, line 66: what does the false do?

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -0,0 +1,74 @@
...
+
+      /// <summary>
+      /// Number of cores to use concurrently. Use 0 or a negative value for using the maximum available.
+      /// </summary>
+      int NumberOfCores { get; set; }
+
+      /// <summary>
+      /// List of Simulations
+      /// </summary>
+      void AddSimulation(IModelCoreSimulation simulation);
```
msevestre, line 26: reason why you are not passing the list of simulations in RunConcurrently?

msevestre (changes requested):
I think this is looking very good. A few questions

src/OSPSuite.Core/Domain/Services/ISimModelBatch.cs
abdelr: Yes sure, but the current code does not support running the SimulationBatch still. As soon as I complete the code for running also the SimulationBatch I will remove the clone and all the old support.

src/OSPSuite.Core/Domain/Services/SimModelBatch.cs
abdelr: Same here

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
abdelr: The cconcurrentSimulationRunner is never canceling the run since it will be used from R. I am not sure if this should be a feature (canceling the call). I am not sure either on how to do it from R in case it is possible. If this is something we should include, please tell me and I will include it. Still, the ConcurrencyManager does support canceling since it could be used from somewhere else.

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
abdelr: So you can prepare everything once combining IModelCoreSimulation objects and SimulationBatch objects. Even if you pass the lists (two instead of one) from R you will need to have Add functions on the settings object you will provide. Remember you cannot pass a list of objects directly from R. I think this way, having the wrapper object handling the settings through functions is more compact for the R caller.

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
abdelr: Why creating a new object every time. Also, with injection you decouple the implementations. Whoever is using the ConcurrentSimulationRunner could decide on the implementation of the ConcurrencyManager to use by just injecting it. Why should we create the object ourselves in the code and break the inversion of control?

src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
abdelr: You are right... adding the fix now

src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
abdelr: So we return a Task you could await. Again, from R you cannot await so the ConcurrentSimulationRunner does not use this feature but from other parts of the code you could have async/await

msevestre: This is not what I mean.results is already a dictionary. Why are your returning another dictionary based of this? The async/await is created for you automatically because you are using await /async in your code.

src/OSPSuite.Core/Domain/Services/ISimModelBatch.cs
msevestre: I can't merge the code until this is removed. So please remove, then we can merge, and then we add the other use case. Thanks

src/OSPSuite.Core/Domain/Services/SimModelBatch.cs
msevestre: Please remove, then we merge, then we implement the other use case

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
msevestre: You can absolutely pass array to R using args object as we discussed I believe. and if not, please see here
https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/develop/src/PKSim.R/Domain/IndividualCharacteristics.cs#L9

The problem with the implementation is that the caller needs to remember to clear. I find side effect-free stuff much easier to deal with in general. That's why the SimulationRunner takes the simulation as parameter. I am not sure why the ConcurrentSImulationRunner should have a different logic

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
msevestre: I did not say NEW. I say ensuring that we create a. new instance here so that there is no risk of memory leaks. Typically using a factory method to retrieve a new instance in a Dispose construct. That way, if something goes wrong, we clear after ourselves. 

So we do not create the object ourselves, we do not break inversion of control. We just ensure that we deal with memory leaks intrinsically by not holding references to objects

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
msevestre: My question is: What does the false do? :)

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
msevestre: But let's see what the interface in R will be to be honest. This should drive the .NET implementation

src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
abdelr: To the best of my knowledge, ConcurrentDictionary and Dictionary are different classes. They do not share a common interface or anything like that. We need a ConcurrentDictionary to avoid concurrent access issues but beyond ConcurrencyManager class, we do not need it anymore so we should turn it into a Dictionary. Similarly, we use a ConcurrentQueue. I have just updated the signature so it accepts a IEnumerable instead of a ConcurrentQueue since this data structure is only needed as a robust implementation but should not be affecting the signature of the method.

src/OSPSuite.Core/Domain/Services/ISimModelBatch.cs
abdelr: Done

src/OSPSuite.Core/Domain/Services/SimModelBatch.cs
abdelr: Done

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
abdelr: Initializes the state of the token as not cancelled.

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
abdelr: I agree. Sorry, I did not understood before. I will add the changes for this.

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
abdelr: What I meant by passing the list directly is that you have to create a settings object with add methods... so it means a bit more of wrapping. Maybe I am wrong and @PavelBal could help here but I think that you would normally prepare the settings once and use it several times so I do not see a need for clearing in between. Let us say you are in a PI scenario, you will need to prepare your simulations and then invoke the RunConcurrently again and again with the same simulations just modifying the values you want to set as inputs but the set of simulations should not change since your scenario has not changed at all. The way we have set the api now you would prepare your ConcurrentSimulationRunner once and use it once and again in every iteration by just specifying the input values to set to the simulation objects by directly setting this if use case 1 or by the use of the appropriate SimulationBatchRunValues.

src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
msevestre: >To the best of my knowledge, ConcurrentDictionary and Dictionary are different classes. They do not share a common interface or anything like that.

oh yeah they share IDictionary or IReadOnlyDictionary. 

IReadOnlyDictionary is what we should return in that case anyways so that the caller does not mutate (at least not without explicit casting) the results

image

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
msevestre: oh I see. Fair enough

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -0,0 +1,110 @@
...
+               new List<IModelCoreSimulation>(_simulations),
+               runSimulation
+            ).Result;
+            return _simulations.Select(s => results[s]).ToArray();
+         }
+
+         if (_simulationBatches.Count > 0)
+            return null;
+
+         return Enumerable.Empty<SimulationResults>().ToArray();
```
msevestre, line 100: Array.Empty :)

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -0,0 +1,110 @@
...
+      }
+
+      private async Task<SimulationResults> runSimulation(int coreIndex, CancellationToken cancellationToken, IModelCoreSimulation simulation)
+      {
+         return await Api.GetSimulationRunner().RunAsync(new SimulationRunArgs() { Simulation = simulation, SimulationRunOptions = SimulationRunOptions });
```
msevestre, line 105: np need to Async away. Just return the Task

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -0,0 +1,110 @@
...
+      }
+
+      public SimulationResults[] RunConcurrently()
+      {
+         //Currently we only allow for running simulations or simulation batches, but not both
+         if (_simulationBatches.Count > 0 && _simulations.Count > 0) 
+            //Temporal Exception. We should allow for mixing both use cases but we need to discuss first
+            throw new Exception("You already have Simulation and SimulationBatch objects and should not mix, please invoke Clear to start adding objects from a fresh start");
+
+         if (_simulations.Count > 0)
```
msevestre, line 86: Simulation.Count = 0
We should probably throw an exception

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -0,0 +1,110 @@
...
+         _simulations.Clear();
+         _simulationBatches.Clear();
+      }
+
+      public SimulationResults[] RunConcurrently()
+      {
+         //Currently we only allow for running simulations or simulation batches, but not both
+         if (_simulationBatches.Count > 0 && _simulations.Count > 0) 
+            //Temporal Exception. We should allow for mixing both use cases but we need to discuss first
+            throw new Exception("You already have Simulation and SimulationBatch objects and should not mix, please invoke Clear to start adding objects from a fresh start");
```
msevestre, line 84: Never throw Exception when coming from us. Use OSPSuiteException (unless you intend to see the whole call stack etc... Here this is clearly more for the user to see

Also error in Error files?

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -0,0 +1,110 @@
...
+      void Clear();
+
+      /// <summary>
+      /// Runs all preset settings concurrently
+      /// </summary>
+      /// <returns></returns>
+      SimulationResults[] RunConcurrently();
+   }
+
+   public class ConcurrentSimulationRunner : IConcurrentSimulationRunner
```
msevestre, line 47: I would make this guy umplement IDisposable and ensure that Dispose called Clear if not done already

This is the snippet we use

```

      private bool _disposed;

      public void Dispose()
      {
         if (_disposed) return;

         Cleanup();
         GC.SuppressFinalize(this);
         _disposed = true;
      }

      ~ConcurrentSimulationRunner ()
      {
         Cleanup();
      }

      protected virtual void Cleanup()
      {
      }
```

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -0,0 +1,110 @@
...
+         if (_simulationBatches.Count > 0 && _simulations.Count > 0) 
+            //Temporal Exception. We should allow for mixing both use cases but we need to discuss first
+            throw new Exception("You already have Simulation and SimulationBatch objects and should not mix, please invoke Clear to start adding objects from a fresh start");
+
+         if (_simulations.Count > 0)
+         {
+            var results = _concurrentManager.RunAsync(
+               NumberOfCores,
+               new CancellationToken(false),
+               new List<IModelCoreSimulation>(_simulations),
```
msevestre, line 91: Why do you pass a New List?

PR #954, opened 2021-04-20
953 enable s imulation batch with simulation concurrent runner

abdelr:
Check with @msevestre why ICoreUserSettings is not available through container on Core and restore ConcurrenceManager to use this settings instead of the hardcoded 2 on MaximumNumberOfCoresToUse

### DIFF
--- a/src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
+++ b/src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
@@ -23,17 +23,20 @@ namespace OSPSuite.Core.Domain.Services
 
    public class ConcurrencyManager : IConcurrencyManager
    {
-      private readonly ICoreUserSettings _coreUserSettings;
-
+      //private readonly ICoreUserSettings _coreUserSettings;
+      private int _maximumNumberOfCoresToUse = 2;
+      //TODO: Check with Michael why ICoreUserSettings is not available through container on Core
+      /*
       public ConcurrencyManager(ICoreUserSettings coreUserSettings)
       {
          _coreUserSettings = coreUserSettings;
       }
-
+      */
       public async Task<IReadOnlyDictionary<TData, TResult>> RunAsync<TData, TResult>(int numberOfCoresToUse, CancellationToken cancellationToken, IReadOnlyList<TData> data, Func<int, CancellationToken, TData, Task<TResult>> action)
       {
          if (numberOfCoresToUse <= 0)
-            numberOfCoresToUse = _coreUserSettings.MaximumNumberOfCoresToUse;
+            //    numberOfCoresToUse = _coreUserSettings.MaximumNumberOfCoresToUse;
+            numberOfCoresToUse = _maximumNumberOfCoresToUse;
          var concurrentData = new ConcurrentQueue<TData>(data);
          numberOfCoresToUse = Math.Min(numberOfCoresToUse, concurrentData.Count);
 
--- a/src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
+++ b/src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
@@ -23,7 +23,7 @@ namespace OSPSuite.Core.Domain.Services
 
    public class ConcurrencyManager : IConcurrencyManager
    {
-      private int _maximumNumberOfCoresToUse = Math.Max(1, Environment.ProcessorCount - 1);
+      private readonly int _maximumNumberOfCoresToUse = Math.Max(1, Environment.ProcessorCount - 1);
       public async Task<IReadOnlyDictionary<TData, TResult>> RunAsync<TData, TResult>(int numberOfCoresToUse, CancellationToken cancellationToken, IReadOnlyList<TData> data, Func<int, CancellationToken, TData, Task<TResult>> action)
       {
          if (numberOfCoresToUse <= 0)
@@ -49,6 +49,7 @@ namespace OSPSuite.Core.Domain.Services
          await Task.WhenAll(tasks);
          //all tasks are completed. Can return results
 
+         var tt = results.Values;
          return results;
       }
    }
--- a/src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
+++ b/src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
@@ -23,19 +23,10 @@ namespace OSPSuite.Core.Domain.Services
 
    public class ConcurrencyManager : IConcurrencyManager
    {
-      //private readonly ICoreUserSettings _coreUserSettings;
-      private int _maximumNumberOfCoresToUse = 2;
-      //TODO: Check with Michael why ICoreUserSettings is not available through container on Core
-      /*
-      public ConcurrencyManager(ICoreUserSettings coreUserSettings)
-      {
-         _coreUserSettings = coreUserSettings;
-      }
-      */
+      private int _maximumNumberOfCoresToUse = Math.Max(1, Environment.ProcessorCount - 1);
       public async Task<IReadOnlyDictionary<TData, TResult>> RunAsync<TData, TResult>(int numberOfCoresToUse, CancellationToken cancellationToken, IReadOnlyList<TData> data, Func<int, CancellationToken, TData, Task<TResult>> action)
       {
          if (numberOfCoresToUse <= 0)
-            //    numberOfCoresToUse = _coreUserSettings.MaximumNumberOfCoresToUse;
             numberOfCoresToUse = _maximumNumberOfCoresToUse;
          var concurrentData = new ConcurrentQueue<TData>(data);
          numberOfCoresToUse = Math.Min(numberOfCoresToUse, concurrentData.Count);
--- a/src/OSPSuite.R/Domain/SimulationBatch.cs
+++ b/src/OSPSuite.R/Domain/SimulationBatch.cs
@@ -48,17 +48,6 @@ namespace OSPSuite.R.Domain
       public double[] MoleculeValues => InitialValues.ToNetArray(InitialValue);
    }
 
-   public class SimulationBatchRunConcurrentlyOptions
-   {
-      private List<SimulationBatchRunValues> _simulationBatchRunValues = new List<SimulationBatchRunValues>();
-      public IReadOnlyList<SimulationBatchRunValues> SimulationBatchRunValues { get => _simulationBatchRunValues; }
-
-      public void AddSimulationBatchRunValues(SimulationBatchRunValues runValues)
-      {
-         _simulationBatchRunValues.Add(runValues);
-      }
-   }
-
    public class SimulationBatch : IDisposable
    {
       private readonly ISimModelBatch _simModelBatch;
--- a/src/OSPSuite.R/Domain/SimulationBatch.cs
+++ b/src/OSPSuite.R/Domain/SimulationBatch.cs
@@ -31,6 +31,9 @@ namespace OSPSuite.R.Domain
 
    public class SimulationBatchRunValues
    {
+      //Id to recognize it when running concurrently
+      public string Id { get; set; }
+
       //Potentially null
       public double[] ParameterValues { get; set; }
 
--- a/src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
+++ b/src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
@@ -18,22 +18,12 @@ namespace OSPSuite.R.Services
       public IModelCoreSimulation Simulation { get; set; }
       public List<SimulationBatchRunValues> SimulationBatchRunValues { get; } = new List<SimulationBatchRunValues>();
       public SimulationBatchOptions SimulationBatchOptions { get; set; }
-      private List<SimulationBatchRunValues> _pendingForInitialization = new List<SimulationBatchRunValues>();
-      private Dictionary<SimulationBatchRunValues, SimulationBatch> _simulationBatches = new Dictionary<SimulationBatchRunValues, SimulationBatch>();
-      public IReadOnlyDictionary<SimulationBatchRunValues, SimulationBatch> SimulationBatches => _simulationBatches;
+      internal List<SimulationBatchRunValues> _pendingForInitialization = new List<SimulationBatchRunValues>();
+      internal Dictionary<SimulationBatchRunValues, SimulationBatch> _simulationBatches = new Dictionary<SimulationBatchRunValues, SimulationBatch>();
       public void AddSimulationBatchRunValues(SimulationBatchRunValues simulationBatchRunValues)
       {
          SimulationBatchRunValues.Add(simulationBatchRunValues);
          _pendingForInitialization.Add(simulationBatchRunValues);
-      }
-      public void Initialize(IConcurrencyManager concurrentManager, int NumberOfCores, CancellationToken cancellationToken)
-      {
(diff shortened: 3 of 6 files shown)


src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
```diff
@@ -23,17 +23,20 @@ public interface IConcurrencyManager
 
    public class ConcurrencyManager : IConcurrencyManager
    {
-      private readonly ICoreUserSettings _coreUserSettings;
-
+      //private readonly ICoreUserSettings _coreUserSettings;
+      private int _maximumNumberOfCoresToUse = 2;
+      //TODO: Check with Michael why ICoreUserSettings is not available through container on Core
+      /*
       public ConcurrencyManager(ICoreUserSettings coreUserSettings)
```
msevestre, line 30: ICoreUserSettings is registered via the application. It does not make much sense to use sthg like this in R because we have no such settings. Probably this should not be a dependencies from the ConcurrencyManager?

src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
```diff
@@ -23,17 +23,20 @@ public interface IConcurrencyManager
...
       {
          _coreUserSettings = coreUserSettings;
       }
-
+      */
       public async Task<IReadOnlyDictionary<TData, TResult>> RunAsync<TData, TResult>(int numberOfCoresToUse, CancellationToken cancellationToken, IReadOnlyList<TData> data, Func<int, CancellationToken, TData, Task<TResult>> action)
       {
          if (numberOfCoresToUse <= 0)
```
msevestre, line 37: That should never happen? Or in this case, the numberOCoresToUse is the Max -1?

src/OSPSuite.R/Domain/SimulationBatch.cs
```diff
@@ -31,6 +31,9 @@ public class SimulationBatchOptions
 
    public class SimulationBatchRunValues
    {
+      //Id to recognize it when running concurrently
+      public string Id { get; set; }
```
msevestre, line 35: Who is setting this? 
How is this going to be used? I don't understand

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -13,6 +13,37 @@
...
+      public SimulationBatchOptions SimulationBatchOptions { get; set; }
+   }
+
+   public class ConcurrentSimulationResults
+   {
+      public ConcurrentSimulationResults(string simulationId, SimulationResults results, string additionalId = null)
+      {
+         SimulationId = simulationId;
+         SimulationResults = results;
+         AdditionalId = additionalId;
```
msevestre, line 40: I cannot understand why this is used from the name. Can you explain ?

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -13,6 +13,37 @@
...
+      }
+      public string SimulationId { get; private set; }
```
msevestre, line 42: remove all those private set ... they are redundant

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -13,6 +13,37 @@
...
+      public IModelCoreSimulation Simulation { get; set; }
+      public List<SimulationBatchRunValues> SimulationBatchRunValues { get; } = new List<SimulationBatchRunValues>();
+      public SimulationBatchOptions SimulationBatchOptions { get; set; }
+      public void AddSimulationBatchRunValues(SimulationBatchRunValues simulationBatchRunValues)
+      {
+         SimulationBatchRunValues.Add(simulationBatchRunValues);
+      }
+   }
+
+   class SimulationBatchRunOptions
```
msevestre, line 27: I am a bit confused where
We have 
```
SimulationBatchOptions
SimulationBatchRunValues
SimulationBatchRunOptions
SimulationBatch
SimulationWithBatchOptions
```

is this maybe too much ?

src/OSPSuite.R/Domain/SimulationBatch.cs
msevestre: Is this set maybe by the caller in R? What happened if this is not set?

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -80,35 +111,52 @@ public void Clear()
...
                runSimulation
-            ).Result;
-            return _simulations.Select(s => results[s]).ToArray();
+            ).Result.Values.ToArray();
          }
 
          if (_simulationBatches.Count > 0)
-            return null;
+         {
+            return _concurrentManager.RunAsync<SimulationBatchRunOptions, ConcurrentSimulationResults>(
```
msevestre, line 133: remove all those generic types They can be inferred by the compiler

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -80,35 +111,52 @@ public void Clear()
             _cancellationTokenSource.Cancel();
```
msevestre, line 111: here you can use _cancellationTokenSource?.Cancel()

msevestre (commented):
I think I understand what's going on but I feel that there are too many classes either named almost the same or doing the same thing. This will be abstracted away in R anyways so it's ok but I'd like to discuss if we really need all of this

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -80,35 +111,52 @@ public void Clear()
...
+            simulation.Id,
+            await Api.GetSimulationRunner().RunAsync(new SimulationRunArgs { Simulation = simulation, SimulationRunOptions = SimulationRunOptions })
+         );
+      }
+
+      private async Task<ConcurrentSimulationResults> runSimulationBatch(int coreIndex, CancellationToken cancellationToken, SimulationBatchRunOptions simulationBatchWithOptions)
+      {
+         return new ConcurrentSimulationResults(
+            simulationBatchWithOptions.Simulation.Id,
+            await Api.GetSimulationBatchFactory().Create(simulationBatchWithOptions.Simulation, simulationBatchWithOptions.SimulationBatchOptions).RunAsync(simulationBatchWithOptions.SimulationBatchRunValues),
```
msevestre, line 157: in fact this is not ok here. You are going to create one instance for each instances and thus creating the SImModel multiple times. This is not ok. The Batch should be created only once and then updated

msevestre (changes requested):

src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
abdelr: Ok, should we then have an extra class to handle Environment.ProcessorCount on R? Should we just accept what the user sends regardless the number of physical processors?

src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
abdelr: This is passed by the user. This way the user has control on its resources. You could have 8 cores but would like to use only 6 because you are doing something more. If you pass 0 or below it is interpreted as take all available.

src/OSPSuite.R/Domain/SimulationBatch.cs
abdelr: This is an Id that the caller from R should fill. This is what I explained to you after discussing with Pavel. The list of results is created in a way that you can infer the order but still could be a bit tricky and the best way is you assign a set of ids that you can later compare, so a given result will match the id of a given SimulationBatchRunValues.

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
abdelr: SimulationId <- Id of the simulation that was used for obtaining the results
SimulationResults <- Results after running the simulation
AdditionalId <- In case of Batch, the Id of the SimulationBatchRunValues

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
abdelr: Done

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
abdelr: SimulationBatchRunOptions is internal. It is needed because this is in the signature of the runSimulationBatch method. This method needs all the information on it but only accepts a generic data type in order to be used by the ConcurrencyManager.RunAsync method.

SimulationWithBatchOptions used to prepare the batches from the R side. This is the only new public class which is used for handling the information set from R.

SimulationBatchRunValues, SimulationBatch already existing on SimulationBatch

abdelr:
>Is this set maybe by the caller in R? What happened if this is not set?
It is set by the caller in R. I would say I should raise an exception in the case it is not set or already used.

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
abdelr: Done

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
abdelr: True, done.

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
abdelr: If we share the same copy of the SimModel then the simulation runs will interfere with each other. I think we discussed this before. It is actually similar to why we create a new SimulationRunner with Api.GetSimulationRunner() on every runSimulation method. You could try to keep a reference to the same SimulationBatch and run with different values concurrently and see the errors.

msevestre: I disagree with this. There is absolutely not advantage in that case to using SmulationBatch and we can remove it altogether.

abdelr: You mean that the time necessary for cloning the Simulation is actually higher than the gain of running concurrently I guess. You might be right in that. Removing it altogether is a bit too drastic I would say. What if we would create the SimulationBatch objects when they are first used (on each thread for better performance) and the references are kept until the ConcurrentSimulationRunner is disposed? That could be an intermediate solution, right? This way the first time you RunAsync will be a bit longer but still will worth if you plan to run the simulations a bunch of times. We could also take the initiative and create the SimulationBatches on a concurrent way as well when the method AddSimulationBatchOption is invoked. What do you think?

src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
msevestre: we always take AllAvailable -1 becuase otherwise your computer just freezes completely
So I would say we should stuck with this. This is also a constant in ospsuiteREnv

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
msevestre: Can it be "not" in the case of a batch?

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
msevestre: >You could try to keep a reference to the same SimulationBatch and run with different values concurrently and see the errors

The problem you had before was that you were using the same instance of the SimModelBatch. here you should have  multiple instances. 
Otherwise, and I might be wrong, you will create one SimModelInstance for EACH simulationrun. The gain is then very very limited

msevestre: >What if we would create the SimulationBatch objects when they are first used (on each thread for better performance) and the references are kept until the ConcurrentSimulationRunner is disposed?

This is what I thought you would do after our discussion on Monday. The best of both world. // run and optimized for speed

abdelr: Ok, then I did not understand correct at some point but now that I fully understand, I will do the changes next. I will let you know when the implementation is ready. Thanks for the discussion... was a good feedback for me.

msevestre: Awesome, We can discuss about this tomorrow if you want

abdelr: Sure, I have just committed a possible solution where I keep a list of objects that needs the SimulationBatch to be created. These SimulationBatch objects are only created once before they will be used for the first time and using also the ConcurrencyManager. The only problem is that the cores are not used as efficient as possible. The current solution is consistent in the sense that the class SimulationWithBatchOptions handles the Batches which makes sense since this is the class where you add new SimulationBatchRunValues so you can easily keep track of it. The problem is that if we would move the logic for initializing the batches to ConcurrentSimulationRunner, then the cores could be used in a more performant way but if we do so, I think the solution will be more fragile. If the R caller uses the wrong combination of calls, it could crash since the logic is not well encapsulated. I would prefer to go for robustness now but we could discuss and maybe a couple of external eyes could help me see what I am lacking now. I hope this could be a good compromise for now.

abdelr: I also just improved the solution in terms of performance without making the information public in R. I did mark the data internal which is not very usual. Maybe we could discuss in our meeting in a few hours.

src/OSPSuite.R/Domain/SimulationBatch.cs
abdelr: Now the id is retrieved when the options are added

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
abdelr: Updated to what we discussed today

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -33,82 +74,140 @@ public interface IConcurrentSimulationRunner : IDisposable
...
       }
 
-      public SimulationResults[] RunConcurrently()
+      private Task initializeBatches()
+      {
+          return _concurrencyManager.RunAsync
+         (
+            NumberOfCores,
+            _cancellationTokenSource.Token,
+            _listOfSettingsForConcurrentRunSimulationBatch.SelectMany
```
msevestre, line 138: @abdelr A quick question. Why is it not  a `Select`?

msevestre (commented):
This looks **VERY** good. Just one question :)

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
abdelr: @msevestre I have just added a comment to explain on the code, so it will be there in three months when we do not remember why anymore. I think we could maybe do it on a single select with the sum of all needed batches over the different simulations but still we will need to know for each iteration what settings object it refers to so I would say it is maybe easier and clearer with SelectMany on the list of simulations, Select on the list of batches for the given simulation.

PR #965, opened 2021-04-26
fixes #964 molweight update

### LINKED ISSUE #964: When molecule changed in Observed Data Molweight is not updated
When the user changes the Molecule with which an observed data set is associated, the MolWeight for the the data set does not get updated, but stays set at the old value.

georgeDaskalakis:
fixes #964


src/OSPSuite.Core/Commands/ChangeObservedDataMetaDataCommand.cs
```diff
@@ -21,11 +22,21 @@ protected override void ExecuteWith(IOSPSuiteExecutionContext context)
...
          Description = Command.SetMetaDataChangedCommandDescription(_metaDataChanged.OldName, _metaDataChanged.OldValue, _metaDataChanged.NewName, _metaDataChanged.NewValue);
          SetBuildingBlockParameters(context);
          context.PublishEvent(new ObservedDataMetaDataChangedEvent(_observedData));
       }
 
+      private void updateMolecule(IOSPSuiteExecutionContext context)
+      {
+         if(_metaDataChanged.NewName != Constants.ObservedData.MOLECULE) return;
+
+         var service = context.Resolve<IObservedDataTask>();
```
msevestre, line 36: yes. I would argue that we should call service observedDataTask however:) not a big deal but like variable to be explicit

src/OSPSuite.Core/Domain/Services/ObservedDataTask.cs
```diff
@@ -36,6 +36,7 @@ public interface IObservedDataTask
       /// </summary>
       bool DeleteAll();
 
+      void UpdateMolWeight(DataRepository observedData);
```
msevestre, line 39: this would deserve a space between the definition (resharper to the rescue) but even more so  a summary explaining what this is going to do. Can you add this?

msevestre (changes requested):

src/OSPSuite.Core/Commands/ChangeObservedDataMetaDataCommand.cs
georgeDaskalakis: corrected

src/OSPSuite.Core/Domain/Services/ObservedDataTask.cs
georgeDaskalakis: formated with resharper and summary added

msevestre (commented):
looks good to me

PR #967, opened 2021-04-26
Removing deprecated NumberOfCores

abdelr:


src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -129,11 +122,13 @@ public void Clear()
          _cancellationTokenSource?.Cancel();
       }
 
+      private int NumberOfCores() => SimulationRunOptions?.NumberOfCoresToUse ?? 0;
```
msevestre, line 125: privave props are lowerCase
so getNumberOfCores, or numberOfCores etc..

msevestre (changes requested):

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
abdelr: @msevestre Fixed. I am configuring resharper next... my fault!!

PR #970, opened 2021-04-27
Support for Exception handling

abdelr:


PavelBal (approved):
The API is convenient to use from R.

src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
```diff
@@ -52,5 +88,37 @@ public class ConcurrencyManager : IConcurrencyManager
...
+         {
+            return new ConcurrencyManagerResult<TResult>
+            (
+               id(args),
+               false,
+               null,
+               e.Message
+            );
+         }
+         return new ConcurrencyManagerResult<TResult>
```
msevestre, line 115: why don't you return this from the try? 
Also can you not name the parameter `args`
Last, don't hesitate to name parameters instead of passing false, or true => success: true

src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
```diff
@@ -7,6 +7,22 @@
...
+      public string Id { get; }
+      public bool Succeeded { get; }
+      public string ErrorMessage { get; }
+      public TResult Result { get; }
+
+      public ConcurrencyManagerResult(string id, bool succeeded, TResult result, string error)
+      {
+         Id = id;
+         Succeeded = succeeded;
+         ErrorMessage = error;
```
msevestre, line 21: Either call the parameter errorMessage or the Member Error. No reason to have different names here.

src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
```diff
@@ -52,5 +88,37 @@ public class ConcurrencyManager : IConcurrencyManager
...
+               false,
+               null,
+               e.Message
+            );
+         }
+         return new ConcurrencyManagerResult<TResult>
+         (
+            id(args),
+            true,
+            simulationResult,
```
msevestre, line 119: It feels like two constructor. One with Id(artgs) results and there other one just with the id(args)and error instead of asking the caller to pass the right parameters. (e.g. "", false or null true etc..)

msevestre (changes requested):

msevestre (commented):
awesome 💪

PR #972, opened 2021-04-27
Adding a test logger button on the starter

abdelr:
@msevestre I have created the issue on Core since the implementation for the logger is here. I am adding the test logger to the starter. I added a button to trigger a log creation. You should notice that before clicking the button there is no log.txt file at all. Only when creating the logs, the file is created. Also, I have committed a fix. Serilog was not respecting the MinimumLogLevel. I fixed this but now we need to update PKSim and Mobi to add this change. I would think that either there is no log from the CLI (so no need for a file) or the log level was not high enough since default will filter debug and trace.

### DIFF
--- a/tests/OSPSuite.Starter/Views/TestView.cs
+++ b/tests/OSPSuite.Starter/Views/TestView.cs
@@ -58,7 +58,7 @@ namespace OSPSuite.Starter.Views
          startHistogramTestButton.Text = "Start Histogram Test";
          startMatrixTestButton.Text = "Start Matrix Test";
          startEmptyFormButton.Text = "Start Empty Form";
-         startLoggerButton.Text = "Strt Logger Test";
+         startLoggerButton.Text = "Start Logger Test";
       }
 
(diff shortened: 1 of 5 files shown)


tests/OSPSuite.Starter/Views/TestView.cs
```diff
@@ -57,6 +58,7 @@ public override void InitializeResources()
          startHistogramTestButton.Text = "Start Histogram Test";
          startMatrixTestButton.Text = "Start Matrix Test";
          startEmptyFormButton.Text = "Start Empty Form";
+         startLoggerButton.Text = "Strt Logger Test";
```
msevestre, line 61: "Start" and not "strt"

msevestre (changes requested):
Looks good. Just a typo to fix

tests/OSPSuite.Starter/Views/TestView.cs
msevestre: just did

Issue #982, opened 2021-04-29
Data import: Invalid units in column headers are not properly handled


abdelr:
@Yuri05 commented on [Mon Apr 26 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1717)

[004_TC_InvalidUnit.xlsx](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/6379940/004_TC_InvalidUnit.xlsx)
![grafik](https://user-images.githubusercontent.com/25061876/116144091-352ba200-a6dc-11eb-8b21-17cfddc4f275.png)

After import:
![grafik](https://user-images.githubusercontent.com/25061876/116144373-8e93d100-a6dc-11eb-89f0-b1af2d7acb7b.png)

---

@PavelBal commented on [Wed Apr 28 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1717#issuecomment-828432148)

Every dimension (or unit) that is not in the OSPS-Dimensions should be considered as invalid and the import is not possible until the user (manually) sets the unit.

PR #986, opened 2021-04-29
Do not accept invalid units #982

abdelr:
Do not accept invalid units #982

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Core/DataFormat/MixColumnsDataFormat.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataFormat/MixColumnsDataFormat.cs
@@ -1,6 +1,6 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Linq;
+using OSPSuite.Core.Domain.UnitSystem;
 using OSPSuite.Infrastructure.Import.Core.Extensions;
 using System.Text.RegularExpressions;
 using OSPSuite.Core.Import;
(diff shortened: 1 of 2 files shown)


Yuri05 (approved):

src/OSPSuite.Infrastructure.Import/Core/DataFormat/MixColumnsDataFormat.cs
```diff
@@ -39,6 +44,7 @@ protected override UnitDescription ExtractUnits(string description, IUnformatted
                .Substring(1, units.Length - 2) //remove the brackets
                .Trim()                         //remove whitespace
                .Split(',')                     //split comma separated list
+               .Where(unitName => _dimensionFactory.Dimensions.SelectMany(dimension => dimension.Units).Any(unitFromDimension => unitFromDimension.Name == unitName)) //only accepts valid units
```
msevestre, line 47: I do not think this is correct here. We have a lot of methods checking whether a unit is valid or not (taking care of case, synonyms etc...) 
This needs to change

msevestre (changes requested):
Checking of units is not what I expected. Look at existing method 

For instance
IDimension DimensionForUnit(string unitName)

if the retruend value is null., it is not a valid dimension.

src/OSPSuite.Infrastructure.Import/Core/DataFormat/MixColumnsDataFormat.cs
```diff
@@ -39,6 +44,7 @@ protected override UnitDescription ExtractUnits(string description, IUnformatted
                .Substring(1, units.Length - 2) //remove the brackets
                .Trim()                         //remove whitespace
                .Split(',')                     //split comma separated list
+               .Where(unitName => _dimensionFactory.DimensionForUnit(unitName) != null) //only accepts valid units
```
msevestre, line 47: The same logic should be used for other unit recognition by the way (in column headeR)
Is this the case @abdelr and @georgeDaskalakis

msevestre (commented):
merging. Please review my comment

Issue #985, opened 2021-04-29
Importer Configuration: form caption


Yuri05:
is "BaseModalView". Please change to "Import Observed data"
![grafik](https://user-images.githubusercontent.com/25061876/116527193-b486c480-a8da-11eb-80b3-3ce989049929.png)

PR #989, opened 2021-04-29
#985 form caption

abdelr:
#985 form caption


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -399,7 +399,7 @@ public static class Importer
          public static readonly string CloseAllTabsButThis = "close all tabs but this";
          public static readonly string CloseAllTabsToTheRight = "close all tabs to the right";
          public static readonly string UseFilterForImport = "Use the filters selected not only for visualization but also for importing the data";
-         public static readonly string Title = "Importer";
+         public static readonly string Title = "Import Observed data";
```
Yuri05, line 402: pls change to "Import Observed **D**ata"

abdelr: Done already ;)

PR #993, opened 2021-04-29
Bug fixes 

### LINKED ISSUE #951: Importer: Close tab button
Make sure that during import process, the user can close a tab ("x"-button on the sheet tab).

### LINKED ISSUE #955: Add comment to explain logic
@georgeDaskalakis So that I can close the PR
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/pull/947/files#diff-72b7c8018860de4ec865cdc18b6d1a3d619e7b35093ee685323b510e369bb8a6R198

### LINKED ISSUE #977: Data import: Typo
@Yuri05 commented on [Tue Apr 27 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1720)

grafik

### LINKED ISSUE #978: Importer Crashes when loading file with not supported format
Whe the user tries to import a file which format is not supported the importer crashes. This should not happen: instead the user should be informed through a message that the file does not comply to the supported formats and the import should not further continue.

georgeDaskalakis:
fixes #977, fixes  #955, fixes #951, fixes #978


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -406,6 +406,17 @@ public static class Importer
...
          public static readonly string GroupByTitle = "Group by";
          public static readonly string MappingTitle = "Mapping";
+         public static readonly string ReloadWillCauseChangeOfDataSets = "Reloading will cause the following changes in observed data. Do you really want to reload?";
+         public static readonly string DataSetsWillBeOverwritten = "Datasets that will be overwritten";
+         public static readonly string NewDataStetsWillBeImported = "New datasets that will be imported";
+         public static readonly string ReloadData = "Reload Data";
+
+         public static string FileFormatNotSupported(string fileName)
+         {
+            return "The file: \n \n" + fileName +
```
msevestre, line 416: we should use string interpolation here... $ {} or  string builder. Using + is just allocating strings for nothing all the time. Not a big deal, but we have been diligent everywhere else"

see
ShouldWatermarkBeUsedForChartExportToClipboard for example of string.builder

src/OSPSuite.Presentation/Presenters/Importer/ImporterDataPresenter.cs
```diff
@@ -119,6 +120,12 @@ public IDataSourceFile SetDataSource(string dataSourceFileName)
          if (string.IsNullOrEmpty(dataSourceFileName)) return null;
          Sheets = new Cache<string, DataSheet>();
          _dataSourceFile = _importer.LoadFile(_columnInfos, dataSourceFileName, _metaDataCategories);
+
+         if (_dataSourceFile == null)
```
msevestre, line 124: there is 
{
return null 
}

but the ImporterPresenter uses no { }  in the same PR
Let's be consistent and remove the { }

so 
```
if (_dataSourceFile == null)
  return null
```

msevestre (changes requested):

src/OSPSuite.Presentation/Presenters/Importer/ImporterDataPresenter.cs
georgeDaskalakis: fixed

src/OSPSuite.Assets/UIConstants.cs
georgeDaskalakis: done

Yuri05 (approved):

PR #994, opened 2021-04-29
950 fix

### LINKED ISSUE #950: Importer crashes if there is a missing mapped column
If we map an excel column that is not present in one of the sheets of the excel and then try to import that sheet we get an unhandled exception (s. screenshot). 

importer bug

I would suggest as a solutio to this issue to catch the exception, show an error message to the user and prevent the import.

georgeDaskalakis:
closes #950


src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
```diff
@@ -157,7 +157,7 @@ private void ImportSheetsFromDataPresenter(object sender, ImportSheetsEventArgs
             _configuration.LoadedSheets.AddRange(args.Sheets.Keys);
             _configuration.FilterString = args.Filter;
          }
-         catch (Exception e) when (e is NanException || e is ErrorUnitException)
+         catch (Exception e) when (e is NanException || e is ErrorUnitException || e is MissingColumnException)
```
msevestre, line 160: neat. Did not know that this was a thing. TIL `when`

PR #1004, opened 2021-05-03
Adding category to the format on serilog

abdelr:


msevestre:
ok going to merge this and I can try again

abdelr:
> ok going to merge this and I can try again

Also, notice that we can format the logs as we like. If at some point we need anything more, we should be able to add more information to the logs.

PR #1006, opened 2021-05-04
952 csv separator

georgeDaskalakis:
How do you find the use of char '/0' as a null return? I hope this comparison is not too C++ style...

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Services/ICsvSeparatorSelector.cs
+++ b/src/OSPSuite.Infrastructure.Import/Services/ICsvSeparatorSelector.cs
@@ -1,0 +1,10 @@
+﻿namespace OSPSuite.Infrastructure.Import.Services
+{
+   public interface ICsvSeparatorSelector
+   {
+      /// <summary>
+      /// Returns the separator to be used for the reading of the .csv file
+      /// </summary>
+      char GetCsvSeparator(string fileName);
+   }
+}
--- a/src/OSPSuite.Infrastructure.Import/Services/ICsvSeparatorSelector.cs
+++ b/src/OSPSuite.Infrastructure.Import/Services/ICsvSeparatorSelector.cs
@@ -5,6 +5,6 @@
       /// <summary>
       /// Returns the separator to be used for the reading of the .csv file
       /// </summary>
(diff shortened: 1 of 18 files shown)


src/OSPSuite.Infrastructure.Import/Services/ICsvSeparatorSelector.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace OSPSuite.Infrastructure.Import.Services
+{
+   public interface ICsvSeparatorSelector
+   {
+      /// <summary>
+      /// Returns the separator to be used for the reading of the .csv file
+      /// </summary>
+      char GetCsvSeparator(string fileName);
```
msevestre, line 8: why not char?

src/OSPSuite.UI/Services/CsvSeparatorSelector.cs
```diff
@@ -0,0 +1,27 @@
...
+         _applicationController = applicationController;
+      }
+
+      public char GetCsvSeparator(string fileName)
+      {
+         using (var csvSeparatorPresenter = _applicationController.Start<ICsvSeparatorSelectorPresenter>())
+         {
+            csvSeparatorPresenter.SetFileName(fileName);
+
+            return csvSeparatorPresenter.Canceled() ? '\0' : csvSeparatorPresenter.GetCsvSeparator();
```
msevestre, line 22: this char is weird, especially becuase it leaks outside of the implementation. REturn a char? and if it's null, it's not valid

msevestre (changes requested):

src/OSPSuite.Infrastructure.Import/Services/ICsvSeparatorSelector.cs
georgeDaskalakis: I like the suggestion for nullable types. I have changed it. Thanks Michael :)

src/OSPSuite.UI/Services/CsvSeparatorSelector.cs
georgeDaskalakis: done

PR #1008, opened 2021-05-05
999 reload

abdelr:
Reopen for keeping the code (Core and PKSim) in one place

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Services/IDataImporter.cs
+++ b/src/OSPSuite.Infrastructure.Import/Services/IDataImporter.cs
@@ -56,5 +56,13 @@ namespace OSPSuite.Infrastructure.Import.Services
       /// </summary>
       /// <returns>a list of meta data categories</returns>
       IList<MetaDataCategory> DefaultMetaDataCategories();
+
+      /// <summary>
+      /// Compares if two data repositories come from the same data
+      /// </summary>
+      /// <param name="sourceDataRepository"></param>
+      /// <param name="targetDataRepository"></param>
+      /// <returns></returns>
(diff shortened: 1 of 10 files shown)


src/OSPSuite.Infrastructure.Import/Services/IDataImporter.cs
```diff
@@ -56,5 +56,13 @@ DataImporterSettings dataImporterSettings
       /// </summary>
       /// <returns>a list of meta data categories</returns>
       IList<MetaDataCategory> DefaultMetaDataCategories();
+
+      /// <summary>
+      /// Compares if two data repositories come from the same data
+      /// </summary>
+      /// <param name="sourceDataRepository"></param>
```
msevestre, line 63: If you define entries in the comments,explain what they are or remove

src/OSPSuite.Infrastructure.Import/Services/IDataImporter.cs
```diff
@@ -56,5 +56,13 @@ DataImporterSettings dataImporterSettings
...
+      /// <param name="targetDataRepository"></param>
+      /// <returns></returns>
+      bool AreFromSameSource(DataRepository sourceDataRepository, DataRepository targetDataRepository);
```
msevestre, line 66: Isn't the name a bit misleading since it in fact SKIPS the file name (the source)

msevestre: and the code comment says "Ignore source" while the variable is called FILE lol. I think we should make all of that consistent

msevestre (changes requested):

PR #1010, opened 2021-05-05
fixe #1009

### LINKED ISSUE #1009: Error importing LLOQ values
@Yuri05 commented on [Tue Apr 27 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1719)

[030_TCUnitLLOQ.xlsx](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/6385579/030_TCUnitLLOQ.xlsx)
Imported LLOQ values are 100 greater than expected (0.5 instead of 0.005)
grafik
grafik

---

@georgeDaskalakis commented on [Fri Apr 30 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1719#issuecomment-829713154)

The reason for this error is that the values are read from the excel file as numbers whereas the LLOQ as a string. You will notice in your screenshots that 0.1 is with dot whereas the LLOQ <0,01 with comma. Interestingly, in the Core starter I do not get the issue with the specific file, but on PK-Sim I do. 

We can guarantee that we parse the numeric values with the same culture they are read from the excel with, but for the LLOQ that we are reading as a string we cannot do this. I could change the culture this is parsed, but I could break another file that way.

---

@Yuri05 commented on [Fri Apr 30 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1719#issuecomment-829943120)

> I could change the culture this is parsed, but I could break another file that way.

I would not do this. Instead, I would make 2 replacements before converting the string into numeric:
1. replace **point** with the current decimal separator 
2. replace **comma** with the current decimal separator 

Theoretically it still could be situations where this does not work (e.g. if you have "." as 1000-separator like "<2.000.000,1") - but as LLOQ are usually small number this will virtually never happen, so in the 99,99% of all cases the conversion above would work.

---

@msevestre commented on [Fri Apr 30 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1719#issuecomment-830098310)

>I could change the culture this is parsed, but I could break another file that way.

no. The culture is set explicitly to US in PKSim and MoBi. We should never change the culture otherwise

---

@georgeDaskalakis commented on [Wed May 05 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1719#issuecomment-832588608)

Implemented according to Juri's suggestion.

georgeDaskalakis:
fixes #1009

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
@@ -1,7 +1,6 @@
 ﻿using OSPSuite.Infrastructure.Import.Core.Extensions;
 using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.Linq;
 using OSPSuite.Utility.Extensions;
 using OSPSuite.Core.Domain;
@@ -265,6 +264,7 @@ namespace OSPSuite.Infrastructure.Import.Core.DataFormat
 
          var element = row.Data.ElementAt(columnDescription.Index).Trim();
          var unit = currentParameter.MappedColumn.Unit.ExtractUnit(columnName => data.GetColumnDescription(columnName).Index, row.Data);
+         
          if (double.TryParse(element, out var result))
             return new SimulationPoint()
             {
@@ -274,10 +274,7 @@ namespace OSPSuite.Infrastructure.Import.Core.DataFormat
             };
          if (element.StartsWith("<"))
          {
-            var separatorChar = Convert.ToChar(CultureInfo.CurrentCulture.NumberFormat.NumberDecimalSeparator);
-            
-            var lloqString = element.Substring(1).Replace(',', separatorChar).Replace('.',separatorChar);
-            double.TryParse(lloqString, out result);
+            result = element.Substring(1).ConvertedTo<double>();
             return new SimulationPoint()
             {
                Lloq = result,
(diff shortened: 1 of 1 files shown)


src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -273,7 +274,10 @@ private SimulationPoint parseMappingOnSameColumn(MappingDataFormatParameter curr
             };
          if (element.StartsWith("<"))
          {
-            double.TryParse(element.Substring(1), out result);
+            var separatorChar = Convert.ToChar(CultureInfo.CurrentCulture.NumberFormat.NumberDecimalSeparator);
+            
+            var lloqString = element.Substring(1).Replace(',', separatorChar).Replace('.',separatorChar);
+            double.TryParse(lloqString, out result);
```
msevestre, line 280: Isn't it weird to not verify that you can parse? What happens if this does not work? 
Also we have a method that converts a string to double 
https://github.com/Open-Systems-Pharmacology/OSPSuite.Utility/blob/master/src/OSPSuite.Utility/Extensions/ConversionExtensions.cs#L31

can this be used?

georgeDaskalakis: I was not aware of this function. Yes this is a better alternative, since it would fill the lloq value with NaN if it cannot be parsed.  In the current way the function is usd this is not a possibility, since the values of each column are checked in the beginning and if they are not numeric you cannot map them to measurement or error. But still it does not hurt to have even this possibility covered within the function.

PR #1028, opened 2021-05-12
1022 moving moBi excel import logic to core

### LINKED ISSUE #1022: Replace all the SmartXLS usages that have been left
@georgeDaskalakis commented on [Mon May 10 2021](https://github.com/Open-Systems-Pharmacology/MoBi/issues/559)

SmartXLS is still being used in MoBi for import of some things (like Molecule Start Values ). This should be replaced with the use of npoi.

georgeDaskalakis:
fixes #1022


src/OSPSuite.Infrastructure.Import/Core/DataSourceFileReaders/ExcelReader.cs
```diff
@@ -84,6 +84,21 @@ public bool MoveToNextRow()
...
+         for (var i = 0; i < _book.NumberOfSheets; i++)
+         {
+            if (!excludeEmptySheets || !isSheetEmpty(_book.GetSheetAt(i)))
+               yield return _book.GetSheetName(i);
+         }
+      }
+
+      private bool isSheetEmpty(ISheet sheet)
+      {
+
```
msevestre, line 98: can you run resharper please to style this file?

src/OSPSuite.Infrastructure.Import/Core/DataSourceFileReaders/ExcelReader.cs
```diff
@@ -84,6 +84,21 @@ public bool MoveToNextRow()
          return true;
       }
 
+      public IEnumerable<string> RetrieveExcelSheets(bool excludeEmptySheets)
```
msevestre, line 87: should that have a default of true (excludeEmptySheetS)

src/OSPSuite.Infrastructure.Import/Extensions/ExcelReadingExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+      /// Adds a list of values as a row to a DataTable
+      /// </summary>
+      /// <param name="rowValues">The values to be added to the DataTable</param>
+      public static void AddRowToDataTable(this DataTable dataTable, IReadOnlyList<string> rowValues)
+      {
+         var dataRow = dataTable.NewRow();
+         var i = 0; //IS THIS ZERO ACTUALLY RIGHT????
+         foreach (var cellContent in rowValues)
+         {
+            dataRow[i] = cellContent;
```
msevestre, line 19: what is this? lol . You are always overriding the same entry

src/OSPSuite.Infrastructure.Import/Services/IImportFromExcelTask.cs
```diff
@@ -0,0 +1,111 @@
...
+      public IReadOnlyList<string> RetrieveExcelSheets(string fileName, bool excludeEmptySheets = false)
+      {
+         try
+         {
+            var reader = new ExcelReader(fileName);
+            return reader.RetrieveExcelSheets(excludeEmptySheets).ToList();
+         }
+         catch
+         {
+            return Enumerable.Empty<string>().ToList();
```
msevestre, line 49: Array.empty

src/OSPSuite.Infrastructure.Import/Services/IImportFromExcelTask.cs
```diff
@@ -0,0 +1,111 @@
...
+      /// </summary>
+      /// <param name="filePath">The path to the excel file</param>
+      /// <param name="firstRowAsCaption">Whether or not the first row in the sheet should be interpreted as a caption</param>
+      /// <returns>The data tables corresponding to the file</returns>
+      IReadOnlyList<DataTable> GetAllDataTables(string filePath, bool firstRowAsCaption);
+   }
+
+   public class ImportFromExcelTask : IImportFromExcelTask
+   {
+      public IReadOnlyList<string> RetrieveExcelSheets(string fileName, bool excludeEmptySheets = false)
```
msevestre, line 40: why is default false? Does not true make more sense?

msevestre (changes requested):
all good except this AddRowToDataTable which looks very suspicious to me
And please run the formatted:)

src/OSPSuite.Infrastructure.Import/Core/DataSourceFileReaders/ExcelReader.cs
georgeDaskalakis: done. (it removed a few empty lines mainly)

georgeDaskalakis:
> all good except this AddRowToDataTable which looks very suspicious to me
> And please run the formatted:)

but you mean just the format here? The formatter did not change anything. But I do think I will rename the class to DataTableExtension, it makes more sense.

src/OSPSuite.Infrastructure.Import/Core/DataSourceFileReaders/ExcelReader.cs
georgeDaskalakis: I agree, it is better with a default value. Done.

msevestre:
no I mean the code in AddRowToDataTable  seems wrong. You have an index I that does not change in the loop. WTF?

src/OSPSuite.Infrastructure.Import/Extensions/ExcelReadingExtensions.cs
georgeDaskalakis: corrected

georgeDaskalakis:
Yeap, about that you are right, I just corrected it. But since it was also in a different comment I got confused about what you meant. I will also add a unit test for the reading - the existing unit tests that were in MoBi and I moved to Core are not testing if any of the values read are correct (possibly also did not need to since SmartXLS seemed to have a method to directly return a DataTable).

src/OSPSuite.Infrastructure.Import/Services/IImportFromExcelTask.cs
georgeDaskalakis: corrected

src/OSPSuite.Infrastructure.Import/Services/IImportFromExcelTask.cs
georgeDaskalakis: I have kept this exactly the way it was in MoBi. My understanding is that if you just want the list of names of all the sheets, it is the default that you get even the empty ones, since they exist as sheets.

msevestre:
@georgeDaskalakis Did you push? Still looks off to me

msevestre:
@georgeDaskalakis Can you push your changes when you have time. I can't merge the PR with some broken code sorry

georgeDaskalakis:
Sorry, I pushed just now because I needed to clarify one thing and finish the unit test. I ended up renaming the extension class to ImporterDataTableExtensions. There exists also a DataTableExtensions class in OSPSuite.Utility, which does not have this add List<string> as row functionality. Sill, since this is too specific, I would not like to extend the Utilities.

PR #1030, opened 2021-05-12
Add methods for getting Unit and Dimension of a quantity by path

PavelBal:
Sorry, I still cannot write tests (I simply don't know how it works in .NET).

This is insane, getting output values for my use case now takes 1 second as compared to 13 second through direct .net-objects access, and 34 seconds using R6 objects.

### DIFF
--- a/src/OSPSuite.R/Services/ContainerTask.cs
+++ b/src/OSPSuite.R/Services/ContainerTask.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Text.RegularExpressions;
 using OSPSuite.Assets;
 using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Formulas;
 using OSPSuite.Core.Domain.Services;
 using OSPSuite.Core.Extensions;
 using OSPSuite.Utility.Exceptions;
@@ -42,6 +43,26 @@ namespace OSPSuite.R.Services
       string[] AllMoleculesPathsIn(IModelCoreSimulation simulation);
       string[] AllParameterPathsIn(IModelCoreSimulation simulation);
       string[] AllStateVariableParameterPathsIn(IModelCoreSimulation simulation);
+
+      /// <summary>
+      ///    Returns names of base units of entities with given path (may contain wildcards)
+      /// </summary>
+      string[] BaseUnitNamesFromPath(IModelCoreSimulation simulation, string path);
+
+      /// <summary>
+      ///    Returns names of dimension of entities with given path (may contain wildcards)
+      /// </summary>
+      string[] DimensionNamesFromPath(IModelCoreSimulation simulation, string path);
+
+      /// <summary>
+      ///    Returns if the start values of entities with given path (may contain wildcards) are defined by an explicit formula
+      /// </summary>
+      bool[] IsExplicitFormulasFromPath(IModelCoreSimulation simulation, string path);
+
+      /// <summary>
+      ///    Adds quantities with given path (may contain wildcards) to output selections of the simulation.
+      /// </summary>
+      void AddQuantitiesToSimulationOutputFromPath(IModelCoreSimulation simulation, string path);
    }
 
    public class ContainerTask : IContainerTask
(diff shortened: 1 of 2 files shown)


msevestre:
I'll add the test and tag you

Any other methods like this??

PavelBal:
> Any other methods like this??

Yes, I added `IsExplicitFormulaFromPath` and will now check one last spot. I will let you know when I think everything is there.

src/OSPSuite.R/Services/ContainerTask.cs
```diff
@@ -115,6 +115,27 @@ public IParameter[] AllParametersForSensitivityAnalysisMatching(ISimulation simu
 
       public string[] AllStateVariableParameterPathsIn(IModelCoreSimulation simulation) => AllStateVariableParameterPathsIn(simulation?.Model?.Root);
 
+      public String[] BaseUnitNameFromPath(IModelCoreSimulation simulation, string path)
```
msevestre, line 118: WOW what is this method doing here.  Can you explain the user case?
AT any rate we need

1. Method to be defined in the interface (IContainerTask) WITH comments explaining what it does
2. Use string not String
3. Use plurals since it returns an array (BaseUnitNames)
4. You will have duplicates in this code (mg etc.. will be multiple time I guess depending on the path)
5. Or is path supposed to be a REAL path?
Anyways please explain use case and I can update

src/OSPSuite.R/Services/ContainerTask.cs
```diff
@@ -115,6 +115,27 @@ public IParameter[] AllParametersForSensitivityAnalysisMatching(ISimulation simu
...
+      {
+         var unitNames = new List<string>();
+         var allQuantities = AllQuantitiesMatching(simulation?.Model?.Root, path);
+         foreach (var quantity in allQuantities)
+         {
+            unitNames.Add(WithDimensionExtensions.BaseUnitName(quantity));
+         }
+         return unitNames.ToArray();
+      }
+      public String[] DimensionNameFromPath(IModelCoreSimulation simulation, string path)
```
msevestre, line 128: same comment for this baby

msevestre (changes requested):

src/OSPSuite.R/Services/ContainerTask.cs
PavelBal: I can address the points but probably not today.

In short - it will be used in R in getOutputValues

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/79bc93da81078f57584951fd4e98f0820cb691e4/R/utilities-simulation-results.R#L86

to get the unit and dimension without creating an object.

msevestre: but isn't `quantities` used for something else? The line you are pointing too is **not** creating a quantity

PavelBal: I have re-written the method that it does not create quantities at all. If quantities instead of paths are passed, paths are extracted and then proceeded with.

PavelBal:
Argh sorry, forgot to add the method definition in the interface.

Added two more "FromPath" methods. I do not have anything else to optimize for now, the steady-state use case does have less than 2 seconds overhead (total time 18 seconds, pure simulation time 17 seconds). Starting from 5 minutes yesterday!

msevestre:
@PavelBal your work is done here

msevestre:
@PavelBal I updated the code to be more .net idiomatic. But this is the same stuff otherwise. Adding tests now

src/OSPSuite.R/Services/ContainerTask.cs
```diff
@@ -115,6 +136,18 @@ public IParameter[] AllParametersForSensitivityAnalysisMatching(ISimulation simu
 
       public string[] AllStateVariableParameterPathsIn(IModelCoreSimulation simulation) => AllStateVariableParameterPathsIn(simulation?.Model?.Root);
 
+      public string[] BaseUnitNamesFromPath(IModelCoreSimulation simulation, string path) =>
```
msevestre, line 139: @PavelBal slightly shorted all of this :)

src/OSPSuite.R/Services/ContainerTask.cs
```diff
@@ -115,6 +136,18 @@ public IParameter[] AllParametersForSensitivityAnalysisMatching(ISimulation simu
...
+         AllQuantitiesMatching(simulation, path).Select(x => x.BaseUnitName()).ToArray();
+
+      public string[] DimensionNamesFromPath(IModelCoreSimulation simulation, string path) => 
+         AllQuantitiesMatching(simulation, path).Select(x => x.DimensionName()).ToArray();
+
+      public bool[] IsExplicitFormulasFromPath(IModelCoreSimulation simulation, string path) =>
```
msevestre, line 145: I am wondering how this will be used to be honest. 
Also I renamed add added a `s` to be plural again

msevestre:
added some tests too

tests/OSPSuite.R.Tests/Services/ContainerTaskSpecs.cs
```diff
@@ -305,4 +305,37 @@ public void should_only_return_parameters_with_a_RHS_formula_defined()
...
+      }
+   }
+
+
+   public class When_retrieving_the_base_unit_names_for_a_given_path : concern_for_ContainerTask
+   {
+      [Observation]
+      public void should_return_the_expected_entries()
+      {
+         sut.BaseUnitNamesFromPath(_simulation, pathFrom(_liver.Name, INTRACELLULAR, $"Vol{Constants.WILD_CARD}")).ShouldOnlyContain(_volumeLiverCell.Dimension.BaseUnit.Name);
```
msevestre, line 329: @PavelBal We have no fake stuff here.
in this case, you call you method and you verify that it does what you expect. Very simple

PR #1034, opened 2021-05-17
fixing padding and renaming UnitsEditorCaption

abdelr:


src/OSPSuite.UI/Views/Importer/MappingParameterEditorView.Designer.cs
```diff
@@ -142,7 +142,7 @@ private void InitializeComponent()
          // 
          this.unitsLayoutControlItem.Control = this.unitsPanelControl;
          this.unitsLayoutControlItem.Location = new System.Drawing.Point(0, 0);
-         this.unitsLayoutControlItem.Name = "Units";
+         this.unitsLayoutControlItem.Name = Captions.UnitsEditorCaption;
```
msevestre, line 145: THis should be done in InitializeResouces method in the view. We never change the Designer file by hand

msevestre (changes requested):

PR #1037, opened 2021-05-17
1017 file format not supported

georgeDaskalakis:
I am marking this as a draft because I would first like to discuss two points:

a) The way this is solved changes the dialogCreator.MessageBoxError() function. Alternatively, if you find it less invasive, I could write an extra function that does exactly this ( MessageBoxErrorWithLink() ) and avoid the optional parameter.  
The way the arguments are written right now nothing should be changing nothing for every other messagebox - although icon and Button come directly from windows forms and system icons they are essentially the same. The only difference is html mode is always turned on now, but if you think this is a problem it is easy to change.

b) I would say the ShowErrorMessage() function in the ImporterView.cs is redundant and also causes some code duplication. I would suggest the presenter just calls the dialogCreator messagebox directly instead.

### DIFF
--- a/src/OSPSuite.Core/Services/IDialogCreator.cs
+++ b/src/OSPSuite.Core/Services/IDialogCreator.cs
@@ -12,7 +12,7 @@ namespace OSPSuite.Core.Services
 
    public interface IDialogCreator
    {
-      void MessageBoxError(string message);
+      void MessageBoxError(string message, bool hyperlink = false);
 
       ViewResult MessageBoxYesNoCancel(string message, ViewResult defaultButton = ViewResult.Yes);
 
--- a/src/OSPSuite.Core/Services/IDialogCreator.cs
+++ b/src/OSPSuite.Core/Services/IDialogCreator.cs
@@ -10,9 +10,16 @@ namespace OSPSuite.Core.Services
       No
    }
 
+   public enum ViewIcon
+   {
+      Information,
+      Error,
+      Question
+   }
+
    public interface IDialogCreator
    {
-      void MessageBoxError(string message, bool hyperlink = false);
+      void MessageBoxError(string message);
 
       ViewResult MessageBoxYesNoCancel(string message, ViewResult defaultButton = ViewResult.Yes);
 
--- a/src/OSPSuite.Core/Services/IDialogCreator.cs
+++ b/src/OSPSuite.Core/Services/IDialogCreator.cs
@@ -10,13 +10,6 @@ namespace OSPSuite.Core.Services
       No
    }
 
-   public enum ViewIcon
-   {
-      Information,
-      Error,
-      Question
-   }
-
    public interface IDialogCreator
    {
       void MessageBoxError(string message);
--- a/src/OSPSuite.UI/Services/DataImporter.cs
+++ b/src/OSPSuite.UI/Services/DataImporter.cs
@@ -109,7 +109,7 @@ namespace OSPSuite.UI.Services
             }
             catch (Exception e) when (e is UnsupportedFormatException || e is UnsupportedFileTypeException)
             {
-               _dialogCreator.MessageBoxError(e.Message);
+               _dialogCreator.MessageBoxError(e.Message, true);
                return (new List<DataRepository>(), null);
             }
 
@@ -153,7 +153,7 @@ namespace OSPSuite.UI.Services
          }
          catch (Exception e) when (e is UnsupportedFormatException || e is UnsupportedFileTypeException)
          {
-            _dialogCreator.MessageBoxError(e.Message);
+            _dialogCreator.MessageBoxError(e.Message, true);
             return new List<DataRepository>();
          }
 
--- a/src/OSPSuite.UI/Services/DataImporter.cs
+++ b/src/OSPSuite.UI/Services/DataImporter.cs
@@ -109,7 +109,7 @@ namespace OSPSuite.UI.Services
             }
             catch (Exception e) when (e is UnsupportedFormatException || e is UnsupportedFileTypeException)
             {
-               _dialogCreator.MessageBoxError(e.Message, true);
+               _dialogCreator.MessageBoxError(e.Message);
                return (new List<DataRepository>(), null);
             }
 
@@ -153,7 +153,7 @@ namespace OSPSuite.UI.Services
          }
          catch (Exception e) when (e is UnsupportedFormatException || e is UnsupportedFileTypeException)
          {
-            _dialogCreator.MessageBoxError(e.Message, true);
+            _dialogCreator.MessageBoxError(e.Message);
             return new List<DataRepository>();
          }
 
--- a/src/OSPSuite.UI/Services/DialogCreator.cs
+++ b/src/OSPSuite.UI/Services/DialogCreator.cs
@@ -30,12 +30,12 @@ namespace OSPSuite.UI.Services
 
       public void MessageBoxError(string message)
       {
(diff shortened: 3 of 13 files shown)


src/OSPSuite.UI/Views/Importer/ImporterView.cs
```diff
@@ -29,7 +29,18 @@ public ImporterView()
 
       public void ShowErrorMessage(string message)
       {
-         XtraMessageBox.Show(this, message, "", MessageBoxButtons.OK, MessageBoxIcon.Error);
+         var args = new XtraMessageBoxArgs
```
msevestre, line 32: Why is there a call to the user in the view? THis is not ok and was missed before. The presenter is the one that needs to manager usser interaction. Never the view

src/OSPSuite.UI/Services/DialogCreator.cs
```diff
@@ -25,9 +27,19 @@ public DialogCreator(IDialogResultToViewResultMapper mapper, DirectoryMapSetting
          _applicationConfiguration = applicationConfiguration;
       }
 
-      public void MessageBoxError(string message)
+      public void MessageBoxError(string message, bool hyperlink)
       {
-         XtraMessageBox.Show(message, _applicationConfiguration.ProductNameWithTrademark, MessageBoxButtons.OK, MessageBoxIcon.Error);
+         var args = new XtraMessageBoxArgs { Caption = _applicationConfiguration.ProductNameWithTrademark, Text = message, Buttons = new DialogResult[] {
+            DialogResult.OK } ,
+            Icon = DevExpress.Utils.Drawing.Helpers.StockIconHelper.GetWindows8AssociatedIcon(SystemIcons.Error), AllowHtmlText = DefaultBoolean.True };
```
msevestre, line 34: What is this SystemIcons.Error? What aren't you using the SuystemsIcon.Errror
Also why do you think this feature should only be available for MessageBoxError? What about Info? Or yes no etc..

src/OSPSuite.UI/Services/DataImporter.cs
```diff
@@ -109,7 +109,7 @@ DataImporterSettings dataImporterSettings
             }
             catch (Exception e) when (e is UnsupportedFormatException || e is UnsupportedFileTypeException)
             {
-               _dialogCreator.MessageBoxError(e.Message);
+               _dialogCreator.MessageBoxError(e.Message, true);
```
msevestre, line 112: This is a very leaky abstracvtion. Here you need to know know that the message as hyperlink. Why do we need a parameter to specify use HTML or not? Cannot we always use it? In my opinion, this looks like some internal stuff. that should not be visible to the caller

src/OSPSuite.Core/Services/IDialogCreator.cs
```diff
@@ -12,7 +12,7 @@ public enum ViewResult
 
    public interface IDialogCreator
    {
-      void MessageBoxError(string message);
+      void MessageBoxError(string message, bool hyperlink = false);
```
msevestre, line 15: if you need in the code to differentiate usage of hyperlink or not, I would search the message for href  or sthg like this?

src/OSPSuite.UI/Views/Importer/ImporterView.cs
```diff
@@ -29,7 +29,18 @@ public ImporterView()
...
+         {
+            Caption = "",
```
msevestre, line 34: Why is the caption empty? This should never be the case for a message, especially an error.
At any rate, this code should go

msevestre (changes requested):

src/OSPSuite.UI/Views/Importer/ImporterView.cs
georgeDaskalakis: yeap, I also agree, is what I was saying in the description of the PR. I will change this accordingly then.

src/OSPSuite.UI/Views/Importer/ImporterView.cs
georgeDaskalakis: Was like this before and just left it there, since I also agree it should go

src/OSPSuite.UI/Services/DialogCreator.cs
georgeDaskalakis: You mean use MessageBoxIcon.Error instead right? Because they are of different types. The show() method actually belongs to the base class MessageBox, not XtaMessagebox, hence the XtraMessageBoxArgs don't have it as an argument.  I used the suggested from devexpress: [](https://supportcenter.devexpress.com/ticket/details/t524421/can-i-use-messageboxicon-with-xtramessageboxargs) . 

About the rest of the MessageBoxes: of course if you also agree on actually expanding them (but then it really will have to be like this, with SystemIcon.Error and so on), I will implement it for the rest of them too, otherwise I will just write a new function for an error message box with the link, as suggested in the description of the PR.

src/OSPSuite.Core/Services/IDialogCreator.cs
georgeDaskalakis: I like the suggestion, U had not thought of that. I will adjust it accordingly then.

src/OSPSuite.UI/Views/Importer/ImporterView.cs
georgeDaskalakis: done

src/OSPSuite.UI/Services/DataImporter.cs
georgeDaskalakis: done

src/OSPSuite.UI/Services/DialogCreator.cs
georgeDaskalakis: encapsulated this into a function

msevestre (commented):
Looking good. LEt me know what you think about my questions and I can merge

src/OSPSuite.UI/Services/DialogCreator.cs
```diff
@@ -47,40 +79,57 @@ public ViewResult MessageBoxYesNo(string message, ViewResult defaultButton = Vie
...
-            return MessageBoxDefaultButton.Button1;
-
-         if (defaultButton == ViewResult.No)
-            return MessageBoxDefaultButton.Button2;
-
-         if (defaultButton == ViewResult.Cancel)
-            return MessageBoxDefaultButton.Button3;
+         switch (icon)
+         {
+            case ViewIcon.Error:
```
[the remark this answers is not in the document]
georgeDaskalakis: fixed

src/OSPSuite.UI/Services/DialogCreator.cs
```diff
@@ -27,7 +30,35 @@ public DialogCreator(IDialogResultToViewResultMapper mapper, DirectoryMapSetting
...
+            Caption = _applicationConfiguration.ProductNameWithTrademark,
+            Text = message,
+            Buttons = buttons,
+            Icon = icon,
+            DefaultButtonIndex = defaultButtonIndex,
+            AllowHtmlText = DefaultBoolean.True
+         };
+
+         if (containsHyperlink(message))
+            args.HyperlinkClick += delegate(object sender, HyperlinkClickEventArgs e) { System.Diagnostics.Process.Start(e.Link); };
```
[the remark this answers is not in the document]
georgeDaskalakis: done

georgeDaskalakis:
@msevestre ready to merge I would say :)

msevestre (commented):
awesome!

PR #1039, opened 2021-05-17
fixes #1018

### LINKED ISSUE #1018: Default Error type is Arithmetic when no unit is availalbe in the column. It should be geeometric
Loading the file attached should default to Geometric error and not arithmetic

image
[ObservedData.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/6454857/ObservedData.xlsx)

georgeDaskalakis:
closes #1018

### DIFF
--- a/src/OSPSuite.Core/Import/DataFormatParameter.cs
+++ b/src/OSPSuite.Core/Import/DataFormatParameter.cs
@@ -134,13 +134,10 @@ namespace OSPSuite.Core.Import
             return Captions.Importer.MappingHintGeometricError;
 
          if (!MappedColumn.Unit.ColumnName.IsNullOrEmpty())
-         {
             return Captions.Importer.MappingHintUnitColumn(ColumnName, MappedColumn.Name, MappedColumn.Unit.ColumnName);
-         }
-
+         
          if (!MappedColumn.Unit.SelectedUnit.IsNullOrEmpty())
             return Captions.Importer.MappingHint(ColumnName, MappedColumn.Name, MappedColumn.Unit.SelectedUnit);
-
 
          return Captions.Importer.MappingHintNoUnit(ColumnName, MappedColumn.Name);
       }
--- a/src/OSPSuite.Core/Import/DataFormatParameter.cs
+++ b/src/OSPSuite.Core/Import/DataFormatParameter.cs
@@ -1,4 +1,6 @@
 ﻿using OSPSuite.Assets;
+using OSPSuite.Core.Domain;
+using OSPSuite.Utility.Extensions;
 
 namespace OSPSuite.Core.Import
 {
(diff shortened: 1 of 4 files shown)


src/OSPSuite.Core/Import/DataFormatParameter.cs
```diff
@@ -128,8 +130,19 @@ public override bool EquivalentTo(DataFormatParameter other)
 
       public override string TooltipDescription()
       {
-         
-         return Captions.Importer.MappingHint(ColumnName, MappedColumn.Name, MappedColumn.Unit.SelectedUnit);
+         if (MappedColumn.ErrorStdDev == Constants.STD_DEV_GEOMETRIC)
+            return Captions.Importer.MappingHintGeometricError;
+
+         if (!MappedColumn.Unit.ColumnName.IsNullOrEmpty())
+         {
```
msevestre, line 137: Remoe the { } if before and after you don't use them.  I don't care if we use {} or not , but the code within the same function needs to be consistent

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -90,7 +90,10 @@ protected virtual void ExtractQualifiedHeadings(List<string> keys, List<string>
                };
                if (columnInfos.First(ci => ci.DisplayName == header).IsAuxiliary())
                {
-                  col.ErrorStdDev = Constants.STD_DEV_ARITHMETIC;
+                  if(units.ColumnName == "" && units.SelectedUnit == UnitDescription.InvalidUnit)
```
msevestre, line 93: string.isNullOrEmpty.

src/OSPSuite.UI/Views/Importer/ColumnMappingView.cs
```diff
@@ -344,15 +346,24 @@ public string Format(DataFormatParameter model)
 
          if (model is MappingDataFormatParameter mapping)
          {
-            var str = $"Units: {mapping.MappedColumn.Unit.SelectedUnit}";
+            var str = "";
+
+            if (!mapping.MappedColumn.Unit.ColumnName.IsNullOrEmpty())
+            {
```
msevestre, line 352: same stuff. Ensure the same styling in the same function. Otherwise the code is not cohesive

src/OSPSuite.UI/Views/Importer/ColumnMappingView.cs
```diff
@@ -344,15 +346,24 @@ public string Format(DataFormatParameter model)
 
          if (model is MappingDataFormatParameter mapping)
          {
-            var str = $"Units: {mapping.MappedColumn.Unit.SelectedUnit}";
+            var str = "";
```
msevestre, line 349: Suggestion. This allocates the old and new string all the time
I would recommend creating a list, adding error to the list and then using toString() `extension` method to create one string from an Ienumerable. 
In general += on string can be easily avoided

msevestre (changes requested):

src/OSPSuite.Core/Import/DataFormatParameter.cs
georgeDaskalakis: corrected

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
georgeDaskalakis: corrected

src/OSPSuite.UI/Views/Importer/ColumnMappingView.cs
georgeDaskalakis: done

src/OSPSuite.UI/Views/Importer/ColumnMappingView.cs
georgeDaskalakis: Yeap I like this suggestion a lot. I changed this. Also changed class signature to internal.

georgeDaskalakis:
So this works now. Weirdly returning an empty string was breaking the contents of the grid (screenshot below). My guess is this happens because of some check somewhere in the AutoBind for the Grid. 

image

I am using the existing Caption EmptyColumn. I thought this would be better thar returning " " in this case, or creating a new Caption.Importer.EmptyColumn. Would you agree @msevestre ?

msevestre:
looks good to me

PR #1040, opened 2021-05-18
fixes #1035

### LINKED ISSUE #1035: "Reload under same settings" - crash if the sheet is not available
I import the data from [Book 3.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/6493279/Book.3.xlsx)
Then I select "Reload under same settings" and select [Book 3 - Kopie SheetRenamed.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/6493293/Book.3.-.Kopie.SheetRenamed.xlsx) (which is the copy of the original file, but the (only) sheet was renamed in the excel file). Then it crashes with the error message below.
As long as https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1005 is not implemented: it would be good to get an "explicit" error message ("Excel sheet 'Sheet 1' could not be found in ...") instead of the "oops" error message
grafik

georgeDaskalakis:
closes #1035

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -425,6 +425,15 @@ namespace OSPSuite.Assets
             return sb.ToString();
          }
 
+         public static string SheetsNotFound(List<string> listOfSheetNames)
+         {
+            var sb = new StringBuilder();
+            sb.AppendLine("The following excel sheets were not found in the file and will not be imported ");
+            sb.AppendLine();
+            listOfSheetNames.ForEach(item => sb.Append(item + "\n"));
+            return sb.ToString();
+         }
+
          public static readonly string UseFiltersForImport = "Use the filters for importing the data";
          
          public static readonly string AddGroupByTitle = "Add group by";
--- a/src/OSPSuite.UI/Services/DataImporter.cs
+++ b/src/OSPSuite.UI/Services/DataImporter.cs
@@ -180,10 +180,19 @@ namespace OSPSuite.UI.Services
          dataSource.SetDataFormat(dataSourceFile.Format);
          dataSource.SetNamingConvention(configuration.NamingConventions);
          var sheets = new Cache<string, DataSheet>();
+         var missingSheets = new List<string>();
          foreach (var key in configuration.LoadedSheets)
          {
+            if (!dataSourceFile.DataSheets.Contains(key))
+            {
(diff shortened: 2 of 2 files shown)


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -425,6 +425,15 @@ public static string ConfirmDroppingExcelColumns(string listOfExcelColumns)
             return sb.ToString();
          }
 
+         public static string SheetsNotFound(List<string> listOfSheetNames)
+         {
+            var sb = new StringBuilder();
+            sb.AppendLine("The following excel sheets were not found in the file and will not be imported ");
+            sb.AppendLine();
+            listOfSheetNames.ForEach(item => sb.Append(item + "\n"));
```
msevestre, line 433: any reason not to use ?AppendLine. Why are you adding the \n by hand?

src/OSPSuite.UI/Services/DataImporter.cs
```diff
@@ -180,11 +180,20 @@ DataImporterSettings dataImporterSettings
          dataSource.SetDataFormat(dataSourceFile.Format);
          dataSource.SetNamingConvention(configuration.NamingConventions);
          var sheets = new Cache<string, DataSheet>();
+         var missingSheets = new List<string>();
          foreach (var key in configuration.LoadedSheets)
          {
+            if (!dataSourceFile.DataSheets.Contains(key))
```
msevestre, line 186: Can we add some tests for this? Or is this too complicated at the moment?

msevestre (changes requested):

src/OSPSuite.Assets/UIConstants.cs
georgeDaskalakis: corrected

src/OSPSuite.UI/Services/DataImporter.cs
georgeDaskalakis: This is a good question. Adding a test here would mean testing the whole ImportFromConfiguration() function of the service. To test just this specific missing sheet we would then have to mock many things. I would suggest this becomes part of a new integration test, where excel files and conifgurations are being tested for import. 

Adding integration tests was in our ToDo list anyway, we could start with that as soon as the open PRs are finished.

PR #1042, opened 2021-05-18
Fixes #1041 disabled color

### LINKED ISSUE #1041: Ensure that app disabled color is used for ValueOrigin control

msevestre:


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -2307,7 +2307,7 @@ public static class Colors
       /// </summary>
       public static Color ChartDiagramBack = Color.White;
 
-      public static readonly Color Disabled = Color.LightGray;
+      public static Color Disabled = Color.FromArgb(255, 247, 247, 249);
```
msevestre, line 2310: This is our new default disabled in PKSim. Let's use it as default here too

src/OSPSuite.UI/Controls/UxGridView.cs
```diff
@@ -37,7 +37,7 @@ public class UxGridView : GridView
       /// <summary>
       ///    Color used for cell that are locked/disabled (End of gradient)
       /// </summary>
-      protected Color _colorDisabled = Color.LightGray;
+      protected Color _colorDisabled = Colors.Disabled;
```
msevestre, line 40: This was not using the right constnat lol

PR #1049, opened 2021-05-21
Fix and only possible to manually edit molecule

abdelr:


abdelr:
@msevestre @georgeDaskalakis @Yuri05 I would say, this is the last important bug fix. After merging we should update PKSim and Core to use the last Core and distribute to the testers. Only two questions remain from my side.
1. I added "Edit manually" to the text on extra fields. I find the suggestion from @stephan better but I was not able to implement it. The proposal was to visually disable all the cells that are not editable. Now, this is not always the style we use. If you find adding "Edit manually" enough I am cool with it. Otherwise we could always change it before V10 is released. This should be a UI change only.
2. We need to test this logic. I think this is a good topic for integration tests since it is in the interaction of different components. Maybe a good point for us to take a look on how to write integration tests on the UI but as you find it better.

Yuri05 (approved):

PR #1057, opened 2021-05-26
1056 unit not updating fixed

### LINKED ISSUE #1056: unit does not get updated after change
In the importer the unit cannot be set or validated anymore

georgeDaskalakis:
fixes #1056

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
@@ -24,7 +24,6 @@ namespace OSPSuite.Presentation.Presenters.Importer
       private MappingProblem _mappingProblem = new MappingProblem() { MissingMapping = new List<string>(), MissingUnit = new List<string>() };
       private readonly IMappingParameterEditorPresenter _mappingParameterEditorPresenter;
       private readonly IMetaDataParameterEditorPresenter _metaDataParameterEditorPresenter;
-      private ColumnMappingDTO _currentModel;
       public ColumnMappingPresenter
       (
          IColumnMappingView view,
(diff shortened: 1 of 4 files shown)


src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -145,44 +144,45 @@ public void SetDescriptionForRow(ColumnMappingDTO model)
...
-
-         var metaData = _currentModel.Source as MetaDataFormatParameter;
-         if (metaData == null) return;
+         if (mappingSource == null)
+            return;
 
-         _mappings.First(m => m.MappingName == metaData.MetaDataId).ExcelColumn = _metaDataParameterEditorPresenter.Input;
-         metaData.ColumnName = _metaDataParameterEditorPresenter.Input;
-         metaData.IsColumn = false;
+         _mappings.First(m => m.MappingName == mappingSource.MetaDataId).ExcelColumn = _metaDataParameterEditorPresenter.Input;
```
msevestre, line 152: First will throw an exception if not found...Or it cannot happen in this scenario?

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -145,44 +144,45 @@ public void SetDescriptionForRow(ColumnMappingDTO model)
...
+         var model = _mappings.FirstOrDefault(x => x.Source == mappingSource);
+
+         if (model == null) 
+            return;
+
+         var column = ((MappingDataFormatParameter)model.Source).MappedColumn;
          column.Unit = _mappingParameterEditorPresenter.Unit;
          if (!string.IsNullOrEmpty(column.Unit.ColumnName))
          {
             column.Unit = new UnitDescription(_rawData.GetColumn(column.Unit.ColumnName).First(u => !string.IsNullOrEmpty(u)), column.Unit.ColumnName);
```
msevestre, line 174: same question here with the First (I know it's not from this code change but since I see it :))

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -211,13 +211,7 @@ public void UpdateDescriptrionForModel()
...
       {
          var metaDataCategory = _metaDataCategories.FindByName(model.MappingName);
-         if (metaDataCategory != null && metaDataCategory.AllowsManualInput)
-         {
-            _currentModel = model;
-            return true;
-         }
-
-         return false;
+         return metaDataCategory != null && metaDataCategory.AllowsManualInput;
```
msevestre, line 214: much better. No side effect

msevestre (commented):
a question regarding First

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
georgeDaskalakis: Well in that case it would mean that we changed the unit for a mapping that doesn't exist - which would really be a software bug. I would let the exception be raised i this quite improbable case.

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
georgeDaskalakis: same answer here: not found would mean we mapped a column that did not exist, I suggest we keeo the same behaviour as above.

msevestre:
ok merging then. I think next time, it would be good to add comments explaining why we want to use First.

PR #1060, opened 2021-05-26
Fixes #1059 rename

### LINKED ISSUE #1059: Refresh Problem when renaming BBs or Simulations
@Yuri05 commented on [Tue May 25 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1768)

Happens both in V9 and V10.
When I rename a building block or a simulation and give it a long name: it's truncated in the BB-/Sim-Explorer
 

The long name is only displayed properly after I reopen the project or add another building block/simulation

grafik

---

@msevestre commented on [Tue May 25 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1768#issuecomment-847885798)

Yep. Nothing we can do here without a lot of time to be spent. It is because we have turned on the scrolling horizontal a while back. If you add a new node with a long name, I think it will refresh but renaming nodes does not. 

---

@Yuri05 commented on [Tue May 25 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1768#issuecomment-847989401)

Add a dummy node and remove it immediately every time when a node is renamed?

---

@msevestre commented on [Wed May 26 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1768#issuecomment-849077177)

that works!

msevestre:


src/OSPSuite.Presentation/Presenters/AbstractExplorerPresenter.cs
```diff
@@ -380,5 +382,13 @@ public virtual void Handle(ProjectClosedEvent eventToHandle)
       {
          _view.DestroyNodes();
       }
+
+      protected void RefreshTreeAfterRename()
```
msevestre, line 386: @Yuri05 I am creating this method in core so that it can be used in MoBi and PKSim. Credit to you:)

PR #1062, opened 2021-05-27
null check

### LINKED ISSUE #1061: Importer crashes when loading a dataset without error column
1. Select WorkShopData.xlsx
2. Load first sheet

This results on an unhandled exception when trying to access a member of a null object.

abdelr:
Fixes #1061

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
@@ -153,7 +153,7 @@ namespace OSPSuite.Infrastructure.Import.Core
                      var measurementColumn = set.Data.FirstOrDefault(x => x.Key.ColumnInfo.Name == column.Name);
                      var errorColumn = set.Data.FirstOrDefault(x => x.Key.ColumnInfo.Name == relatedColumn.Name);
 
-                     if (measurementColumn.Value.Count != errorColumn.Value.Count)
+                     if (errorColumn.Value != null && measurementColumn.Value.Count != errorColumn.Value.Count)
                         throw new OSPSuiteException(Error.MismatchingArrayLengths);
 
                      if (errorColumn.Key == null)


abdelr:
Unit.HasSynonym fails to detect "µmol/L" => "µmol/l". Is this intended?

msevestre:
@abdelr I would have assume it would work. Can you create another issue for this?

msevestre:
Just checked the code. l L is not implemented as default synonym indeed

msevestre:
but I don't think you are using the right method

```
 /// <summary>
      ///    Returns the unit defined with the given name (or synonym) or null if not found
      /// </summary>
      /// <param name="unitName">Name of unit</param>
      /// <param name="ignoreCase">Should ignore case when matching units. Default is <c>false</c></param>
      Unit FindUnit(string unitName, bool ignoreCase = false);

      /// <summary>
      ///    Returns the true if a unt with the given name (or synonym) is found otherwise false
      /// </summary>
      /// <param name="unitName">Name of unit</param>
      /// <param name="ignoreCase">Should ignore case when matching units. Default is <c>false</c></param>
      bool SupportsUnit(string unitName, bool ignoreCase = false);

```

msevestre:
or 
```
  public IDimension DimensionForUnit(string unitName)
      {
         var matches = Dimensions.Where(x => x.SupportsUnit(unitName, ignoreCase: true)).ToList();
         if (!matches.Any())
            return null;

         //Try to find the first one that matches EXACTLY 
         return matches.FirstOrDefault(x => x.SupportsUnit(unitName, ignoreCase: false)) ?? matches.First();
      }
```

msevestre:
I think ` Unit.HasSynonym` is too low level @abdelr where is this being useD?

msevestre:
@abdelr If you can let me know where you are using it (maybe indirectly) I can see if another method should be used. I can't seem to find it

PR #1071, opened 2021-05-31
Fixes #1069

### LINKED ISSUE #1069: Add find unit method in DImension Factory

msevestre:


src/OSPSuite.Core/Domain/UnitSystem/DimensionFactory.cs
```diff
@@ -119,6 +119,12 @@ public IDimension DimensionForUnit(string unitName)
          return matches.FirstOrDefault(x => x.SupportsUnit(unitName, ignoreCase: false)) ?? matches.First();
       }
 
+      public (IDimension dimension, Unit unit) FindUnit(string unitName)
```
msevestre, line 122: @georgeDaskalakis I think it makes sense to also return the dimension otherwise there is not much you can do with the unit alone. Let me know your thoughts

Yuri05 (approved):

PR #1072, opened 2021-05-31
Fixes #1070 add silent obs data

### LINKED ISSUE #1070: Allow silent for ObservedData Delete
@georgeDaskalakis I'll take care of this right now.

msevestre:


src/OSPSuite.Core/Domain/Services/ObservedDataTask.cs
```diff
@@ -65,16 +65,13 @@ protected ObservedDataTask(IDialogCreator dialogCreator, IOSPSuiteExecutionConte
          _objectTypeResolver = objectTypeResolver;
       }
 
-      public bool Delete(IEnumerable<DataRepository> observedDataToBeRemoved)
+      public bool Delete(IEnumerable<DataRepository> observedDataToBeRemoved, bool silent = false)
```
msevestre, line 68: @Yuri05 and @georgeDaskalakis 
I just added this flag here for the method with IEnumerable only. Why? Because calling the method with only ONE entry is typically something that is not done in a batch mode. (at least I think it should not). We can add it to the other method as well but for now, I don't think it's required

Yuri05 (approved):

PR #1075, opened 2021-06-01
initial fix

### LINKED ISSUE #1065: Cannot import Ontogeny file (Ranorex test data)
Trying to run the ranorex tests for PK-Sim. Importing Fraction Data does not work as expected

image
[OntogenyImport.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/6565166/OntogenyImport.xlsx)

Even if trying by hand, I cannot select the empty unit!
image

georgeDaskalakis:
fixes #1065

### DIFF
--- a/src/OSPSuite.Core/Serialization/SerializationTransaction.cs
+++ b/src/OSPSuite.Core/Serialization/SerializationTransaction.cs
@@ -25,7 +25,7 @@ namespace OSPSuite.Core.Serialization
          ICloneManagerForModel cloneManagerForModel = null, 
          IEnumerable<DataRepository> dataRepositories = null)
       {
-         return new SerializationContext(dimensionFactory ?? container.Resolve<IDimensionFactory>(), objectBaseFactory, withIdRepository, dataRepositories, cloneManagerForModel, container);
+         return new SerializationContext(dimensionFactory, objectBaseFactory, withIdRepository, dataRepositories, cloneManagerForModel, container);
       }
    }
 }
--- a/src/OSPSuite.Core/Serialization/SerializationTransaction.cs
+++ b/src/OSPSuite.Core/Serialization/SerializationTransaction.cs
@@ -25,7 +25,7 @@ namespace OSPSuite.Core.Serialization
          ICloneManagerForModel cloneManagerForModel = null, 
          IEnumerable<DataRepository> dataRepositories = null)
       {
-         return new SerializationContext(dimensionFactory, objectBaseFactory, withIdRepository, dataRepositories, cloneManagerForModel, container);
+         return new SerializationContext(dimensionFactory ?? container.Resolve<IDimensionFactory>(), objectBaseFactory, withIdRepository, dataRepositories, cloneManagerForModel, container);
       }
    }
 }
--- a/src/OSPSuite.Core/Serialization/Xml/ColumnXmlSerializer.cs
+++ b/src/OSPSuite.Core/Serialization/Xml/ColumnXmlSerializer.cs
@@ -1,7 +1,5 @@
 ﻿using System.Xml.Linq;
-using OSPSuite.Core.Domain;
 using OSPSuite.Core.Import;
-using OSPSuite.Utility.Extensions;
 
 namespace OSPSuite.Core.Serialization.Xml
 {
@@ -14,22 +12,6 @@ namespace OSPSuite.Core.Serialization.Xml
          Map(x => x.Name);
          Map(x => x.Unit);
          Map(x => x.Dimension);
-      }
-
-      protected override void TypedDeserialize(Column objectToDeserialize, XElement outputToDeserialize, SerializationContext context)
-      {
-         base.TypedDeserialize(objectToDeserialize, outputToDeserialize, context);
-
-         if (objectToDeserialize.ErrorStdDev == Constants.STD_DEV_GEOMETRIC)
-         {
-            objectToDeserialize.Unit = new UnitDescription();
-            objectToDeserialize.Dimension = Constants.Dimension.NO_DIMENSION;
-         }
-
-         if (!objectToDeserialize.Unit.ColumnName.IsNullOrEmpty())
-         {
-            objectToDeserialize.Dimension = null;
-         }
       }
    }
 }
--- a/src/OSPSuite.Core/Serialization/Xml/ColumnXmlSerializer.cs
+++ b/src/OSPSuite.Core/Serialization/Xml/ColumnXmlSerializer.cs
@@ -1,6 +1,7 @@
 ﻿using System.Xml.Linq;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Import;
+using OSPSuite.Utility.Extensions;
 
 namespace OSPSuite.Core.Serialization.Xml
 {
@@ -18,9 +19,17 @@ namespace OSPSuite.Core.Serialization.Xml
       protected override void TypedDeserialize(Column objectToDeserialize, XElement outputToDeserialize, SerializationContext context)
       {
          base.TypedDeserialize(objectToDeserialize, outputToDeserialize, context);
-         
+
          if (objectToDeserialize.ErrorStdDev == Constants.STD_DEV_GEOMETRIC)
+         {
             objectToDeserialize.Unit = new UnitDescription();
+            objectToDeserialize.Dimension = Constants.Dimension.NO_DIMENSION;
+         }
+
+         if (!objectToDeserialize.Unit.ColumnName.IsNullOrEmpty())
+         {
+            objectToDeserialize.Dimension = null;
+         }
       }
    }
 }
--- a/src/OSPSuite.Core/Serialization/Xml/ColumnXmlSerializer.cs
+++ b/src/OSPSuite.Core/Serialization/Xml/ColumnXmlSerializer.cs
@@ -12,6 +12,7 @@ namespace OSPSuite.Core.Serialization.Xml
          Map(x => x.LloqColumn);
          Map(x => x.Name);
          Map(x => x.Unit);
+         Map(x => x.Dimension);
       }
 
       protected override void TypedDeserialize(Column objectToDeserialize, XElement outputToDeserialize, SerializationContext context)
--- a/src/OSPSuite.Infrastructure.Import/Core/DataFormat/MixColumnsDataFormat.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataFormat/MixColumnsDataFormat.cs
@@ -53,7 +53,7 @@ namespace OSPSuite.Infrastructure.Import.Core.DataFormat
          var unitKey = data.GetHeaders().FindHeader(description + "_UNIT");
          if (unitKey == null)
          {
-            return new UnitDescription();
+            return new UnitDescription(""); //in this case return the default unit for Dimensionless
          }
          keys.Remove(unitKey);
          rank++;
--- a/src/OSPSuite.Infrastructure.Import/Core/DataFormat/MixColumnsDataFormat.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataFormat/MixColumnsDataFormat.cs
@@ -53,7 +53,7 @@ namespace OSPSuite.Infrastructure.Import.Core.DataFormat
          var unitKey = data.GetHeaders().FindHeader(description + "_UNIT");
          if (unitKey == null)
          {
-            return new UnitDescription(""); //in this case return the default unit for Dimensionless
+            return new UnitDescription();
          }
          keys.Remove(unitKey);
          rank++;
--- a/src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
@@ -100,8 +100,8 @@ namespace OSPSuite.Infrastructure.Import.Core.Mappers
             var adjustedValue = truncateUsingLLOQ(value);
             if (double.IsNaN(adjustedValue))
                values[i++] = float.NaN;
-            else if (unit != null && !string.IsNullOrEmpty(value.Unit))
-               values[i++] = (float) dataColumn.Dimension.UnitValueToBaseUnitValue(dimension.Unit(value.Unit), adjustedValue);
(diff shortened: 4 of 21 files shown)


src/OSPSuite.Infrastructure.Import/Core/DataFormat/MixColumnsDataFormat.cs
```diff
@@ -53,7 +53,7 @@ protected override UnitDescription ExtractUnits(string description, IUnformatted
          var unitKey = data.GetHeaders().FindHeader(description + "_UNIT");
          if (unitKey == null)
          {
-            return new UnitDescription();
+            return new UnitDescription(""); //in this case return the default unit for Dimensionless
```
georgeDaskalakis, line 56: is this OK here? Or should I actually be getting the default unit for dimensionless from the dimensionFactory? I was thinking since the unit in the header is empty, it seems more consistent to leave it like this - maybe dimensionless is not only available in the dimensions in the end, and it ends up being of dimension fraction instead.

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
```diff
@@ -59,12 +60,13 @@ public DataRepository ConvertImportDataSet(ImportedDataSet dataSet)
...
-         //for now it stays like this, cause otherwise we have to have the association
-         //var dimension_test = _dimensionFactory.Dimension(column.Key.Name);
-         //otherwise we can also just get the defaultDimension from the columnInfos when we
-         //create the extended column
-         var dimension = _dimensionFactory.DimensionForUnit(column.Key.Column.Unit.SelectedUnit) ?? Constants.Dimension.NO_DIMENSION;
+
+         IDimension dimension;
+         if (column.Key.Column.Dimension != null)
+            dimension = column.Key.Column.Dimension;
+         else
```
msevestre, line 67: what about you create a varialbe for `column.Key.Column.Unit.SelectedUnit` and use it?
(you actgualoly create one later on).

This syntax column.key.Column is very strange by the way...

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -104,6 +109,24 @@ private void setDataFormat(IList<DataFormatParameter> formatParameters)
...
          ValidateMapping();
          InitializeErrorUnit();
+         setDimensionsForAllMappings();
+      }
+
+      private void setDimensionsForAllMappings()
+      {
+         foreach (var mapping in _mappings)
+         {
+            if (mapping.Source == null) return;
```
msevestre: return or continue?
At any rate, on a new line

georgeDaskalakis: done

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
georgeDaskalakis: column renamed to columnAndData (it is a cache of column variables like Dimension and a list of simulation points)

georgeDaskalakis: variable "unit" intriduced earlier on

src/OSPSuite.Core/Serialization/Xml/ColumnXmlSerializer.cs
```diff
@@ -12,14 +13,23 @@ public override void PerformMapping()
...
          base.TypedDeserialize(objectToDeserialize, outputToDeserialize, context);
-         
+
          if (objectToDeserialize.ErrorStdDev == Constants.STD_DEV_GEOMETRIC)
+         {
             objectToDeserialize.Unit = new UnitDescription();
+            objectToDeserialize.Dimension = Constants.Dimension.NO_DIMENSION;
+         }
+
+         if (!objectToDeserialize.Unit.ColumnName.IsNullOrEmpty())
```
msevestre, line 29: Can Unit be NULL here?

src/OSPSuite.Core/Serialization/Xml/ColumnXmlSerializer.cs
```diff
@@ -12,14 +13,23 @@ public override void PerformMapping()
...
+         {
+            objectToDeserialize.Dimension = null;
```
msevestre, line 31: and why can this dimensio be NOT null already? It does not look like we have a lot of confidence in what is being serialized. Can we talk about this live?

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -104,6 +109,29 @@ private void setDataFormat(IList<DataFormatParameter> formatParameters)
...
+            if (mappingColumn?.Unit == null) 
+               continue;
+
+            //initial settings for fraction dimension
+            if (mapping.ColumnInfo.DefaultDimension?.Name == Constants.Dimension.FRACTION && 
+                mappingColumn.Unit.ColumnName.IsNullOrEmpty() && 
+                mappingColumn.Unit?.SelectedUnit == UnitDescription.InvalidUnit)
+            {
+               mappingColumn.Dimension = mapping.ColumnInfo.DefaultDimension;
+               mappingColumn.Unit = new UnitDescription("");
```
msevestre, line 129: this should be mappingColumn.Dimension.BaseUnit or sthg like this and not ""

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -104,6 +109,29 @@ private void setDataFormat(IList<DataFormatParameter> formatParameters)
...
+      {
+         foreach (var mapping in _mappings)
+         {
+            var mappingColumn = (mapping.Source as MappingDataFormatParameter)?.MappedColumn;
+            if (mappingColumn?.Unit == null) 
+               continue;
+
+            //initial settings for fraction dimension
+            if (mapping.ColumnInfo.DefaultDimension?.Name == Constants.Dimension.FRACTION && 
+                mappingColumn.Unit.ColumnName.IsNullOrEmpty() &&
```
msevestre, line 125: mappingColumn.Unit and mappingColumn.Unit?.. I think we need to be consistent in the same if...

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -190,6 +223,7 @@ public void UpdateDescriptionForModel(MappingDataFormatParameter mappingSource)
             {
                column.ErrorStdDev = Constants.STD_DEV_GEOMETRIC;
                column.Unit = new UnitDescription(""); //geometric error has no unit
```
msevestre, line 225: same stuff here. Use the dimension base unit

src/OSPSuite.Presentation/Presenters/Importer/UnitsEditorPresenter.cs
```diff
@@ -44,7 +47,8 @@ public void SelectDimension(string dimension)
       {
          this.DoWithinExceptionHandler(() =>
          {
-            _selectedUnit = (_dimensions.FirstOrDefault(d => d.Name == dimension) ?? _dimensions.First()).DefaultUnit.Name;
+            Dimension = _dimensions.FirstOrDefault(d => d.Name == dimension) ?? _dimensions.First();
```
msevestre, line 50: _dimensions.First();

What is _dimensions?

tests/OSPSuite.Infrastructure.Tests/Import/DataSetToDataRepositoryMapperSpecs.cs
```diff
@@ -12,14 +12,40 @@
 
 namespace OSPSuite.Infrastructure.Import
 {
+   public class DimensionForUnitTests : Dimension
```
msevestre, line 15: We have some dimensions available already for testing. DomainHelperForSpecs.xxx Please use those

msevestre (changes requested):

src/OSPSuite.Core/Serialization/SerializationTransaction.cs
```diff
@@ -25,7 +25,7 @@ public static class SerializationTransaction
          ICloneManagerForModel cloneManagerForModel = null, 
          IEnumerable<DataRepository> dataRepositories = null)
       {
-         return new SerializationContext(dimensionFactory, objectBaseFactory, withIdRepository, dataRepositories, cloneManagerForModel, container);
+         return new SerializationContext(dimensionFactory ?? container.Resolve<IDimensionFactory>(), objectBaseFactory, withIdRepository, dataRepositories, cloneManagerForModel, container);
```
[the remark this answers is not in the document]
abdelr: You are right. I fixed it on the caller side now.

src/OSPSuite.Core/Serialization/Xml/ColumnXmlSerializer.cs
abdelr: It could be empty. In case the units are set manually there is no ColumnName. I would not expect null but just to be on the safe side with code evolving in the future I rather keep IsNullOrEmpy() and not != null.

src/OSPSuite.Core/Serialization/Xml/ColumnXmlSerializer.cs
abdelr: I agree with you @msevestre. I have just made some changes that should make the process more reliable. So, before there was a problem when saving the dimension in case of geometric error. Now the dimension is properly set when serializing so we do not need any check when deserializing. I know @georgeDaskalakis was fixing some different errors so I would like to discuss these changes with him when he returns so we are sure we did not overlooked anything and that is why I am adding this comment so I know where to look after.

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -104,6 +109,29 @@ private void setDataFormat(IList<DataFormatParameter> formatParameters)
...
+            if (mapping.ColumnInfo.DefaultDimension?.Name == Constants.Dimension.FRACTION && 
+                mappingColumn.Unit.ColumnName.IsNullOrEmpty() && 
+                mappingColumn.Unit?.SelectedUnit == UnitDescription.InvalidUnit)
+            {
+               mappingColumn.Dimension = mapping.ColumnInfo.DefaultDimension;
+               mappingColumn.Unit = new UnitDescription("");
+               continue;
+            }
+
+            mappingColumn.Dimension = !mappingColumn.Unit.ColumnName.IsNullOrEmpty() ? null : _dimensionFactory.DimensionForUnit(mappingColumn.Unit.SelectedUnit);
```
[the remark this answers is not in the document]
abdelr: Well, on line 120 there is a 
`if (mappingColumn?.Unit == null) 
               continue;`
meaning at this point in the code, mappingColumn cannot be null anymore, right? Here the check is on ColumnName  and it could be empty string (not null)

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
abdelr: You are right, there is no way mappingColumn.Unit could be null here so no need for the check. Fixed.

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
abdelr: Fixed

src/OSPSuite.Presentation/Presenters/Importer/UnitsEditorPresenter.cs
abdelr: _dimensions is the list of authorized dimensions as you mentioned. Now the check is find the FirstOrDefault with this name (so if the name came wrong from the caller it won't crash) and if it was not found (default is returned) then give me the first dimension no matter the name. I am not sure how to optimize in one single line. If the problem is readability as I think, then it might be something like:
`Dimension = _dimensions.FirstOrDefault(d => d.Name == dimension);`
`if (Dimension == null) Dimension = _dimensions.First();`

abdelr: Although I see no big difference in readability between the current code and the one I have just proposed. Maybe some comments?

tests/OSPSuite.Infrastructure.Tests/Import/DataSetToDataRepositoryMapperSpecs.cs
abdelr: Done

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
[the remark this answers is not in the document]
msevestre: oh yeah missed it in the diff. Thanks

src/OSPSuite.Presentation/Presenters/Importer/UnitsEditorPresenter.cs
msevestre: yeah I Was just wondering if _dimensiosn could be empty that's all. If yes, First will throw

tests/OSPSuite.Starter/Bootstrapping/ApplicationStartup.cs
```diff
@@ -46,7 +46,7 @@ public class ApplicationStartup
       public static void Initialize()
       {
          initializeDependency();
-         fillDimensions(IoC.Resolve<IDimensionFactory>());
+         fillDimensions(IoC.Container.Resolve<IDimensionFactory>());
```
msevestre, line 49: Why Container?? Does not matter but this should not be required

msevestre (commented):
Looks good to me

tests/OSPSuite.Starter/Bootstrapping/ApplicationStartup.cs
abdelr: You are right... I have just removed it...

msevestre:
@abdelr Ok I just checked and the list of _diomensiosn can indeed be empty and the call will break. I will fix this in the next PR as well

PR #1077, opened 2021-06-04
1067 import configuration not read

georgeDaskalakis:


src/OSPSuite.Core/Import/ImporterConfiguration.cs
```diff
@@ -14,11 +17,13 @@ public void CloneParametersFrom(IReadOnlyList<DataFormatParameter> parameters)
          _parameters = new List<DataFormatParameter>(parameters);
       }
       public void AddParameter(DataFormatParameter parameter) { _parameters.Add(parameter); }
-      public List<string> LoadedSheets { get; } = new List<string>();
+      public List<string> LoadedSheets { get => _loadedSheets.ToList(); }
```
msevestre, line 20: Can't yuo just return an IEnumerable<string> ?

msevestre: I think this can be an enumerable. Not a big deal
@georgeDaskalakis Please fix when you get back

PR #1080, opened 2021-06-08
Fixes #1079 and Fixes #1058

### LINKED ISSUE #1058: Remove "DimensionInfo"

### LINKED ISSUE #1079: Import Ontogeny and cancel => crash
@msevestre commented on [Tue Jun 08 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1786)

FYI: @abdelr 

Application:
PK-Sim® 10.0.0

Value cannot be null.
Parameter name: source

Stack trace:
```
at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source)
   at PKSim.Presentation.Services.OntogenyTask`1.LoadOntogenyForMolecule(IndividualMolecule molecule, TSimulationSubject simulationSubject) in C:\projects\PK-Sim\src\PKSim.Presentation\Services\OntogenyTask.cs:line 65
   at PKSim.Presentation.Presenters.Individuals.OntogenySelectionPresenter`1.LoadOntogeny() in C:\projects\PK-Sim\src\PKSim.Presentation\Presenters\Individuals\OntogenySelectionPresenter.cs:line 96
```

---

@msevestre commented on [Tue Jun 08 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1786#issuecomment-856836186)

@abdelr The reason is that upon cancel,. the method ImportDataSets returns NULL and not an empty liust

msevestre:


src/OSPSuite.Infrastructure.Import/Core/ColumnInfo.cs
```diff
@@ -10,7 +10,7 @@ public class ColumnInfo
 
       public string Name
       {
-         get { return _name; }
+         get => _name;
```
msevestre, line 13: Resharper suggestion to shorten the code. Let's use them unless they make the code less readable

src/OSPSuite.Infrastructure.Import/Core/ColumnInfo.cs
```diff
@@ -29,11 +29,17 @@ public string Name
       ///    Column name of related column. The related column must have an auxiliary type as meta data category.
       /// </summary>
       public string RelatedColumnOf { get; set; }
-      public IList<DimensionInfo> DimensionInfos { get; private set; }
+
+      /// <summary>
+      ///    List of possible supported dimensions.This List may be empty if all dimensions are supported
```
msevestre, line 34: I am not sure if this is true or not. Or should the caller add all dimensions?

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -290,26 +292,24 @@ public void SetSubEditorSettingsForMapping(ColumnMappingDTO model)
...
-            var errorColumn = ((MappingDataFormatParameter)errorColumnDTO?.Source)?.MappedColumn;
-            if (errorColumn?.Unit != null && !errorColumn.Unit.ColumnName.IsNullOrEmpty())
-            {
+            var errorColumn = ((MappingDataFormatParameter) errorColumnDTO?.Source)?.MappedColumn;
+
+            if (errorColumn?.Unit != null && !errorColumn.Unit.ColumnName.IsNullOrEmpty()) 
                columns.Add(errorColumn.Unit.ColumnName);
-            }
 
             dimensions.AddRange(_columnInfos
```
msevestre, line 300: Here. THe ist can potentially be empty. Then the next call to setUnitOPtiosn makes a First and this could crash

msevestre: all the other changes in this file are because of wrong formatting.

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
```diff
@@ -21,28 +22,29 @@ public ModalImporterPresenter(IModalImporterView view) : base(view)
       }
 
       public (IReadOnlyList<DataRepository> DataRepositories, ImporterConfiguration Configuration) ImportDataSets(
-         IImporterPresenter presenter, 
+         IImporterPresenter presenter,
          string configurationId = null
       )
       {
-         IReadOnlyList<DataRepository> result = null;
+         IReadOnlyList<DataRepository> results = Array.Empty<DataRepository>();
```
msevestre, line 29: This was the bug here. The results were null upon creation but we should always return something.
The way this presenter is setup makes it very hard to test. Why is the ImporterPresenter passed a parameter and not in the constructor? This looks to me like a potential leak also

src/OSPSuite.Presentation/Presenters/Importer/IUnitsEditorPresenter.cs
```diff
@@ -6,11 +6,11 @@ namespace OSPSuite.Presentation.Presenters.Importer
 {
    public interface IUnitsEditorPresenter : IDisposablePresenter
    {
-      void SetOptions(Column importDataColumn, IEnumerable<IDimension> dimensions, IEnumerable<string> availableColumns);
+      void SetOptions(Column importDataColumn, IReadOnlyList<IDimension> dimensions, IEnumerable<string> availableColumns);
```
msevestre, line 9: IReadOnly whenever possible. We almost always have Lists etc.. anyways.

src/OSPSuite.Presentation/Presenters/Importer/UnitsEditorPresenter.cs
```diff
@@ -10,44 +11,41 @@ namespace OSPSuite.Presentation.Presenters.Importer
...
       }
 
-      public void SelectDimension(string dimension)
+      public void SelectDimension(string dimensionName)
       {
          this.DoWithinExceptionHandler(() =>
          {
-            Dimension = _dimensions.FirstOrDefault(d => d.Name == dimension) ?? _dimensions.First();
+            Dimension = _dimensions.FirstOrDefault(d => string.Equals(d.Name, dimensionName)) ??
+                        _dimensions.FirstOrDefault() ??
```
msevestre, line 46: here. First can crash.

src/OSPSuite.UI/Views/Importer/ModalImporterView.cs
```diff
@@ -22,26 +22,21 @@ public override void InitializeResources()
...
       }
 
-
       public void AttachImporterPresenter(IImporterPresenter presenter)
       {
-         presenter.OnTriggerImport += (s, d) =>
-         {
-            DialogResult = System.Windows.Forms.DialogResult.OK;
-         };
+         presenter.OnTriggerImport += (s, d) => { DialogResult = DialogResult.OK; };
```
msevestre, line 34: This is so weird. A view that knows about another presenter and reacts to event... why?

src/OSPSuite.Infrastructure.Import/Core/ColumnInfo.cs
abdelr: I am not sure either. But, is there any case where we have all dimensions?

src/OSPSuite.Infrastructure.Import/Core/ColumnInfo.cs
```diff
@@ -38,15 +35,10 @@ public string Name
       public IDimension DefaultDimension { get; set; }
       public IList<MetaDataCategory> MetaDataCategories { get; }
 
-      public Type DataType { get; set; }
```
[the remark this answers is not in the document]
abdelr: I am not sure we are strongly relying on data types. Actually all we need to differentiate for the importer is if the data is categorical or numeric.

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
abdelr: If you see the use of it on DataImporter service, you will see that the ImporterPresenter is configured there and then passed to the ModalImporterPresenter. So, the DataImporter is responsible for handling the ImporterPresenter while the ModalImporterPresenter is just presenting it as a modal view. I am wondering if we should have a single class for both but then, DataImporter is not aware of any UI specifics, so it is just a service while ModalImporterPresenter is UI specific. This was the way to separate the service from the UI which was mixed before. So, all in all, the DataImporter is a service you could potentially (not currently) use without DevExpress and the way to present it is by means of the ModalImporterPresenter. Maybe in the future we will need the importer logic from R or CLI and we will need a different implementation of the service (DataImporter) that will do more or less the same by means of the console or by interactions with R wrappers. We could think this better later if you wish.

src/OSPSuite.UI/Views/Importer/ModalImporterView.cs
abdelr: Look comment above on what is the intention of the ModalImporterView. This is just presenting the DataImporterPresenter as a dialog and when it imports, it should close the dialog with OK message. Again, we could rethink this a little bit.

abdelr (approved):

src/OSPSuite.Infrastructure.Import/Core/ColumnInfo.cs
msevestre: yes. MoBi adds everything for instance in a loop

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
msevestre: I am talking bout the ImporterPresneter not the DataImporter. There are two presenters created and one is passed to the other. This is what is very weird and makes testing very hard to write also. At any rate, it works

src/OSPSuite.UI/Views/Importer/ModalImporterView.cs
msevestre: a view should never know of another presenter. This breaks encapsulation. Too late for now

PR #1083, opened 2021-06-10
1081 alignment

### LINKED ISSUE #1081: Combo box height for Error Type is too  large.
image

For some reason, the combo box is larger than the default size. can we use the default size here?

abdelr:
Fixes #1081


msevestre:
merging to see the output. NExt time, I'd suggest to put screenshots of the fix :))

abdelr:
> merging to see the output. NExt time, I'd suggest to put screenshots of the fix :))

Sure, I was actually wondering about it since the diff says nothing in this case :)

PR #1084, opened 2021-06-10
Conditionally prompt

### LINKED ISSUE #1068: Observed data / Reload under same settings: inconsistent state
[Organs1.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/6569866/Organs1.xlsx) contains observed data for **Brain** and **Liver**.
[Organs2.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/6569867/Organs2.xlsx) contains observed data for **Brain** and **Heart**.
I import Organs1.xlsx first.
grafik
Now I call "Reload under the same settings" and select Organs2.xlsx
The confirmation dialog is shown
grafik
I click OK. Another confirmation dialog is shown
grafik
Here, I can see that the new Observed data set for heart was already added bevor I confirm the deletion of Liver data.
Now I select **No** - and end up in inconsistent state, where observed data from both Organs1.xlsx and Organs2.xlsx is in the project.
grafik

So I think the update order is wrong: new datasets are added first (and changed datasets are probably updated as well) and only after that datasets which should be deleted are deleted.
It should be done the other way around: try to delete first and only if the user confirms deletion - continue with adding/updating

abdelr:
fixes #1068

### DIFF
--- a/src/OSPSuite.Core/Domain/Services/ObservedDataTask.cs
+++ b/src/OSPSuite.Core/Domain/Services/ObservedDataTask.cs
@@ -18,9 +18,10 @@ namespace OSPSuite.Core.Domain.Services
 
       /// <summary>
       ///    Deletes the <paramref name="observedData" /> from the project.
+      ///    If <paramref name="promptBeforeDeleting"/> then prompts the user before deleting
       ///    Returns <c>true</c> if the deletion was confirm by the user otherwise <c>false</c>
       /// </summary>
-      bool Delete(DataRepository observedData);
+      bool Delete(DataRepository observedData, bool promptBeforeDeleting = true);
 
       /// <summary>
       ///    Deletes the <paramref name="observedDataEnumerable" /> from the project. User prompt can be turned off (<paramref name="silent"/> set to <c>true</c>).
@@ -118,16 +119,19 @@ namespace OSPSuite.Core.Domain.Services
 
       public abstract void UpdateMolWeight(DataRepository observedData);
 
-      public bool Delete(DataRepository observedData)
+      public bool Delete(DataRepository observedData, bool promptBeforeDeleting = true)
       {
          var usersOfObservedData = allUsersOfObservedData(observedData).ToList();
 
          if (usersOfObservedData.Any())
             throw new CannotDeleteObservedDataException(observedData.Name, usersOfObservedData.Select(typeNamed).ToList());
 
-         var viewResult = _dialogCreator.MessageBoxYesNo(Captions.ReallyDeleteObservedData(observedData.Name));
-         if (viewResult == ViewResult.No)
-            return false;
+         if (promptBeforeDeleting)
+         {
+            var viewResult = _dialogCreator.MessageBoxYesNo(Captions.ReallyDeleteObservedData(observedData.Name));
+            if (viewResult == ViewResult.No)
+               return false;
+         }
 
          var removeCommand = new RemoveObservedDataFromProjectCommand(observedData).Run(_executionContext);
          _executionContext.AddToHistory(removeCommand);


src/OSPSuite.Core/Domain/Services/ObservedDataTask.cs
```diff
@@ -18,9 +18,10 @@ public interface IObservedDataTask
 
       /// <summary>
       ///    Deletes the <paramref name="observedData" /> from the project.
+      ///    If <paramref name="promptBeforeDeleting"/> then prompts the user before deleting
       ///    Returns <c>true</c> if the deletion was confirm by the user otherwise <c>false</c>
       /// </summary>
-      bool Delete(DataRepository observedData);
+      bool Delete(DataRepository observedData, bool promptBeforeDeleting = true);
```
msevestre, line 24: I don't think this is right as the prompt will be asked  FOR each single data repo. I think the method should delete using the other method taking a IEnumerable. That's why I did not add the prompt on this method in the first place

msevestre (changes requested):
I think this should be reverted and the delete method should call the list with the IEnumerable

src/OSPSuite.Core/Domain/Services/ObservedDataTask.cs
abdelr: @msevestre I think it makes more sense to pass the IEnumerable but still this will not answer the original question. We would like to control from the caller whether we want to prompt the user or not "do you really want to delete..?" So in some cases it makes sense to ask but if you look into the issue discussion, Pavel and I agreed on Juri's possible course of action where we do not to ask in case of a reload since the user was already informed about what was going to be deleted. So from the user point of view it is like:
1. I will remove this, update this and add this -> ok/cancel
2. Do you really want to remove this? -> yes/no

I normally do not like when I am prompt that much about something and we have the history for undoing commands so the final proposal was not to prompt in this case. So, all in all, either we have a boolean promptBeforeDeleting or we have two methods one for prompting and another one without it. Sure, internally they both will use the same private method in case we have two public methods for this and the private will have then the flag.

All in all, I agree it is better to pass the whole IEnumerable and then the delete should prompt only once in case of a list but it should also have the option not to prompt at all.

msevestre: Wel the option with the IEnumeralbe also as the flag. So I do not see a contradiction here?

msevestre: https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/7ed7872ceeef5b204c5152374cd06634d38ee4be/src/OSPSuite.Core/Domain/Services/ObservedDataTask.cs#L30

msevestre: and it's called `silent`. We should at least align the name of the flags...

msevestre: SO my suggestion is till
1-Undo the changes
2-In the method, instead of calling Delete (obsData, prompt=false) for each obsData to delete, do Delete(ListOfObsData, silent=true)

msevestre (changes requested):

msevestre:
Happy to discuss if this is easier.

src/OSPSuite.Core/Domain/Services/ObservedDataTask.cs
abdelr: There is no contradiction. I was stressing that we need both things ;)

abdelr:
> Happy to discuss if this is easier.

I do not think we have to discuss it. The was I see it we both agree ;)

abdelr:
Wait, I overlooked something. There is a Delete for a single element and a Delete for a list with a silent mode but they do not interact with each other. I would expect one method to invoke the other, otherwise future maintenance could be problematic.

PR #1086, opened 2021-06-11
Adding first rows to the label

### LINKED ISSUE #1082: Import of CSV: show first rows
When importing a CSV, a dialogue is presented to select the separator.

- By default, `,` should be the pre-selected separator.
- Show the first 2 rows of the file (raw data?) to allow the user to "guess" the correct separator.

abdelr:
Fixes #1082

### DIFF
--- a/src/OSPSuite.UI/Views/Importer/CsvSeparatorSelectorView.cs
+++ b/src/OSPSuite.UI/Views/Importer/CsvSeparatorSelectorView.cs
@@ -16,6 +16,12 @@ namespace OSPSuite.UI.Views.Importer
          separatorComboBoxEdit.EditValueChanged += onSeparatorChanged;
       }
 
+      public override void InitializeResources()
+      {
+         base.InitializeResources();
+         separatorDescriptionLabelControl.Appearance.TextOptions.WordWrap = DevExpress.Utils.WordWrap.Wrap;
+      }
+
       private void fillSeparatorComboBox()
       {
          var separatorList = new List<char>() {',', '.', ' ', ';'};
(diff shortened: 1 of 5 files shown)


src/OSPSuite.UI/Views/Importer/CsvSeparatorSelectorView.cs
```diff
@@ -38,8 +38,25 @@ public void AttachPresenter(ICsvSeparatorSelectorPresenter presenter)
 
       public void SetFileName(string fileName)
       {
+         separatorDescriptionLabelControl.Appearance.TextOptions.WordWrap = DevExpress.Utils.WordWrap.Wrap;
          separatorDescriptionLabelControl.Text = Captions.Importer.CsvSeparatorDescription(fileName);
+         addRowsToText(fileName);
          separatorComboBoxEdit.SelectedIndex = 0;
       }
+
+      private void addRowsToText(string fileName)
```
msevestre, line 47: This is the right idea but not the right place. This should be moved to the presenter and passed to the view as "data" I thinkso that it can be tested.


Last, there is for sure a using() missing somewhere
From 
https://docs.microsoft.com/en-us/dotnet/api/system.io.streamreader.readline?view=net-5.0

using (StreamReader sr = new StreamReader(path))
            {
                while (sr.Peek() >= 0)
                {
                    Console.WriteLine(sr.ReadLine());
                }
            }

msevestre (changes requested):

src/OSPSuite.UI/Views/Importer/CsvSeparatorSelectorView.cs
msevestre: but I would in fact use File.ReadLines 
https://docs.microsoft.com/en-us/dotnet/api/system.io.file.readlines?view=net-5.0

It is lazy evaluated so you can just do Take(3) and be good to go (to see if this work if the file only has 2)

The ReadLines and ReadAllLines methods differ as follows: When you use ReadLines, you can start enumerating the collection of strings before the whole collection is returned; when you use ReadAllLines, you must wait for the whole array of strings be returned before you can access the array. Therefore, when you are working with very large files, ReadLines can be more efficient.

PR #1087, opened 2021-06-14
fixes #1078

### LINKED ISSUE #1078: loading of second sheet results in exception
If the user tries to load more than one sheets of one excel documents, this results in an exception.

georgeDaskalakis:
closes #1078

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -412,6 +412,8 @@ namespace OSPSuite.Assets
          public static readonly string MappingTitle = "Mapping";
          public static readonly string ReloadWillCauseChangeOfDataSets =
             "Reloading will cause the following changes in observed data. Do you really want to reload?";
+         public static readonly string UnexpectedExceptionWhenLoading =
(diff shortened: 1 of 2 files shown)


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -412,6 +412,8 @@ public static class Importer
          public static readonly string MappingTitle = "Mapping";
          public static readonly string ReloadWillCauseChangeOfDataSets =
             "Reloading will cause the following changes in observed data. Do you really want to reload?";
+         public static readonly string UnexpectedExceptionWhenLoading =
+            "Unexpected exception when trying to load. File format probably not supported";
```
msevestre, line 416: I would not use the word "Exception" for user-facing errors

"An unexpected error occurred while loading the file. The file format is probably not supported. Please check the documentation for more details" 

And can we add the link to the documentation here?

msevestre (changes requested):
Small wording change

src/OSPSuite.Assets/UIConstants.cs
georgeDaskalakis: fixed :)

PR #1088, opened 2021-06-14
WIP refactor import presenter

msevestre:


src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
```diff
@@ -45,7 +97,7 @@ public ModalImporterPresenter(IModalImporterView view, IImporterPresenter import
             results.Each(r => r.ConfigurationId = configurationId);
          }
 
-         _view.AttachImporterPresenter(presenter);
+         _view.AttachImporterPresenter(_importerPresenter);
```
msevestre, line 100: I Still don't thin that this is required (as we discussed yesterday). Let's the presenter deals with its own view

msevestre (commented):
looks much cleaner! One change perhaps with the view using another presenter

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
abdelr: Sure, there is also something more to discuss. I do not like how the DataImporter service looks right now. If you check lines 122... it feels like this code should not be there. On the one hand ModalImporterPresenter takes care of the UI for the importer but it also prepares everything (SetSettings, SetSourceFile...), the code remaining in DataImporter is not UI related so I do not think should land in ModalImporterPresenter but it is strongly related to the code on ModalImporterPresenter since it is mimicking the same behavior (without the UI) as if you would open the view and click Import right after. I need to think on this a bit more but feel free to give any suggestion ;)

PR #1096, opened 2021-06-17
1092 fixing icons for organs

### LINKED ISSUE #1092: PeripheralVenusBlood has no icon
Also, check alphabetical order (predefined then columns)

abdelr:
Fixes #1092 (Special icons for organs and proper order)

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
@@ -399,7 +399,8 @@ namespace OSPSuite.Presentation.Presenters.Importer
          }
 
          options.AddRange(excelColumns.Select(c => new RowOptionDTO() {Description = c, ImageIndex = ApplicationIcons.IconIndex(ApplicationIcons.ObservedDataForMolecule)}));
-         return options.OrderBy(o => o.ImageIndex).ThenBy(o => o.Description);
+         var metaDataIconIndex = ApplicationIcons.IconIndex(ApplicationIcons.ObservedDataForMolecule);
+         return options.OrderBy(o => o.ImageIndex == metaDataIconIndex).ThenBy(o => o.Description);
       }
 
       public IEnumerable<ColumnMappingOption> GetAvailableOptionsFor(ColumnMappingDTO model)
--- a/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
@@ -377,6 +377,10 @@ namespace OSPSuite.Presentation.Presenters.Importer
             {
                options.AddRange(metaDataCategory.ListOfValues.Keys.Select(v =>
                {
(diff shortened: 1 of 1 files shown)


Yuri05 (approved):

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -395,7 +399,8 @@ public IEnumerable<RowOptionDTO> GetAvailableRowsFor(ColumnMappingDTO model)
          }
 
          options.AddRange(excelColumns.Select(c => new RowOptionDTO() {Description = c, ImageIndex = ApplicationIcons.IconIndex(ApplicationIcons.ObservedDataForMolecule)}));
-         return options.OrderBy(o => o.ImageIndex).ThenBy(o => o.Description);
+         var metaDataIconIndex = ApplicationIcons.IconIndex(ApplicationIcons.ObservedDataForMolecule);
+         return options.OrderBy(o => o.ImageIndex == metaDataIconIndex).ThenBy(o => o.Description);
```
msevestre, line 403: I don't understand this? We are sorting by Icon??

msevestre (commented):
@Yuri05 Approved so I Am going to merge but I don't understand the code change

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -377,6 +377,10 @@ public IEnumerable<RowOptionDTO> GetAvailableRowsFor(ColumnMappingDTO model)
             {
                options.AddRange(metaDataCategory.ListOfValues.Keys.Select(v =>
                {
+                  metaDataCategory.ListOfImages.TryGetValue(v, out var value);
```
msevestre, line 380: I have no idea what's going on here
Why are we trying to find the image based on the value ?  Shouldn't those two lists have the same length by construction?

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
abdelr: We want the excel columns (shown with metaDataIconIndex) to be listed at the end. Otherwise, by description alphabetically.

msevestre: So you are sorting by a boolean (FALSE, FALSE, TRUE, TRUE). ?Interesting

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
abdelr: I thought they should, but they do not have the same length. Actually, the ListOfImages is rarely used. So currently we check first if there is an image already for this key and if there is none, then we check the icons by the name of the organ, compartment... PeripheralVenousBlood for example has an image on ListOfImages but most of the other organs do not.

msevestre: ok.... somehow this is all surprising but clearly sthg that was there before the refactoring

PR #1099, opened 2021-06-17
Update AddImporterConfigurationToProjectCommand.cs

### LINKED ISSUE #1098: App crashes when cancelling import

abdelr:
Fixes #1098 (App crashes when cancelling import)

### DIFF
--- a/src/OSPSuite.Core/Commands/AddImporterConfigurationToProjectCommand.cs
+++ b/src/OSPSuite.Core/Commands/AddImporterConfigurationToProjectCommand.cs
@@ -19,9 +19,6 @@ namespace OSPSuite.Core.Commands
 
       protected override void ExecuteWith(IOSPSuiteExecutionContext context)
       {
-         if (string.IsNullOrEmpty(_configuration.Id))
-            return;
-
          var project = context.Project;
          project.AddImporterConfiguration(_configuration);
          Description = Assets.Command.AddObservedDataToProjectDescription(_configuration.FileName, project.Name);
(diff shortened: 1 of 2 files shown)


Yuri05 (approved):

src/OSPSuite.Core/Commands/AddImporterConfigurationToProjectCommand.cs
```diff
@@ -19,6 +19,9 @@ public AddImporterConfigurationToProjectCommand(ImporterConfiguration configurat
 
       protected override void ExecuteWith(IOSPSuiteExecutionContext context)
       {
+         if (string.IsNullOrEmpty(_configuration.Id))
```
msevestre, line 22: Why do we trigger a command for an invalid configuration? The command will be added to the stack of executed command even though nothing happened. I think this is not ok

msevestre (changes requested):

src/OSPSuite.Core/Commands/AddImporterConfigurationToProjectCommand.cs
abdelr: You are right, I already cloned the issue on PKSim [1794](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1794) and fix it there, so the command will not be triggered on the first place. I would still keep this check to avoid crash on non-intended usage.

Yuri05 (approved):

src/OSPSuite.Core/Commands/AddImporterConfigurationToProjectCommand.cs
msevestre: hum...I almos twant to it to crash to be honest so that we are aware of non intended usage. THis SHOULD NEVER HAPPEN if we are calling this stuff properly so IMO, this is an exception that should be thrown

abdelr: Yes, you are right... I already remove the check.

PR #1100, opened 2021-06-17
1095 compartments order

### LINKED ISSUE #1095: OSP10 beta: Order of compartments (Importer-Redesign)
A very minor thing: In the pull-down menu of the ”Compartments” within the  "Mapping Name" field of the redesigned data import: The order of compartments seems strange to me. First there are the compartments Blood Cells, Interstitial & Intracellular, then there some others and at the end there is Plasma. This order has the consequence that one has to scroll to the compartment which is selected most. I would put Plasma to its peers at the beginning.

abdelr:
Fixes #1095 (Plasma and Venous Blood goes to the top of the list)

### DIFF
--- a/src/OSPSuite.Core/Domain/Constants.cs
+++ b/src/OSPSuite.Core/Domain/Constants.cs
@@ -456,6 +456,9 @@ namespace OSPSuite.Core.Domain
          public static readonly string PATIENT_ID = "Patient Id";
          public static readonly string GENDER = "Gender";
          public static readonly string DOSE = "Dose";
+         public static readonly string PLASMA_COMPARTMENT = "Plasma";
+         public static readonly string PERIPHERAL_VENUS_BLOOD_ORGAN = "Peripheral Venous Blood";
+         public static readonly string VENUS_BLOOD_ORGAN = "Venous Blood";
       }
 
       public static class LLOQModes
--- a/src/OSPSuite.Core/Domain/Constants.cs
+++ b/src/OSPSuite.Core/Domain/Constants.cs
@@ -458,7 +458,7 @@ namespace OSPSuite.Core.Domain
          public static readonly string DOSE = "Dose";
          public static readonly string PLASMA_COMPARTMENT = "Plasma";
          public static readonly string PERIPHERAL_VENUS_BLOOD_ORGAN = "Peripheral Venous Blood";
-         public static readonly string VENUS_BLOOD_ORGAN = "Venous Blood";
+         public static readonly string VENUS_BLOOD_ORGAN = "VenousBlood";
       }
 
       public static class LLOQModes
--- a/src/OSPSuite.Core/Domain/Constants.cs
+++ b/src/OSPSuite.Core/Domain/Constants.cs
@@ -457,7 +457,7 @@ namespace OSPSuite.Core.Domain
          public static readonly string GENDER = "Gender";
          public static readonly string DOSE = "Dose";
          public static readonly string PLASMA_COMPARTMENT = "Plasma";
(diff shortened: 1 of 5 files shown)


src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -456,6 +456,9 @@ public static class ObservedData
          public static readonly string PATIENT_ID = "Patient Id";
          public static readonly string GENDER = "Gender";
          public static readonly string DOSE = "Dose";
+         public static readonly string PLASMA_COMPARTMENT = "Plasma";
+         public static readonly string PERIPHERAL_VENUS_BLOOD_ORGAN = "Peripheral Venous Blood";
+         public static readonly string VENUS_BLOOD_ORGAN = "VenousBlood";
```
Yuri05, line 461: typo in the const name: VEN**O**US_BLOOD_ORGAN

src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -456,6 +456,9 @@ public static class ObservedData
          public static readonly string PATIENT_ID = "Patient Id";
          public static readonly string GENDER = "Gender";
          public static readonly string DOSE = "Dose";
+         public static readonly string PLASMA_COMPARTMENT = "Plasma";
+         public static readonly string PERIPHERAL_VENUS_BLOOD_ORGAN = "Peripheral Venous Blood";
```
Yuri05, line 460: peripheral ist still venus :)

abdelr: Right... :(

Yuri05 (approved):

src/OSPSuite.UI/Services/DataImporter.cs
```diff
@@ -47,10 +47,13 @@ public IList<MetaDataCategory> DefaultMetaDataCategories()
 
          var organCategory = createMetaDataCategory<string>(Constants.ObservedData.ORGAN, isMandatory: true, isListOfValuesFixed: true);
          organCategory.Description = ObservedData.ObservedDataOrganDescription;
+         organCategory.TopNames.Add(Constants.ObservedData.PERIPHERAL_VENOUS_BLOOD_ORGAN);
```
msevestre, line 50: I don't quite understand why this is being done in CORE and not in the caller. What happens if this is also done in PKSim? Will the entry be duplicated? If feels to me that the order should be dictated by the caller of and not by the  core... @Yuri05 thoughts?

Yuri05: Yep, think you are right. If the caller fills the initial list it's better to add "top names" by the caller as well

abdelr: @Yuri05 and @msevestre I am not sure. On the one hand I agree that the caller should be managing this but on the second hand, keep in mind that DefaultMetaDataCategories is only giving the callers a default (or base if you would prefer) configuration for the MetaDataCategories and the caller could then modify anything such as the "top names". Of course if you think in MoBi these "top names" do not make sense I would say we should move them to the PKSim side only but if it does make sense, and I think it does right?, then we should keep them here since the callers could still further edit these MetaDataCategories before feeding them back to Core.

msevestre:
@abdelr Can you resolve the conflict on the DataImproter file

abdelr:
> @abdelr Can you resolve the conflict on the DataImproter file

@msevestre I just merged develop into this branch and it looks like there is no conflicts.

msevestre:
The build is failing now :)

abdelr:
> The build is failing now :)

Yes, I noticed. It was because of a very weird change committed. I reverted it but I wonder how did it end up in the history anyways. I would wait for appveyor but should be fine now.

PR #1101, opened 2021-06-17
Update DataSetToDataRepositoryMapper.cs

### LINKED ISSUE #1090: LLOQ values from column are not used although configured
PKSim Version from 15.6.2021
Import attached file.
Filter for molecules GLP-1_7...  and Insulin and load and import
For the concentration values is the LLOQ value given in the concentration_LLOQ column:
for first molecule LLOQ = 15 , for the second molecule LLOQ = 100.
But not in the preview nor in the import results in PKSim the LLOQ is used.

I would expect display of LLOQ in the preview both in the table right of Error and in the chart.
[CSV JJ.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/6662781/CSV.JJ.xlsx)
LLOQnotUsed

abdelr:
Fixes #1090 (Show LLOQ)

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
@@ -100,7 +100,6 @@ namespace OSPSuite.Infrastructure.Import.Core.Mappers
          {
             if (!double.IsNaN(value.Lloq))
                lloqValue = value;
-            columnAndData.Value.FirstOrDefault(v => !double.IsNaN(v.Lloq));
             var adjustedValue = truncateUsingLLOQ(value);
             if (double.IsNaN(adjustedValue))
                values[i++] = float.NaN;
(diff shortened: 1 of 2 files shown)


src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
```diff
@@ -105,6 +108,8 @@ private void convertParsedDataColumn(DataRepository dataRepository, KeyValuePair
             else
                values[i++] = (float) adjustedValue;
          }
+         if (lloqValue != null)
+            dataInfo.LLOQ = Convert.ToSingle(dimension.UnitValueToBaseUnitValue(dimension.FindUnit(lloqValue.Unit), lloqValue.Lloq));
```
msevestre, line 112: I think this should really be tested. A core function, no UI.

msevestre (commented):
Ok to merge but I'd like to see tests for this

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
abdelr: I agree... I will add some test on next week

abdelr: I fixed the ignored tests for the mapper and added a new one for the lloq

msevestre (approved):
Looks. Good. if possible, let's use the BDDHelper methods next time (ShouldBeEqualTo etc..)

PR #1116, opened 2021-06-18
Adding tooltip and correcting label

### LINKED ISSUE #1104: Importer: Add tooltip to "Use the filters for importing the data"
image

1) I would rename it to "Use filters for importing data" (get rid of "the" to save space)
2) Add a tool tip briefly explaining what it does.

abdelr:
Fixes #1104 (Fix label of filters checkbox and adding tooltip)

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -440,7 +440,7 @@ namespace OSPSuite.Assets
          }
 
          public static readonly string UseFiltersForImport = "Use filters for importing data";
-         public static readonly string UseFiltersForImportTooltip = "When checked, the data comming from the file will also be filter by the importer. When unchecked, the filters only affect this view. Check documentation for more information on defining filters: <href=https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering>https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering</href>";
+         public static readonly string UseFiltersForImportTooltip = "When selected, the filter will apply to the data during the import process. When deselected, the filter only affects this view. Check documentation for more information on defining filters: <href=https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering>https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering</href>";
 
          public static readonly string AddGroupByTitle = "Add group by";
          public static readonly string MetaDataTitle = "Meta data";
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -440,7 +440,7 @@ namespace OSPSuite.Assets
          }
 
          public static readonly string UseFiltersForImport = "Use filters for importing data";
-         public static readonly string UseFiltersForImportTooltip = "When checked, the data comming from the file will also be filter by the importer. When unchecked, the filters only affect this view";
+         public static readonly string UseFiltersForImportTooltip = "When checked, the data comming from the file will also be filter by the importer. When unchecked, the filters only affect this view. Check documentation for more information on defining filters: <href=https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering>https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering</href>";
 
          public static readonly string AddGroupByTitle = "Add group by";
          public static readonly string MetaDataTitle = "Meta data";
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -439,8 +439,9 @@ namespace OSPSuite.Assets
             return sb.ToString();
          }
 
-         public static readonly string UseFiltersForImport = "Use the filters for importing the data";
-         
+         public static readonly string UseFiltersForImport = "Use filters for importing data";
+         public static readonly string UseFiltersForImportTooltip = "When checked, the data comming from the file will also be filter by the importer. When unchecked, the filters only affect this view";
+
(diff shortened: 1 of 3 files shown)


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -439,8 +439,9 @@ public static string SheetsNotFound(List<string> listOfSheetNames)
             return sb.ToString();
          }
 
-         public static readonly string UseFiltersForImport = "Use the filters for importing the data";
-         
+         public static readonly string UseFiltersForImport = "Use filters for importing data";
+         public static readonly string UseFiltersForImportTooltip = "When checked, the data comming from the file will also be filter by the importer. When unchecked, the filters only affect this view";
```
PavelBal, line 443: "When checked, filters applied in the preview will also be applied to imported data. When unchecked, the filters only affect this view"

It took me some time to realize that there are filters in the preview-table (and I am not sure many users know of the existence of that filters).

PavelBal (changes requested):

src/OSPSuite.Assets/UIConstants.cs
Yuri05: >It took me some time to realize that there are filters in the preview-table (and I am not sure many users know of the existence of that filters).

After so many years with PK-Sim/MoBi? lol
There is a good guide by DevExpress for Filters: https://docs.devexpress.com/WindowsForms/114645/controls-and-libraries/data-grid/getting-started/walkthroughs/filter-and-search/tutorial-filter-editor
It's for developers, but the first half of this page describes it nicely from the user perspective - I think we should link it in the documentation.
We could also add the link to the tooltip of the "Use filters..." checkbox.

Yuri05: P.S. OSP docu gives also some info on it:
https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering

PavelBal (approved):

src/OSPSuite.UI/Views/Importer/ImporterDataView.resx
```diff
@@ -120,4 +120,7 @@
   <metadata name="errorProvider.TrayLocation" type="System.Drawing.Point, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
     <value>17, 17</value>
   </metadata>
+  <data name="useForImportCheckEdit.ToolTip" xml:space="preserve">
+    <value>When checked, the data comming from the file will also be filter by the importer. When unchecked, the filters only affect this view. Check documentation for more information on defining filters: &lt;href=https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering&gt;https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering&lt;/href&gt;</value>
```
msevestre, line 124: remove from resources. It should be set using the constant

msevestre: and not in the resx file. we never touch those guys

msevestre (changes requested):

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -439,8 +439,9 @@ public static string SheetsNotFound(List<string> listOfSheetNames)
             return sb.ToString();
          }
 
-         public static readonly string UseFiltersForImport = "Use the filters for importing the data";
-         
+         public static readonly string UseFiltersForImport = "Use filters for importing data";
+         public static readonly string UseFiltersForImportTooltip = "When checked, the data comming from the file will also be filter by the importer. When unchecked, the filters only affect this view. Check documentation for more information on defining filters: <href=https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering>https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering</href>";
```
Yuri05, line 443: _When checked, the data comming from the file will also be filter by the importer._
Would reformulate this part to something like:
"When checked, the filter will be applied to the data during the import process"

Also maybe it's better to use "selected" and "deselected" instead of "checked" and "unchecked"?

Yuri05 (approved):

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -439,8 +439,9 @@ public static string SheetsNotFound(List<string> listOfSheetNames)
             return sb.ToString();
          }
 
-         public static readonly string UseFiltersForImport = "Use the filters for importing the data";
-         
+         public static readonly string UseFiltersForImport = "Use filters for importing data";
+         public static readonly string UseFiltersForImportTooltip = "When selected, the filter will apply to the data during the import process. When deselected, the filter only affects this view. Check documentation for more information on defining filters: <href=https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering>https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering</href>";
```
msevestre, line 443: hm... I don;t see WHERE this constant is used in the code? Am I missing something

msevestre (changes requested):

src/OSPSuite.UI/Views/Importer/ImporterDataView.Designer.cs
```diff
@@ -85,7 +85,7 @@ private void InitializeComponent()
          this.useForImportCheckEdit.Size = new System.Drawing.Size(478, 24);
          this.useForImportCheckEdit.StyleController = this.rootLayoutControl;
          this.useForImportCheckEdit.TabIndex = 11;
-         this.useForImportCheckEdit.ToolTip = resources.GetString("useForImportCheckEdit.ToolTip");
+         this.useForImportCheckEdit.ToolTip = Captions.Importer.UseFiltersForImportTooltip;
```
abdelr, line 88: It is used here

PR #1123, opened 2021-06-22
Renaming

### LINKED ISSUE #1093: OSP 10 beta: Naming of "Load" buttons and and "Confirmation" Tab (Importer-Redesign)
In the redesigned data importer, it was confusing (not only for me) that it was not possible to reload the sheets, i.e. press again the button “Load current sheet” after adjusting some settings. 
What is actually done after clicking “Load current sheet” is that a preview is generated in the Tab currently called “Confirmation”. 
Since changes are (or should be) directly displayed in this Tab, there is no need to click the button again once the preview has been generated. We suggest thus to rename the buttons now called “Load current sheet” and “Load all sheets” (cf. Figure 1) to “Generate preview based on current sheet” and “Generate preview based on all sheets”, respectively. Accordingly, we suggest to rename the Tab “Confirmation” (cf. Figure 2) to “Import-Preview”.

Figure1
Figure2

abdelr:
Fixes #1093 (Renaming load button and confirmation tab)

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -365,7 +365,7 @@ namespace OSPSuite.Assets
          public static readonly string Mappings = "Mappings";
          public static readonly string FormatPlain = "Format";
          public static readonly string DataMapping = "Data Mapping";
-         public static readonly string Confirmation = "Confirmation";
+         public static readonly string Confirmation = "Import preview";
          public static readonly string ThreeDots = "...";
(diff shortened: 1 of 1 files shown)


abdelr:
I did not include "to selection". It is way too long and the buttons cannot hold the text while keeping the normal size (which is already a red flag) and I believe it is meaningless. Again, changing the caption is quite easy so if there is a better proposal we could still change it further.

Yuri05 (approved):

PR #1128, opened 2021-06-23
1119 configuration based on current sheet

### LINKED ISSUE #1119: Importer: It should be possible to create a configuration based on any sheet, not only the first one
Sheet one is:

Time [min] | Measurement [ng/ml] | Error [ng/ml]
-- | -- | --
144,0159882 | 5,57572792 | 1,238466311

Sheet two is:

Time [min] | Fraction [%] | Error [%]
-- | -- | --
0 | 100 | &nbsp;
13,17227 | 93,75 | &nbsp;

It is currently not possible to select columns from the second sheet for the mapping, only the columns present in the first sheet are available. It is therefore not possible to import the second sheet at all.

Proposal: it should be possible to create a mapping not only from the **first** sheet, but from the currently active sheet.

georgeDaskalakis:
closes #1119


msevestre:
Looks good to me

PR #1129, opened 2021-06-24
1090 check lloq is the same

### LINKED ISSUE #1090: LLOQ values from column are not used although configured

abdelr:
Fixes #1090 (Prompt the user before importing when different LLOQ values are present in a single data repository)

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -421,7 +421,7 @@ namespace OSPSuite.Assets
          public static readonly string NewDataStetsWillBeImported = "New datasets that will be imported";
          public static readonly string ReloadData = "Reload Data";
          public static readonly string SeparatorSelection = "Separator Selection";
-         public static string LLOQInconsistentValuesAt(string dataRepositoryName) => $"There were different LLOQ values detected for data repository under name {dataRepositoryName}";
+         public static string LLOQInconsistentValuesAt(string dataRepositoryName) => $"There were different LLOQ values detected for data repository under name {dataRepositoryName}, are you sure you want to continue with import?";
          public static string CsvSeparatorDescription(string fileName) => $"Please select the separator for the file \r\n'{fileName}':";
 
          public static readonly string SheetFormatNotSupported = "The format of the sheet you are trying to use is not supported.You can find a documentation of the supported formats<href =https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/import-edit-observed-data#supported-formats > here </href>";
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -421,7 +421,7 @@ namespace OSPSuite.Assets
          public static readonly string NewDataStetsWillBeImported = "New datasets that will be imported";
          public static readonly string ReloadData = "Reload Data";
          public static readonly string SeparatorSelection = "Separator Selection";
-         public static string LLOQInconsistentValuesAt(string dataRepositoryName) => $"There were different LLOQ values detected for data repository under name {dataRepositoryName}, are you sure you want to continue with import?";
+         public static string LLOQInconsistentValuesAt(string dataRepositoryName) => $"There were different LLOQ values detected for the data from a single source. Please check data under name {dataRepositoryName}. Are you sure you want to continue with import?";
          public static string CsvSeparatorDescription(string fileName) => $"Please select the separator for the file \r\n'{fileName}':";
 
          public static readonly string SheetFormatNotSupported = "The format of the sheet you are trying to use is not supported.You can find a documentation of the supported formats<href =https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/import-edit-observed-data#supported-formats > here </href>";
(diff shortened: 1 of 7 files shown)


src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
```diff
@@ -34,9 +49,10 @@ public DataRepository ConvertImportDataSet(ImportedDataSet dataSet)
             dataRepository.ExtendedProperties.Add(new ExtendedProperty<string>() { Name = metaDataDescription.Name, Value = metaDataDescription.Value });
          }
 
+         var warningFlag = false;
          foreach (var column in dataSet.ParsedDataSet.Data)
          {
-            convertParsedDataColumn(dataRepository, column, dataSet.FileName);
+            warningFlag |= convertParsedDataColumnAndReturnWarningFlag(dataRepository, column, dataSet.FileName);
```
msevestre, line 55: Isn't this a case for a any?
```
var warningFlag = dataSet.ParsedDataSet.Data.any(x=>convertParsedDataColumnAndReturnWarningFlag(dataRepository, x, dataSet.FileName))
```

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
```diff
@@ -99,7 +116,12 @@ private void convertParsedDataColumn(DataRepository dataRepository, KeyValuePair
          foreach (var value in columnAndData.Value)
          {
             if (!double.IsNaN(value.Lloq))
-               lloqValue = value;
+            {
+               if (lloqValue != null && !ValueComparer.AreValuesEqual(lloqValue.Lloq, value.Lloq))
+                  warningFlag = true;
```
msevestre, line 121: return true right away? why do you stil need to iterate?

msevestre (changes requested):
Looking very good. Just some small changes maybe

src/OSPSuite.Presentation/Services/DataImporter.cs
```diff
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Xml.Linq;
```
msevestre, line 4: remove

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
abdelr: Yes, it was already on a foreach and I did not really realized it... thanks!!

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
abdelr: We need to iterate still because we want to give the user the opportunity to import even when there are inconsistencies so, we still need to populate the DataRepositories and show the user the option. We could have two functions, and then it would be as you propose but it would mean parse the data twice which is not very performant so if we agree this is the less frequent case, I believe it is best to keep iterating even when there is problems in case the user wants to import the data anyways instead of parsing the data twice always (one for checking and one for parsing)

msevestre: ok I see

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
abdelr: Actually, same as below here. If I use any, it will return as soon as it finds some true. Any is not intended for changing the data. Normally foreach should not translate to any. Don't you agree?

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -421,6 +421,7 @@ public static class Importer
          public static readonly string NewDataStetsWillBeImported = "New datasets that will be imported";
          public static readonly string ReloadData = "Reload Data";
          public static readonly string SeparatorSelection = "Separator Selection";
+         public static string LLOQInconsistentValuesAt(string dataRepositoryName) => $"There were different LLOQ values detected for data repository under name {dataRepositoryName}, are you sure you want to continue with import?";
```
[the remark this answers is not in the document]
abdelr: I agree. Thanks. I rephrase it a bit but I am not sure if I found the best sentences. Please, let me know any suggestion you would have.

src/OSPSuite.Presentation/Services/DataImporter.cs
abdelr: Done!

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
msevestre: yeah. The problem is that your method has SIDE EFFECTS and is not void. That's why it through me off.
I am ok with this for now because we are running out of time
But side effects method should not return values. If we make our functions as pure as possible, there is no surprises as to what the code does and one does not need to read the details of each function

msevestre (approved):

PR #1135, opened 2021-06-25
Add meta data

### LINKED ISSUE #1134: Observed data import Redesign: add Subject ID as default mapping name
@AndreDlm commented on [Wed Jun 16 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1792)

Subject Id should be added as default mapping name here:

image

---

@PavelBal commented on [Fri Jun 25 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1792#issuecomment-868354511)

- Add "Patient Id" as meta data.

---

@Yuri05 commented on [Fri Jun 25 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1792#issuecomment-868387316)

>Add "Patient Id" as meta data.

**Subject Id** is a generalization of "Patient Id" and should replace it.

abdelr:
Fixes #1134 (Adding Subject Id to the list of meta data)

### DIFF
--- a/src/OSPSuite.Core/Domain/Constants.cs
+++ b/src/OSPSuite.Core/Domain/Constants.cs
@@ -451,6 +451,7 @@ namespace OSPSuite.Core.Domain
          public static readonly string MOLECULARWEIGHT = "Molecular Weight";
          public static readonly string SPECIES = "Species";
          public static readonly string STUDY_ID = "Study Id";
+         public static readonly string SUBJECT_ID = "Subject Id";
          public static readonly string GROUP_ID = "Group Id";
          public static readonly string ROUTE = "Route";
          public static readonly string PATIENT_ID = "Patient Id";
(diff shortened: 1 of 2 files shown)


src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -451,6 +451,7 @@ public static class ObservedData
          public static readonly string MOLECULARWEIGHT = "Molecular Weight";
          public static readonly string SPECIES = "Species";
          public static readonly string STUDY_ID = "Study Id";
+         public static readonly string SUBJECT_ID = "Subject Id";
          public static readonly string GROUP_ID = "Group Id";
          public static readonly string ROUTE = "Route";
```
msevestre, line 456: Should we remove this entry then?

msevestre (changes requested):

src/OSPSuite.Core/Domain/Constants.cs
abdelr: Why? @AndreDlm requested "Subject Id" to be added. @PavelBal said should not be "Subject Id" but "Patient Id" and @Yuri05 replied "Subject Id" is a generalization of "Patient Id" so it should be as in code now, right?

msevestre: Exactly. We do not need patientid and subject id because one is the generalization of the other. Don't you think?

abdelr:
It was on constants... sure, I was looking on the wrong file on the code... ;)

msevestre (approved):

PR #1143, opened 2021-06-29
Fixes #980 Expose C++ export for R

### LINKED ISSUE #980: Expose C++ export for R

### LINKED ISSUE #1113: Update to latest SimModel

Yuri05:
Fixes #1113 Update to latest SimModel

### DIFF
--- a/src/OSPSuite.R/Domain/SimulationBatch.cs
+++ b/src/OSPSuite.R/Domain/SimulationBatch.cs
@@ -6,6 +6,7 @@ using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Data;
 using OSPSuite.Core.Domain.Services;
 using OSPSuite.R.Extensions;
+using OSPSuite.SimModel;
 using OSPSuite.Utility.Extensions;
 
 namespace OSPSuite.R.Domain
@@ -73,6 +74,11 @@ namespace OSPSuite.R.Domain
          //This needs to be done after initialization of the SimModelBatch so that we can check parameters
          validate(simulationBatchOptions);
          _simulationPersistableUpdater.UpdateSimulationPersistable(simulation);
+      }
+
+      public void ExportToCPPCode(string outputFolder)
+      {
+         _simModelBatch.ExportToCPPCode(outputFolder, CodeExportMode.Values);
       }
 
       private void validate(SimulationBatchOptions simulationBatchOptions)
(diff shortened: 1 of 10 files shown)


src/OSPSuite.R/Domain/SimulationBatch.cs
```diff
@@ -69,12 +70,37 @@ ISimulationPersistableUpdater simulationPersistableUpdater
...
+      /// <param name="modelName">
+      ///   If empty (default): model will be named Standard and exported to Standard.cpp
+      ///   Otherwise: model will be named to "modelName" in the C++ code and exported to modelName.cpp.
+      ///              modelName must be both valid file name AND valid C++ identifier in such a case
+      /// </param>
+      public void ExportToCPPCode(string outputFolder, bool fullMode, string modelName="")
+      {
+         var exportMode = fullMode ? CodeExportMode.Formula : CodeExportMode.Values;
+         _simModelBatch.ExportToCPPCode(outputFolder, exportMode, modelName);
+      }
```
Yuri05, line 102: @PavelBal @msevestre **This** is the function which can be called from R to export the model (and variable parameters/initial values as they are set in the SimulationBatch) to C++.

With the current implementation of the SimulationBatch exporting in **Formula** and in **Value** mode would produce the same result I think (which corresponds to the Value mode) - which is ok anyway I think.

Exporting in **Formula** mode would make sense if e.g.
 - you would like to read the exported C++ code for whatever reason (I can hardly imagine any) or
 - if at the model creation time you don't know which parameters you will vary later and want to set ALL parameters as variable

PavelBal: @Yuri05 Could you please explain what 
` valid C++ identifier in such a case` is (sorry if it's a trivial question).

>if at the model creation time you don't know which parameters you will vary later and want to set ALL parameters as variable

Setting all parameters as variable has a negative impact on the performance?

Yuri05: >Could you please explain what

https://www.krellinst.org/archive/ais/textbook/unit7/CPP7.3.2.html

>Setting all parameters as variable has a negative impact on the performance?

Yes.

PavelBal (approved):

Yuri05:
@msevestre Any concerns to merge this PR?

msevestre:
Nope...i don't understand half of it but ...

PR #1146, opened 2021-06-30
dirty quick solution

### LINKED ISSUE #1145: After import: observed data may not be modified anymore (except for the meta data)
@Yuri05 commented on [Tue Jun 29 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1820)

It should not be possible to edit observed **data** in PK-Sim (**meta data** can be edited as before)
If data needs to be changed: the new "Update" functionality must be used instead.

grafik

abdelr:
Fixes #1145 (Not possible to edit DataRepository after import)

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/Importer/ImportConfirmationPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Importer/ImportConfirmationPresenter.cs
@@ -26,7 +26,6 @@ namespace OSPSuite.Presentation.Presenters.Importer
          _dataPresenter = dataPresenter;
          View.AddChartView(_chartPresenter.BaseView);
          View.AddDataView(_dataPresenter.View);
-         _dataPresenter.DisableEdition();
       }
       public void DataSetSelected(int index)
       {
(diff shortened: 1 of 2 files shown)


abdelr:
Please, do notice this is not the definite solution. I think we would like to release V10 soon so this will solve the issue but the code is not in a good shape now. So BaseDataRepositoryData(View/Presenter) have functionality for editing that is never used anymore. I believe we should go there and remove all this functionality from the very beginning instead of adding the functionality just to remove it later. Still, the BaseDataRepositoryDataView is the base class for WeightedDataRepositoryDataView and I guess there we could need the edition abilities. So either we change the inheritance on classes if nothing else is shared, we keep it the way it is... We have to take a deeper look into this so I would suggest not to close the issue now but just to merge the PR to have the feature requested by @Yuri05 ready for V10.

Yuri05:
as we decided not to implement this in V10 I think we can close this PR?

abdelr:
> as we decided not to implement this in V10 I think we can close this PR?

Sure

abdelr:
Closing since we decided not to include

PR #1152, opened 2021-07-01
Keeping track of the search tree to avoid infinite recursion

### LINKED ISSUE #1151: Crash when calculating PK-parameters for molecules that are educts and products of each other.
@PavelBal commented on [Thu Jul 01 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1821)

If molecule A is metabolized to molecule B, and molecule B is metabolized to molecule A, simulating the model and selecting "Show PK Parameters" crashes PKSim.

SUch a model structure leads to an infinite loop in 

https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/2ebf3f549551f58feb7fef0fbfeea8b5c5d133a9/src/OSPSuite.Core/Domain/ModelCoreSimulationExtensions.cs#L45

Minimal example attached.

[AdministrationLoop.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/6748464/AdministrationLoop.zip)

---

@msevestre commented on [Thu Jul 01 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1821#issuecomment-872309064)

Nice....

abdelr:
Fixes #1151 (Breaking infinite recursion when requesting PKParameters with educts and products from each other)

### DIFF
--- a/src/OSPSuite.Core/Domain/ModelCoreSimulationExtensions.cs
+++ b/src/OSPSuite.Core/Domain/ModelCoreSimulationExtensions.cs
@@ -39,11 +39,11 @@ namespace OSPSuite.Core.Domain
          var allApplications = applicationEventGroup.SelectMany(x => x.GetAllChildren<IContainer>(c => c.ContainerType == ContainerType.Application))
             .ToList();
 
-         return getApplicationsForAppliedAncestorMolecule(simulation.Reactions, moleculeName, allApplications);
+         return getApplicationsForAppliedAncestorMolecule(simulation.Reactions, moleculeName, allApplications, new List<string>());
       }
 
       private static IReadOnlyList<IContainer> getApplicationsForAppliedAncestorMolecule(IEnumerable<IReactionBuilder> reactions, string moleculeName,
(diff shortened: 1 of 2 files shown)


abdelr:
I am still working on adding tests, but please look if this solution is good enough. This is a completely new area for me ;)

PavelBal (approved):

msevestre:
Can't really tell until I see a test;-) make it red, fix it, make it green boom

abdelr:
It is not really possible to catch an StackOverflowException as far as I know. So the test will not really be red before solving the issue, but will not turn green either. If anyone has any suggestion please, let me know.

msevestre:
Please remove all the fake stuff and create real Reaction, Container etc..
Then I 'll download the code and see what you mean with StackOverflow. I am pretty sure that if the tests does not run, (because of a SO) the test is NOT Green (red or inclusive) which is what we need

msevestre:
YES!! WELL DONE BROTHER

PR #1154, opened 2021-07-01
Fixes #1153 throw if not found

### LINKED ISSUE #1153: extend setParameterValuesByPath function
@Yuri05 commented on [Thu Jul 01 2021](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/541)

Extend the function `setParameterValuesByPath` with an optional argument `ignoreNonExisting ` (or something like that).
The new argument should be **false** per default. 
If set to **true**: `setParameterValuesByPath `should skip non existing parameters with a warning and not exit with error.

Such a behaviour might be required in situations like [here](https://github.com/Open-Systems-Pharmacology/OSPSuite.ReportingEngine/issues/497)

---

@msevestre commented on [Thu Jul 01 2021](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/541#issuecomment-872520413)

going ot use the nomenclature , stopIfNotFound = TRUE, as this is what we have for other methods

msevestre:


src/OSPSuite.R/Services/ContainerTask.cs
```diff
@@ -159,17 +164,23 @@ public IParameter[] AllParametersForSensitivityAnalysisMatching(ISimulation simu
...
          var pathArray = path.ToPathArray();
          var quantity = simulation.Model.Root.EntityAt<IQuantity>(pathArray);
-         if (quantity == null)
+         if (quantity != null)
+         {
+            quantity.Value = value;
+            return;
+         }
+
+         if (throwIfNotFound)
```
msevestre, line 180: @Yuri05 Required for the chnages in R,. Merging

PR #1161, opened 2021-07-07
Adding sensitivities when the CalculateSensitivity switch is on

### LINKED ISSUE #1160: Return Jacobi matrix
@PavelBal commented on [Wed Dec 02 2020](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/453)

Is there a way to return the Jacobi matrix of the simulation?

---

@msevestre commented on [Wed Dec 02 2020](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/453#issuecomment-737346832)

no

---

@PavelBal commented on [Wed Dec 02 2020](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/453#issuecomment-737353902)

Why not resp. where should this be done? Jacobi is calculated during simulations (I saw a respective method somewhere in SimModel) und is used by the internal PI, so it must be there..

---

@msevestre commented on [Wed Dec 02 2020](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/453#issuecomment-737395201)

@PavelBal you ask if there was a way to return the Jacobian matrix at the moment and no there is not.

This is something that could potentially be added in the future but we need to specify it.
for PI, we calculate partial derivatives for the parameters varied in the PI and not for all parameters obviously

I am not sure what Jacobian matrix is calculated when we run a simulation. @Yuri05 ?

---

@Yuri05 commented on [Thu Dec 03 2020](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/453#issuecomment-737894401)

@PavelBal What are you interested in? `dStateVariable_i(t)/dParameter_j`? 

---

@msevestre commented on [Thu Dec 03 2020](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/453#issuecomment-737902239)

@Yuri05 what Jacobian is calculated when we run a simulation? 

---

@Huan-Yang commented on [Fri Apr 23 2021](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/453#issuecomment-825688660)

Hello @Yuri05 @msevestre, indeed for PI, I wonder whether we can have dStateVariable_i(t)/dParameter_j from `ospsuite::runSimulation` or alike, as I presumed that PKSim utilizes CVODES of SUNDIALS to integrate the ODEs and to calculate associated sensitivities. 

---

@Yuri05 commented on [Sat Apr 24 2021](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/453#issuecomment-826079629)

>as I presumed that PKSim utilizes CVODES of SUNDIALS to integrate the ODEs and to calculate associated sensitivities.

Correct.
>indeed for PI, I wonder whether we can have dStateVariable_i(t)/dParameter_j from ospsuite::runSimulation or alike

You can :) At the moment OSPSuite.Core allows to 1) calculate sensitivities for all parameters which were set to "ToBeVaried" and 2) get output sensitivity values:

Thus we would just need to provide the CalcSensi-option and extend returned values with calculated sensitivities: 

* OSPSuite.Core: extend [OSPSuite.R/SimulationBatchOptions](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/483b2e3670f4cd670cc4ca462d9bb98999e8fff6/src/OSPSuite.R/Domain/SimulationBatch.cs#L13) with `CalculateSensitivity`-Flag and pass it to the [initialization of SimModelBatch](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/483b2e3670f4cd670cc4ca462d9bb98999e8fff6/src/OSPSuite.R/Domain/SimulationBatch.cs#L72)
* OSPSuite.Core: extend the [OSPSuite.R/SimulationBatch](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/483b2e3670f4cd670cc4ca462d9bb98999e8fff6/src/OSPSuite.R/Domain/SimulationBatch.cs#L54) with a call to [SimModelBatch.SensitivityValuesFor(string outputPath, string parameterPath)](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/483b2e3670f4cd670cc4ca462d9bb98999e8fff6/src/OSPSuite.Core/Domain/Services/SimModelBatch.cs#L125)
* wrap both extensions in the ospsuite-R package

abdelr:
Fixes #1160 (Return Jacobi matrix when runSimulation)

### DIFF
--- a/src/OSPSuite.Core/Domain/Data/QuantityValues.cs
+++ b/src/OSPSuite.Core/Domain/Data/QuantityValues.cs
@@ -20,7 +20,7 @@ namespace OSPSuite.Core.Domain.Data
 
       private float[] _values;
 
-      private double[] _sensitivities;
+      private IReadOnlyDictionary<string, double[]> _sensitivities = new Dictionary<string, double[]>();
 
       /// <summary>
       ///    Only required because charts requires ColumnId to defined curves
@@ -49,10 +49,10 @@ namespace OSPSuite.Core.Domain.Data
          set => _values = value ?? new float[] {};
       }
 
-      public virtual double[] Sensitivities
+      public virtual IReadOnlyDictionary<string, double[]> Sensitivities
       {
          get => _sensitivities;
-         set => _sensitivities = value ?? new double[] {};
+         set => _sensitivities = value ?? new Dictionary<string, double[]>();
       }
 
       public virtual IReadOnlyList<string> PathList
--- a/src/OSPSuite.Core/Domain/Data/QuantityValues.cs
+++ b/src/OSPSuite.Core/Domain/Data/QuantityValues.cs
@@ -20,6 +20,8 @@ namespace OSPSuite.Core.Domain.Data
 
       private float[] _values;
 
+      private double[] _sensitivities;
+
       /// <summary>
       ///    Only required because charts requires ColumnId to defined curves
       /// </summary>
@@ -45,6 +47,12 @@ namespace OSPSuite.Core.Domain.Data
       {
          get => _values;
          set => _values = value ?? new float[] {};
+      }
+
+      public virtual double[] Sensitivities
+      {
+         get => _sensitivities;
+         set => _sensitivities = value ?? new double[] {};
       }
 
       public virtual IReadOnlyList<string> PathList
--- a/src/OSPSuite.Core/Domain/Services/SimulationResultsCreator.cs
+++ b/src/OSPSuite.Core/Domain/Services/SimulationResultsCreator.cs
@@ -8,23 +8,35 @@ namespace OSPSuite.Core.Domain.Services
    public interface ISimulationResultsCreator
    {
       SimulationResults CreateResultsFrom(DataRepository dataRepository);
+
+      SimulationResults CreateResultsWithSensitivitiesFrom(DataRepository dataRepository, ISimModelBatch simModel, string[] parameters);
    }
 
    public class SimulationResultsCreator : ISimulationResultsCreator
    {
       public SimulationResults CreateResultsFrom(DataRepository dataRepository)
       {
+         return createResultsFrom(dataRepository, null, new string[] { });
+      }
+
+      public SimulationResults CreateResultsWithSensitivitiesFrom(DataRepository dataRepository, ISimModelBatch simModel, string[] parameters)
+      {
+         return createResultsFrom(dataRepository, simModel, parameters);
+      }
+
+      private SimulationResults createResultsFrom(DataRepository dataRepository, ISimModelBatch simModel, string[] parameters)
+      {
          if (dataRepository.IsNull() || dataRepository.BaseGrid == null)
             return new NullSimulationResults();
 
-         var results = new SimulationResults {Time = valuesFrom(dataRepository.BaseGrid)};
-         
+         var results = new SimulationResults { Time = valuesFrom(dataRepository.BaseGrid, null, new string[] { }) };
+
          foreach (var columnsForIndividual in dataRepository.AllButBaseGrid().GroupBy(selectIndividualIndex))
          {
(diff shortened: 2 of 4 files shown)


src/OSPSuite.Core/Domain/Data/QuantityValues.cs
```diff
@@ -47,6 +49,12 @@ public virtual float[] Values
          set => _values = value ?? new float[] {};
       }
 
+      public virtual double[] Sensitivities
+      {
+         get => _sensitivities;
+         set => _sensitivities = value ?? new double[] {};
```
msevestre, line 55: Why would this be set to null?

src/OSPSuite.R/Domain/SimulationBatch.cs
```diff
@@ -28,6 +28,9 @@ public class SimulationBatchOptions
       public string[] Parameters => VariableParameters.ToNetArray(VariableParameter);
 
       public string[] Molecules => VariableMolecules.ToNetArray(VariableMolecule);
+
+      //Defaults to null. If true, it will trigger the calculation of sensitivity
```
msevestre, line 32: Default to false maybe? but not null

src/OSPSuite.R/Domain/SimulationBatch.cs
```diff
@@ -128,7 +132,12 @@ public Task<SimulationResults> RunAsync(SimulationBatchRunValues simulationBatch
             _simModelBatch.UpdateParameterValues(simulationBatchRunValues.Values);
             _simModelBatch.UpdateInitialValues(simulationBatchRunValues.MoleculeValues);
             var simulationResults = _simModelBatch.RunSimulation();
-            return _simulationResultsCreator.CreateResultsFrom(simulationResults.Results);
+            var results = _simulationResultsCreator.CreateResultsFrom(simulationResults.Results);
+            if (_simulationBatchOptions.CalculateSensitivity)
```
msevestre, line 136: Please rewrite this code. It does not follow our coding standard at all. 
First, return early, then use {} because WHY would you put one if and two foreach without them?

Also why using first. How do you know there is one? 

Last. Why isn't the result prefilled with this stuff already. .TELL Don't ask

msevestre (changes requested):
Where does this requirement comes from in the first place?

src/OSPSuite.Core/Domain/Data/QuantityValues.cs
msevestre: nevermnd. This would be good for serialization so that we do not serialize empty entries

PavelBal:
> Where does this requirement comes from in the first place?

Required for some parameter identification algorithms.

src/OSPSuite.R/Domain/SimulationBatch.cs
abdelr: Actually, this is still a draft... I had to move to something else and did not want to leave things uncommitted. Now I have fixed some things (including the fact that the sensitivities were not properly calculated) and should be better ;)

src/OSPSuite.Core/Domain/Services/SimulationResultsCreator.cs
```diff
@@ -8,23 +8,35 @@ namespace OSPSuite.Core.Domain.Services
    public interface ISimulationResultsCreator
    {
       SimulationResults CreateResultsFrom(DataRepository dataRepository);
+
+      SimulationResults CreateResultsWithSensitivitiesFrom(DataRepository dataRepository, ISimModelBatch simModel, string[] parameters);
```
Yuri05, line 12: let's rename the argument `simModel `to `simModelBatch`

Yuri05 (approved):

PR #1164, opened 2021-07-08
Moving user interaction to prompt for the file path outside the service

### LINKED ISSUE #1163: Moving user interaction to prompt for the file path outside the service

abdelr:
Fixes #1163 (Moving user interaction to prompt for the file path outside the service)

### DIFF
--- a/src/OSPSuite.Presentation/Services/DataImporter.cs
+++ b/src/OSPSuite.Presentation/Services/DataImporter.cs
@@ -87,17 +87,16 @@ namespace OSPSuite.Presentation.Services
       public (IReadOnlyList<DataRepository> DataRepositories, ImporterConfiguration Configuration) ImportDataSets(
          IReadOnlyList<MetaDataCategory> metaDataCategories,
          IReadOnlyList<ColumnInfo> columnInfos,
-         DataImporterSettings dataImporterSettings
+         DataImporterSettings dataImporterSettings,
+         string dataFileName
       )
       {
-         var path = _dialogCreator.AskForFileToOpen(PleaseSelectDataFile, ImportFileFilter, Constants.DirectoryKey.OBSERVED_DATA);
-
-         if (string.IsNullOrEmpty(path))
+         if (string.IsNullOrEmpty(dataFileName))
             return (Array.Empty<DataRepository>(), null);
 
          using (var importerModalPresenter = _applicationController.Start<IModalImporterPresenter>())
          {
-            return importerModalPresenter.ImportDataSets(metaDataCategories, columnInfos, dataImporterSettings, path);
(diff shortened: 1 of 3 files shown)


abdelr:
@Yuri05 We realized we had this call to the DialogCreator inside the service when it should actually happen on the caller. Now that we are writing tests and the service also on the R side, it was quite clear a problem. Could you please merge the changes if you find them OK so we could move on with the other changes depending on this?

src/OSPSuite.Presentation/Services/DataImporter.cs
```diff
@@ -87,43 +87,41 @@ private static MetaDataCategory createMetaDataCategory<T>(string descriptiveName
...
          IReadOnlyList<ColumnInfo> columnInfos,
-         DataImporterSettings dataImporterSettings
+         DataImporterSettings dataImporterSettings,
+         string dataFileName
       )
       {
-         var path = _dialogCreator.AskForFileToOpen(PleaseSelectDataFile, ImportFileFilter, Constants.DirectoryKey.OBSERVED_DATA);
-
-         if (string.IsNullOrEmpty(path))
+         if (string.IsNullOrEmpty(dataFileName))
```
Yuri05, line 94: Now when the file selection is moved **outside** I would expect that the function is always called with the **valid data file name**.Thus 2 questions/remarks:
* if the `dataFileName` is null/empty - I would throw an exception
* I would also check if the `dataFileName` exists - if not, throw an exception

If throwing exceptions is not appropriated here (because the caller expects different behaviour) then at least the check for `dataFileName` exists should be done (returning the empty data repo array)

src/OSPSuite.Presentation/Services/DataImporter.cs
```diff
@@ -87,43 +87,41 @@ private static MetaDataCategory createMetaDataCategory<T>(string descriptiveName
...
          IReadOnlyList<ColumnInfo> columnInfos,
-         DataImporterSettings dataImporterSettings
+         DataImporterSettings dataImporterSettings,
+         string dataFileName
       )
       {
-         var fileName = _dialogCreator.AskForFileToOpen(OpenFile, ImportFileFilter, Constants.DirectoryKey.OBSERVED_DATA);
-         
-         if (string.IsNullOrEmpty(fileName))
+         if (string.IsNullOrEmpty(dataFileName))
```
Yuri05, line 111: same as above

abdelr: Good catch, thanks! Indeed the expected behavior is to return an empty list when no valid path (this would be the case for example if the open dialog is to be cancelled).

Yuri05 (approved):

PR #1165, opened 2021-07-08
1162 importer from r

### LINKED ISSUE #1162: The importer should be available from R
The importer should be available from R. Let us first add the simplest use case: import data using a saved configuration (so, still not possible to edit the configuration)

abdelr:
Fixes #1162 (importer from r)


abdelr:
@Yuri05 and @PavelBal: There are some issues pending:
#1166 
#1167
Also, it might make sense to add features such as editing configuration from R. More than that, this PR might be interesting on V10?

abdelr:
> More than that, this PR might be interesting on V10?

I would suggest not. It should not affect PKSim or MoBi but we need to test this well before going public I think.

PavelBal:
> More than that, this PR might be interesting on V10?

No, I think we can release a separate version of ospsuite-r once we implemented a use case.

src/OSPSuite.R/Services/CsvSeparatorSelector.cs
```diff
@@ -0,0 +1,12 @@
...
+namespace OSPSuite.R.Services
+{
+   public class CsvSeparatorSelector : ICsvSeparatorSelector
+   {
+      public char? GetCsvSeparator(string fileName)
+      {
+         return ',';
+      }
+   }
+}
```
Yuri05, line 12: don't understand the purpose of this function. `fileName `is not used, it's declared as `char?` but always returns the same constant value - and why is it needed at all in R?

src/OSPSuite.R/Services/DataImporter.cs
```diff
@@ -0,0 +1,198 @@
...
+         string dataFileName)
+      {
+         if (string.IsNullOrEmpty(dataFileName))
+            return Enumerable.Empty<DataRepository>().ToList();
+
+         try
+         {
+            var importedData = _importer.ImportFromConfiguration(configuration, columnInfos, dataFileName, metaDataCategories, dataImporterSettings);
+            if (importedData.MissingSheets.Count != 0)
+               _dialogCreator.MessageBoxError(Captions.Importer.SheetsNotFound(importedData.MissingSheets));
```
Yuri05, line 122: why does the function show error message boxes instead of throwing exceptions?
I would expect that the function is called in silent mode in R, e.g. later as part of a qualification workflow (which later on will be based on **referenced** observed data)

src/OSPSuite.R/Services/DataImporter.cs
```diff
@@ -0,0 +1,198 @@
...
+            return importedData.DataRepositories.Select(drm => drm.DataRepository).ToList();
+         }
+         catch (Exception e) when (e is UnsupportedFormatException || e is UnsupportedFileTypeException)
+         {
+            _dialogCreator.MessageBoxError(e.Message);
```
Yuri05, line 127: same

src/OSPSuite.R/Services/CsvSeparatorSelector.cs
abdelr: > don't understand the purpose of this function. `fileName `is not used, it's declared as `char?` but always returns the same constant value - and why is it needed at all in R?

This is an interface that already existed. In PKSim, for example, it will show the first few lines to the user and ask, what separator would you like to use so you can select from a list. On R, there should be a way to do something similar (selecting the separator) but as explained before, I created an issue (#1166) for this, rather than tackling the problem now since I want to discuss with Pavel or other R users how this is more convenient for them. This is the reason why it is on a different issue. After implementing this issue, the function should now retrieve the same char always, it should be editable from settings or by using some setters but for now, I did not want to focus on this but rather I would like to keep it on a different issue. If you would prefer to have everything together, well, we also could. The param (fileName) is only there because of the existing interface, so from R no use, but from PKSim/MoBi it is indeed used.

src/OSPSuite.R/Services/DataImporter.cs
abdelr: In this case, I would not like to throw an exception. This means that some sheets were not found on the configuration but it does not mean it was not able to load anything. 

This will even be pointless if we completely remove sheets from configuration, but then we will need to define what sheets to load.

src/OSPSuite.R/Services/DataImporter.cs
abdelr: Here, we do need to show a feedback to the user somehow. Is your suggestion to let it break and show the exception from the R side? In general I could just leave the exceptions unhandled and that would do the work, right?

Yuri05: >In general I could just leave the exceptions unhandled and that would do the work, right?

yes

src/OSPSuite.R/Services/DataImporter.cs
Yuri05: ok, let's then just remove MsgBox for now. In principle it's a warning which should be logged somehow; for now we can just ignore it

src/OSPSuite.R/Services/CsvSeparatorSelector.cs
Yuri05: Why do we need this `CsvSeparatorSelector `in R at all?

abdelr: ICsvSeparatorSelector is a dependencie of CsvDataSourceFile (This is a class on the core). Normally, the ICsvSeparatorSelector is intended to handle user interaction (selection of the separator). The interface is on the core but the concrete implementation for PKSim or MoBi is on UI. This implementation is not available on R and either we decide to change this on the future or we have a different implementation for R. In any case, we need to be able to ask the user what separator does he want to use for a given csv file.

Yuri05: >In any case, we need to be able to ask the user what separator does he want to use for a given csv file.

Why? In R I would expect that separaor is just set via scripting and not via user prompt.

abdelr: Exactly, we will have a wrapper on the R side. This wrapper should prepare the R implementation of ICsvSeparatorSelector to return whatever the script passed. This should be implemented on the other issue but for now, just to have something to start with, I am always using ','. Later, this implementation should be flexible. We could discuss it on a call if you prefer.

src/OSPSuite.R/Services/DataImporter.cs
msevestre: There is a logger implementation in R. This should be logged
See all other services in R where a logger is being used. You can then log a warning or an error

src/OSPSuite.R/Services/DataImporter.cs
msevestre: In general, _dialogCreator is not necessarily showing a message box. It could simply be showing the error to the console if there is an implementation for this. In any case, throwing exception might NOT be what we want here depending on how it's handled. We do not want to see a stack trace in R so maybe using the logger would be better (catch, log, return null or empty or whatever )

abdelr:
Please, do not merge. I think much more PR will come and we should keep this branch until the implementation is ready so we could merge new issues into this branch to see incrementally.

src/OSPSuite.R/Services/CsvSeparatorSelector.cs
abdelr: Some changes that might explain a bit on #1170

Yuri05: yes, with the last commit in #1170 it becomes much more clear. Thx!

abdelr:
> Please, do not merge. I think much more PR will come and we should keep this branch until the implementation is ready so we could merge new issues into this branch to see incrementally.

#1171

abdelr:
Seems like a problem with the decimal separator... I will research it a bit

abdelr:
> Seems like a problem with the decimal separator... I will research it a bit

Issue created at #1173

Yuri05:
@abdelr Can you please set this PR to draft to prevent us from non-intended merging until it's not fully implemented?

Yuri05 (approved):

Yuri05:
@abdelr Good to merge into develop now?

abdelr:
Sure

PR #1172, opened 2021-07-12
1171 ignoring sheet name when import from r

### LINKED ISSUE #1171: When importing from R, should be possible to ignore sheet names from configuration
Currently, the importer uses the imported sheets from the configuration to decide what data to read. We see the use case where you would have several sheets on an excel file and only want to load part of it. We also see cases where the configuration created for one file should be reused on a different file with different sheets. It should be possible at least to ignore the sheets when loading the data using the script parameters. We should later decide on PKSim and MoBi how to include this but for now, R should support this feature.

abdelr:
Fixes #1171 (Ignores sheet name when importing from R)


src/OSPSuite.Infrastructure.Import/Core/DataImporterSettings.cs
```diff
@@ -22,6 +22,8 @@ public class DataImporterSettings : IWithIcon
 
       public bool PromptForConfirmation { get; set; } = false;
 
+      public bool IgnoreSheetNamesAtImport { get; set; }
```
Yuri05, line 25: No default?

src/OSPSuite.Infrastructure.Import/Services/Importer.cs
```diff
@@ -281,7 +281,9 @@ DataImporterSettings dataImporterSettings
          dataSource.SetNamingConvention(configuration.NamingConventions);
          var sheets = new Cache<string, DataSheet>();
          var missingSheets = new List<string>();
-         foreach (var key in configuration.LoadedSheets)
+         var sheetList = dataImporterSettings.IgnoreSheetNamesAtImport ? dataSourceFile.DataSheets.Keys : configuration.LoadedSheets;
+
```
Yuri05, line 285: what will happen if `IgnoreSheetNamesAtImport = true ` and the file contains e.g. empty sheets (or sheets which do not match the configuration)?

src/OSPSuite.Infrastructure.Import/Core/DataImporterSettings.cs
abdelr: Well, default for bool is false... But you are right, I will add it explicitly.

src/OSPSuite.Infrastructure.Import/Services/Importer.cs
abdelr: In such a case, I expect exceptions on line 297 (AddSheets) but this is how it should be right? we let the exception raise and catch on the R side. We still need to discuss how to catch

Yuri05 (approved):

PR #1198, opened 2021-07-20
Adding get and create error column

### LINKED ISSUE #1191: getErrorColumn for DataColumn for R
Return an error column associated with given DataColumn. If no error column exists, create an empty one.

```
DataColumn getErrorColumn(DataColumn column){
	if (column.ContainsRelatedColumn(AuxiliaryType.ArithmeticStdDev)){
		return (column.GetRelatedColumn(AuxiliaryType.ArithmeticStdDev))
	}
	if (column.ContainsRelatedColumn(AuxiliaryType.GeometricStdDev)){
		return (column.GetRelatedColumn(AuxiliaryType.GeometricStdDev))
	}
        return(NULL)
}
```

```
DataColumn addErrorColumn(DataColumn column, DataRepository repository, String name = "yError", String errorType = AuxiliaryType.ArithmeticStdDev){
	errorColumn = DataColumn("yErrorValues", column.getDimension(), column.getBaseGrid())
	errorColum.DisplayUnit = column.DisplayUnit
	#Set auxiliaryType of DataInfo to AuxiliaryType.ArithmeticStdDev
	#Set Origin of DataInfo to ColumnOrigins.ObservationAuxiliary
	#add errorColumn as related column to `column`
	return(errorColumnn)
}
```

- Check whether DataRepository is required

abdelr:
Fixes #1191 (Adding get and create error column)

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -1431,6 +1431,7 @@ namespace OSPSuite.Assets
       public static readonly string TransportMoleculeNamesNotSet = "Transport molecule names object passed is not set";
       public static readonly string TransportMoleculeNamesBothListsNonEmpty = "Molecule names to transport and molecule names not to transport are both nonempty";
       public static readonly string InvalidFile = "Invalid File";
+      public static readonly string InvalidAuxiliaryType = "The error type is invalid. Valid types are ArithmeticStdDev and GeometricStdDev";
 
       public static string UserDefinedPKParameterNotFound(string pkParameterName) => $"PK-Parameter '{pkParameterName}' not found";
 
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -1431,7 +1431,7 @@ namespace OSPSuite.Assets
       public static readonly string TransportMoleculeNamesNotSet = "Transport molecule names object passed is not set";
       public static readonly string TransportMoleculeNamesBothListsNonEmpty = "Molecule names to transport and molecule names not to transport are both nonempty";
       public static readonly string InvalidFile = "Invalid File";
-      public static readonly string InvalidAuxiliaryType = "The error type is invalid. Valid types are ArithmeticStdDev and GeometricStdDev";
+      public static readonly string InvalidAuxiliaryType = "The error type is invalid. Valid types are 'ArithmeticStdDev' and 'GeometricStdDev'";
 
       public static string UserDefinedPKParameterNotFound(string pkParameterName) => $"PK-Parameter '{pkParameterName}' not found";
 
--- a/src/OSPSuite.R/Services/DataRepositoryTask.cs
+++ b/src/OSPSuite.R/Services/DataRepositoryTask.cs
@@ -14,7 +14,7 @@ namespace OSPSuite.R.Services
 
       DataColumn GetErrorColumn(DataColumn column);
 
-      DataColumn AddErrorColumn(DataColumn column, string name = "yError", string errorType = null);
+      DataColumn AddErrorColumn(DataColumn column, string name, string errorType);
    }
 
    public class DataRepositoryTask : IDataRepositoryTask
@@ -47,7 +47,7 @@ namespace OSPSuite.R.Services
          return null;
       }
 
-      public DataColumn AddErrorColumn(DataColumn column, string name = "yError", string errorType = null)
+      public DataColumn AddErrorColumn(DataColumn column, string name, string errorType)
       {
          var errorColumn = new DataColumn(name, column.Dimension, column.BaseGrid);
 
--- a/src/OSPSuite.R/Services/DataRepositoryTask.cs
+++ b/src/OSPSuite.R/Services/DataRepositoryTask.cs
@@ -1,6 +1,7 @@
 ﻿using OSPSuite.Assets;
 using OSPSuite.Core.Domain.Data;
 using OSPSuite.Core.Serialization.Xml;
+using OSPSuite.Utility.Exceptions;
 using System;
 
 namespace OSPSuite.R.Services
@@ -55,7 +56,7 @@ namespace OSPSuite.R.Services
 
          AuxiliaryType auxiliaryType;
          if (!Enum.TryParse(errorType, out auxiliaryType))
-            throw new Exception(Error.InvalidAuxiliaryType);
+            throw new OSPSuiteException(Error.InvalidAuxiliaryType);
             
          errorColumn.DataInfo.AuxiliaryType = auxiliaryType;
          errorColumn.DisplayUnit = column.DisplayUnit;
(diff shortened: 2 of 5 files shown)


src/OSPSuite.R/Services/DataRepositoryTask.cs
```diff
@@ -28,5 +33,28 @@ public void SaveDataRepository(DataRepository dataRepository, string fileName)
...
+         return null;
+      }
+
+      public DataColumn AddErrorColumn(DataColumn column, string name = "yError", string errorType = null)
+      {
+         var errorColumn = new DataColumn(name, column.Dimension, column.BaseGrid);
+
+         errorColumn.DisplayUnit = column.DisplayUnit;
+         AuxiliaryType auxiliaryType;
+         errorColumn.DataInfo.AuxiliaryType = Enum.TryParse(errorType, out auxiliaryType) ? auxiliaryType : AuxiliaryType.ArithmeticStdDev;
```
Yuri05, line 54: Not possible to pass enum from R?
If so: if `Enum.TryParse(errorType, out auxiliaryType)` returns false - I would throw an exception and not (silently) set error to Arithmetic

PavelBal: No rClr does not like enums.

The idea was that the erro is arithmetic by default - but now I think it is OK to always require error type and throw an error if an unsupported type is set.

abdelr: TryParse will not throw if unsupported value. It will return false so we can set to the default value. If we rather like the function to throw on unsupported mandatory error type, we need to change the code.

Yuri05: I don't have an opinion if it should be mandatory or not. But if it's optional, then these are 2 different use cases for me:
* parameter not passed (null) - use default
* parameter passed - it must be a valid value, otherwise throw exception (at the moment something like `AddErrorColumn(column, "error", "gmetric") `would create an arithmetic error, which is wrong for sure)

PavelBal: Agree with Yuri.

abdelr: I also see what you say @Yuri05 and I agree `AddErrorColumn(column, "error", "gmetric")` should not be valid. It is a bit strange for me though to have it optional since it is a string. So, passing null or "" will also be interpreted as default? This is why I proposed not to have it optional anymore but I can live with it optional as long as we agree on what means default behavior. So, optional? if so, should null and/or "" be interpreted as default value?

Yuri05: I'm fine to make it mandatory. But as I don't know the requirement for this functionality, I'm fine with optional too.
And if optional: exactly as you say, _should null and/or "" be interpreted as default value_

abdelr: Done, now it is optional and if null or empty will take the default value ;)

src/OSPSuite.R/Services/DataRepositoryTask.cs
```diff
@@ -28,5 +33,34 @@ public void SaveDataRepository(DataRepository dataRepository, string fileName)
...
+      public DataColumn AddErrorColumn(DataColumn column, string name = "yError", string errorType = null)
+      {
+         var errorColumn = new DataColumn(name, column.Dimension, column.BaseGrid);
+
+         if (string.IsNullOrEmpty(errorType))
+            errorType = AuxiliaryType.ArithmeticStdDev.ToString();
+
+         AuxiliaryType auxiliaryType;
+         if (!Enum.TryParse(errorType, out auxiliaryType))
+            throw new Exception();
```
Yuri05, line 57: `throw new OSPSuiteException(Consts.Error.InvalidAuxiliaryType )` ?

abdelr: Sure ;)

src/OSPSuite.R/Services/DataRepositoryTask.cs
```diff
@@ -28,5 +34,34 @@ public void SaveDataRepository(DataRepository dataRepository, string fileName)
...
+      public DataColumn AddErrorColumn(DataColumn column, string name = "yError", string errorType = null)
+      {
+         var errorColumn = new DataColumn(name, column.Dimension, column.BaseGrid);
+
+         if (string.IsNullOrEmpty(errorType))
+            errorType = AuxiliaryType.ArithmeticStdDev.ToString();
+
+         AuxiliaryType auxiliaryType;
+         if (!Enum.TryParse(errorType, out auxiliaryType))
+            throw new Exception(Error.InvalidAuxiliaryType);
```
Yuri05, line 58: why not `OSPSuiteException`? usually when we throw a (handled) exception, it's always is (or is derived from) the `OSPSuiteException`. The error dialog knows then that exception was handled (otherwise "Oups.." message is shown)
If the function is called from R only it probably does not really matter, but still..

abdelr: Ok, in that case I would say there are a couple of places in the code where I have thrown Exception and not OSPSuiteException (always on the R end). I would try to fix them together on this PR if you do not mind.

Yuri05: sure :)

abdelr: Done...

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1431,6 +1431,7 @@ public static string WrongColumnDimensions(string columnName, int xDim, int yDim
       public static readonly string TransportMoleculeNamesNotSet = "Transport molecule names object passed is not set";
       public static readonly string TransportMoleculeNamesBothListsNonEmpty = "Molecule names to transport and molecule names not to transport are both nonempty";
       public static readonly string InvalidFile = "Invalid File";
+      public static readonly string InvalidAuxiliaryType = "The error type is invalid. Valid types are ArithmeticStdDev and GeometricStdDev";
```
msevestre, line 1434: Can you put ArithmeticStdDev and GeometricStdDev in quote e.g. 'GeometricStdDev'

abdelr: Sure ;)

src/OSPSuite.R/Services/DataRepositoryTask.cs
```diff
@@ -28,5 +35,34 @@ public void SaveDataRepository(DataRepository dataRepository, string fileName)
...
+         if (column.ContainsRelatedColumn(AuxiliaryType.ArithmeticStdDev))
+            return column.GetRelatedColumn(AuxiliaryType.ArithmeticStdDev);
+
+         if (column.ContainsRelatedColumn(AuxiliaryType.GeometricStdDev))
+            return column.GetRelatedColumn(AuxiliaryType.GeometricStdDev);
+
+         return null;
+      }
+
+      public DataColumn AddErrorColumn(DataColumn column, string name = "yError", string errorType = null)
```
msevestre, line 50: I have foud that optional parameters are an issue wth rCLR. Are you sure this will work?

PavelBal: Yes you are right.

@abdelr Then I think let's just make the errorType mandatory with no default value.

abdelr: They are mandatory now

Yuri05 (approved):

PR #1199, opened 2021-07-20
throw on data set with empty rows

### LINKED ISSUE #1193: Importer: Error when Reseting mapping to current sheet and data already loaded
[CompiledDataSet.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/6828691/CompiledDataSet.xlsx)

- Load attached file
- Select sheet "MissingTimeEntries"
- Reset mapping based on current sheet
- For measurement, select "Conc[nM]"
- Import sheet -> The time column is not strictly blabla...
- For "Subject Id", select column "Patient Id"
- Import 

Application:
PK-Sim® 10.0.181

Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index

Stack trace:
```
at System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument, ExceptionResource resource)
   at System.Collections.Generic.List`1.get_Item(Int32 index)
   at OSPSuite.Infrastructure.Import.Core.DataSource.ValidateDataSource(IReadOnlyList`1 columnInfos, IDimensionFactory dimensionFactory) in D:\Work\OSPS\Repos\OSPSuite.Core\src\OSPSuite.Infrastructure.Import\Core\DataSource.cs:line 166
   at OSPSuite.Presentation.Presenters.Importer.ImporterPresenter.validateDataSource(IDataSource dataSource)
   at OSPSuite.Presentation.Presenters.Importer.ImporterPresenter.loadSheets(IDataSourceFile dataSourceFile, Cache`2 sheets, String filter, String selectedNamingConvention)
   at OSPSuite.Presentation.Presenters.Importer.ImporterPresenter.onImporterDataChanged(Object sender, EventArgs args)
   at OSPSuite.Presentation.Presenters.Importer.ImporterPresenter.onCompletedMapping(Object sender, EventArgs args)
   at System.EventHandler.Invoke(Object sender, EventArgs e)
   at OSPSuite.Presentation.Presenters.Importer.ColumnMappingPresenter.ValidateMapping()
```

abdelr:
Fixes #1193 (throw on data set with empty rows)

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
@@ -91,7 +91,10 @@ namespace OSPSuite.Infrastructure.Import.Core
             else
             {
                dataSet.Value.ClearNan(indicator);
-               var emptyDataSets = dataSet.Value.Data.Where(parsedDataSet => parsedDataSet.Data.All(column => column.Value.Count == 0));
+               var emptyDataSets = dataSet.Value.Data.Where(parsedDataSet => parsedDataSet.Data.All(column => column.Value.Count == 0)).ToList();
+               if (emptyDataSets.Count == 0)
+                  continue;
+
                var emptyDataSetsNames = emptyDataSets.Select(d => string.Join(".", d.Description.Where(metaData => metaData.Value != null).Select(metaData => metaData.Value)));
                throw new EmptyDataSetsException(emptyDataSetsNames);
             }
--- a/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
@@ -156,6 +156,9 @@ namespace OSPSuite.Infrastructure.Import.Core
                      if (errorColumn.Value != null && measurementColumn.Value.Count != errorColumn.Value.Count)
                         throw new OSPSuiteException(Error.MismatchingArrayLengths);
 
+                     if (errorColumn.Value.Count == 0)
+                        throw new OSPSuiteException(Error.EmptyDataSet);
+
                      if (errorColumn.Key == null)
                         return true;
 
--- a/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
@@ -82,12 +82,19 @@ namespace OSPSuite.Infrastructure.Import.Core
          _importer.AddFromFile(_configuration.Format, filterSheets(dataSheets, filter), columnInfos, this);
          if (NanSettings == null || !double.TryParse(NanSettings.Indicator, out var indicator))
             indicator = double.NaN;
-         foreach (var dataSet in DataSets)
+         foreach (var dataSet in DataSets.KeyValues)
          {
             if (NanSettings != null && NanSettings.Action == NanSettings.ActionType.Throw)
-               dataSet.ThrowsOnNan(indicator);
+            {
+               dataSet.Value.ThrowsOnNan(indicator);
+            }
             else
-               dataSet.ClearNan(indicator);
+            {
+               dataSet.Value.ClearNan(indicator);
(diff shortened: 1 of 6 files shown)


src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -156,6 +156,9 @@ public bool ValidateDataSource(IReadOnlyList<ColumnInfo> columnInfos, IDimension
                      if (errorColumn.Value != null && measurementColumn.Value.Count != errorColumn.Value.Count)
                         throw new OSPSuiteException(Error.MismatchingArrayLengths);
```
Yuri05, line 157: how is it possible that `measurementColumn.Value.Count != errorColumn.Value.Count` ?
Or is it when columns are created "manually" e.g. from R?

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -156,6 +156,9 @@ public bool ValidateDataSource(IReadOnlyList<ColumnInfo> columnInfos, IDimension
...
 
+                     if (errorColumn.Value.Count == 0)
+                        throw new OSPSuiteException(Error.EmptyDataSet);
```
Yuri05, line 160: don't understand the use case - how can it happen that we get 0 rows in a dataset?

abdelr: We can discuss tomorrow. It happens when all rows from a dataset are filtered out due to NaN settings. I propose we discuss what solution do we want to have for V10 and after V10.

Yuri05: Can we extend the ErrorMessage with the description (=MEtadata combination) of the empty data set?
E.g. _"Your settings ended up in empty data set **'Venous Blood.Plasma.Aspirin.Day1'**. ~This could mean that for some combination of meta data values all observations are empty.~ Please remove the data set from your data or add at least one observation for it."_

Otherwise it might be hard for the user to find where the problem is.
By the way we should always try to include the description of where exactly the data is problematic. In the past our users had hard times very often, when during import of data file with very many dataset-combinations a message like "times are not unique" etc. appeared and the users tried to figure out where exactly it happened.

Yuri05 (approved):

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
```diff
@@ -184,7 +184,7 @@ private void loadSheetsFromDataPresenter(object sender, ImportSheetsEventArgs ar
          {
             {
                var eMessage = e.Message;
-               if (! (e is NanException || e is ErrorUnitException || e is MissingColumnException || e is PossibleUnsupportedSheetFormatException))
+               if (! (e is NanException || e is ErrorUnitException || e is MissingColumnException || e is PossibleUnsupportedSheetFormatException || e is EmptyDataSetsException))
```
msevestre, line 187: There is somethning that I dont' really like about this. Maybe we need to create another base class for those type of exception that we are catching explicitly.
Thoughts?

msevestre (changes requested):
I would add antoher base class for exception and remove this crazy filtering

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
abdelr: Yes, @georgeDaskalakis and I discussed this already. The code on his PR #1204 is also involved and we decided to merge this and then solve it on a separate issue to avoid conflicts. Will this work for you @msevestre?

msevestre: ok. I am going to createa an issue then and merge

msevestre: https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1218

PR #1201, opened 2021-07-20
Visually separating add group by

### LINKED ISSUE #1177: Highlight the "Add group by" in the importer
image

Add a plus-sign and change background color of the cell.

abdelr:
Fixes #1177 (Visually separating add group by)

### DIFF
--- a/src/OSPSuite.UI/Views/Importer/ColumnMappingView.cs
+++ b/src/OSPSuite.UI/Views/Importer/ColumnMappingView.cs
@@ -220,23 +220,13 @@ namespace OSPSuite.UI.Views.Importer
             columnMappingGridView.CloseEditor();
          });
 
-
-         _addButtonRepository.ButtonClick += (o, e) =>
-            OnEvent(() => _presenter.AddGroupBy(_gridViewBinder.FocusedElement.Source as AddGroupByFormatParameter));
          _disabledRemoveButtonRepository.Buttons[0].Enabled = false;
       }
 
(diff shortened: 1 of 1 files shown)


abdelr:
Solved with a minimal visual changed. I would not like to change the code in the presenter to start differentiating when it is a add group by row and when it is not. Instead, we add the plus icon on the last column.

src/OSPSuite.UI/Views/Importer/ColumnMappingView.cs
```diff
@@ -220,23 +220,13 @@ public override void InitializeBinding()
...
       {
          if (model.Source is AddGroupByFormatParameter)
-         {
-            if (string.IsNullOrEmpty(model.Source.ColumnName))
-            {
-               return _disabledRemoveButtonRepository;
-            }
-
             return _addButtonRepository;
-         }
```
Yuri05, line 239: @abdelr Not sure what happens here. Can u explain?

abdelr: The last column has the x icon for cleaning the row. For the special case of the add group by, it is meaningless so we added a plus instead (already configured in _addButtonRepository). So, just a visual change adding plus icon on add group by row, and keeping x icon for clear row

Yuri05: ok, thx

Yuri05 (approved):

PR #1204, opened 2021-07-21
188 wrong dimension in preview

### LINKED ISSUE #1188: Importer: Wrong dimension in preview
If the unit of the time column is "day(s)", wrong dimension is presented in the preview:

image

georgeDaskalakis:
closes #1188


src/OSPSuite.Infrastructure.Import/Core/DataFormat/DataFormatHeadersWithUnits.cs
```diff
@@ -33,6 +34,7 @@ protected override UnitDescription ExtractUnits(string description, IUnformatted
             .Substring(1, units.Length - 2) //remove the brackets
             .Trim() //remove whitespace
             .Split(',') //split comma separated list
+            .Where(unitName => supportedDimensions.Any(x => x.HasUnit(unitName))) //only accepts valid and supported units
```
Yuri05, line 37: I don't understand this part and it seems to me it will not work if the caption has more then one [] 
E.g. what would the unit be for `"[min] residence time [h]"`? Should be `h`

georgeDaskalakis: Actually for this case, what will be recognized as a unit will be everything that is between the first [ and the last ] ( so in this example "min] residence time [h" ), and thus the unit will not be automatically detected. This means that if somebody tries to import a file with a header "Time [min] residence time [h]", the initial mapping for time will be invalid unit, and the user should fix it manually. Personally I do not see this as that big a problem.  But if you believe it should be adjusted I can do a new issue for this.

In any case, this was like that also before the fix. All thsi added .where() clause does, is making suire that the initial mapping does not assign a unit to a mapping that does not belong to one of its supported mappings: a header with "Concentration[s]" will result in the initial unit for concentration being set to invalid unit. Before this change, concentration would be assigned initially assigned unit "s" and dimension "time" - something that should not be possible.

Yuri05: how about this? Just checks if the (trimmed) caption ends with `[something]` and takes `something `as unit (if it's supported): 
```
if (!name.Trim().EndsWith("]") || name.LastIndexOf("[")<0)
    return new UnitDescription();
var unit = name.Substring(name.LastIndexOf("[") + 1, name.LastIndexOf("]") - name.LastIndexOf("[") - 1);

if (!supportedDimensions.Any(x => x.HasUnit(unit)))
  unit = UnitDescription.InvalidUnit;
         ...
```
Results in:
- "Time [h]" => h
- "[min] Time [h]" =>h
- "Time [blabla] [h]"=> h
- "[Time after dose] [h]" => h
- "[min] Time"=> InvalidUnit
- "Time [semester]" => InvalidUnit
etc.

georgeDaskalakis: I mean.....we could of course implement it like that. But do you really think this is a good timing to change something like this? Is it really a plausible scenario to have more than one pair of brackets in the header? If we need this, let's create a new issue for this and I will implement it.

Yuri05: If it's sufficiently covered by tests - why not?
Also as you already said it's just the **initial** mapping - so I don't see any dangerous side effects.
And having the ß-test phase in front - I would risk the change :)

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -122,7 +124,7 @@ protected virtual void ExtractNonQualifiedHeadings(List<string> keys, List<strin
                );
             if (headerKey == null) continue;
             keys.Remove(headerKey);
-            var units = ExtractUnits(headerKey, data, keys, ref rank);
+            var units = ExtractUnits(headerKey, data, keys, columnInfos.First(ci => ci.DisplayName == header).SupportedDimensions.ToList(),  ref rank);
```
msevestre, line 127: I don't like this ToList here. Make sure ExtractUnits takes a IReadOnly instead of a List and you are good to go no?

src/OSPSuite.Infrastructure.Import/Core/DataFormat/DataFormatNonmem.cs
```diff
@@ -25,7 +26,7 @@ protected override string ExtractLloq(string description, IUnformattedData data,
          return lloqKey;
       }
 
-      protected override UnitDescription ExtractUnits(string description, IUnformattedData data, List<string> keys, ref double rank)
+      protected override UnitDescription ExtractUnits(string description, IUnformattedData data, List<string> keys, List<IDimension> supportedDimensions, ref double rank)
```
msevestre, line 29: Here, Not List but IReadOnly. 
We should aim at having method that only take immutable object as parameter

msevestre (changes requested):

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -70,25 +71,26 @@ private void setSecondaryColumnUnit(IReadOnlyList<ColumnInfo> columnInfos)
...
+         foreach (var header in columnInfos)
          {
-            var headerKey = keys.FindHeader(header);
+            var headerName = header.DisplayName;
+            var headerKey = keys.FindHeader(headerName);
             if (headerKey != null)
             {
                keys.Remove(headerKey);
-               var units = ExtractUnits(headerKey, data, keys, ref rank);
+               var units = ExtractUnits(headerKey, data, keys, header.SupportedDimensions.ToList(), ref rank);
```
msevestre, line 85: here too. . No ToList

src/OSPSuite.Infrastructure.Import/Core/DataFormat/DataFormatHeadersWithUnits.cs
msevestre: @georgeDaskalakis I guess you are still working on this?

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -153,36 +152,80 @@ public bool ValidateDataSource(IReadOnlyList<ColumnInfo> columnInfos, IDimension
...
+                     for (var i = 0; i < column.Value.Count(); i++)
                      {
-                        if (double.IsNaN(errorColumn.Value.ElementAt(i).Measurement))
+                        if (double.IsNaN(column.Value.ElementAt(i).Measurement))
                            continue;
 
-                        if (dimensionFactory.DimensionForUnit(measurementColumn.Value.ElementAt(i).Unit) !=
-                            dimensionFactory.DimensionForUnit(errorColumn.Value.ElementAt(i).Unit))
-                           return false;
+                        if (!columnInfo.SupportedDimensions.Any(x => x.HasUnit(column.Value.ElementAt(i).Unit)))
```
msevestre, line 213: I would love to see some comments as to WHAT We are checking. Same goes for the method above. These are 3 imbricatd forEach etc... very complicated to comprehend for an outsider

src/OSPSuite.Infrastructure.Import/Services/Importer.cs
```diff
@@ -80,7 +80,7 @@ public void AddFromFile(IDataFormat format, Cache<string, DataSheet> dataSheets,
                data.AddData(format.Parse(sheetKeyValue.Value.RawData, columnInfos));
                dataSets.Add(sheetKeyValue.Key, data);
             }
-            catch
+            catch (Exception e)
```
msevestre, line 83: what about e? Don't need it? Then remove

msevestre (changes requested):

src/OSPSuite.Infrastructure.Import/Core/DataFormat/DataFormatHeadersWithUnits.cs
georgeDaskalakis: yes, but in this new issue: https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1217

src/OSPSuite.Infrastructure.Import/Services/Importer.cs
georgeDaskalakis: true, removed it

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
georgeDaskalakis: done

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
georgeDaskalakis: done

src/OSPSuite.Infrastructure.Import/Core/DataFormat/DataFormatNonmem.cs
georgeDaskalakis: done

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
georgeDaskalakis: comments added

msevestre (commented):
Once @Yuri05's changes are addressed, we can merge

georgeDaskalakis:
> Once @Yuri05's changes are addressed, we can merge

I would say we can merge, the changes @Yuri05 requested are implemented here: https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/pull/1219

PR #1206, opened 2021-07-22
Fixes #1205 Clear history does not shrink the database

### LINKED ISSUE #1205: Clear history does not shrink the database
All history and command entries are properly deleted but the database is not shrinked at the end.

Yuri05:


src/OSPSuite.Infrastructure.Serialization/Services/HistoryTask.cs
```diff
@@ -71,6 +71,7 @@ private void clearCommand(DbConnection connection)
          connection.ExecuteNonQuery("DELETE FROM HISTORY_ITEMS;");
          connection.ExecuteNonQuery("DELETE FROM COMMAND_PROPERTIES;");
          connection.ExecuteNonQuery("DELETE FROM COMMANDS;");
+         connection.ExecuteNonQuery("VACUUM;");
```
Yuri05, line 74: @msevestre Was it just forgotten or did we have any issues when trying to compress the database when called from PK-Sim or MoBi?

msevestre: I think we call vaccum on save all the time (to verify). So this should not matter much because when the project is saved next, vacuum is called. But this does not hurt I guess;-)

Yuri05: >I think we call vaccum on save all the time (to verify).

Seems it's not the case. Just had a project in MoBi with very large history (almost 50% of the project size).
Called ClearHistory and closed the project - project size did not change.
Opened the project with sqlite and called vacuum - project shrinked.

msevestre: Did you save the project? Close is not enough

Yuri05: ok, probably not. just tried with another project - indeed, on save it's shrinked.

PR #1212, opened 2021-07-26
Adding AddMetaData to DataRepositoryTask

### LINKED ISSUE #1211: DataRepositoryTask: method addMetaData(key, value)
A wrapper for R that allows to add a key-value (both strings) pair to a data repository.

Something like 

```
addMetaData(String key, String value, DataRepository dataRepository){
    dataRepository.ExtendedProperties.Add(new ExtendedProperty<string>() { Name = key, Value = value });
}
```

Reason is because I cannot create a generic type object `ExtendedProperty<T>` from R.

abdelr:
Fixes #1211 (Adding AddMetaData to DataRepositoryTask)

### DIFF
--- a/tests/OSPSuite.R.Tests/Services/DataRepositoryTaskSpecs.cs
+++ b/tests/OSPSuite.R.Tests/Services/DataRepositoryTaskSpecs.cs
@@ -150,7 +150,7 @@ namespace OSPSuite.R.Services
       }
 
       [Observation]
-      public void should_create_error_column()
+      public void should_create_meta_data()
       {
          sut.AddMetaData(_dataRepository, "meta_data", "value");
          _dataRepository.ExtendedPropertyValueFor("meta_data").ShouldBeEqualTo("value");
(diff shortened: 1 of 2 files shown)


abdelr:
I did not add any check such as key, or value are not null or empty because this will be handled by the ExtendedProperties member anyways but if you think we should add the, please let me know.

tests/OSPSuite.R.Tests/Services/DataRepositoryTaskSpecs.cs
```diff
@@ -138,4 +138,22 @@ public override void Cleanup()
...
+      private DataRepository _dataRepository;
+
+      protected override void Context()
+      {
+         base.Context();
+         _dataRepository = DomainHelperForSpecs.ObservedData("TOTO", _dimensionTask.DimensionByName(Constants.Dimension.TIME), _dimensionTask.DimensionByName(Constants.Dimension.MOLAR_CONCENTRATION));
+      }
+
+      [Observation]
+      public void should_create_error_column()
```
msevestre, line 153: rename the obs to match the test

PR #1214, opened 2021-07-26
DataImporterTask should not cast to ICsvDynamicSeparatorSelector but rely on the DI mechanism

### LINKED ISSUE #1213: DataImporterTask should not cast to ICsvDynamicSeparatorSelector but rely on the DI mechanism

abdelr:
Fixes #1213 (DataImporterTask should not cast to ICsvDynamicSeparatorSelector but rely on the DI mechanism)

### DIFF
--- a/src/OSPSuite.R/RRegister.cs
+++ b/src/OSPSuite.R/RRegister.cs
@@ -32,6 +32,7 @@ namespace OSPSuite.R
 
             // This will be registered as singleton
             scan.ExcludeType<RGroupRepository>();
+            scan.ExcludeType<CsvSeparatorSelector>();
             scan.WithConvention<OSPSuiteRegistrationConvention>();
          });
 
--- a/src/OSPSuite.R/RRegister.cs
+++ b/src/OSPSuite.R/RRegister.cs
@@ -44,7 +44,7 @@ namespace OSPSuite.R
          container.Register<IOSPSuiteExecutionContext, RExecutionContext>(LifeStyle.Singleton);
          container.Register<IOSPSuiteLogger, RLogger, RLogger>(LifeStyle.Singleton);
          container.Register<IEventPublisher, EventPublisher>(LifeStyle.Singleton);
-         container.Register<ICsvSeparatorSelector, CsvSeparatorSelector>(LifeStyle.Singleton);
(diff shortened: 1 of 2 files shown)


abdelr:
Methods are already covered by tests...

src/OSPSuite.R/RRegister.cs
```diff
@@ -44,7 +44,7 @@ public override void RegisterInContainer(IContainer container)
          container.Register<IOSPSuiteExecutionContext, RExecutionContext>(LifeStyle.Singleton);
          container.Register<IOSPSuiteLogger, RLogger, RLogger>(LifeStyle.Singleton);
          container.Register<IEventPublisher, EventPublisher>(LifeStyle.Singleton);
-         container.Register<ICsvSeparatorSelector, CsvSeparatorSelector>(LifeStyle.Singleton);
+         container.Register<ICsvDynamicSeparatorSelector, ICsvSeparatorSelector, CsvSeparatorSelector>(LifeStyle.Singleton);
```
msevestre, line 47: you need to Exclude from auto registration no?

src/OSPSuite.R/Services/DataImporterTask.cs
```diff
@@ -56,7 +56,7 @@ IContainer container
          //Should actually ask for ICsvDynamicSeparatorSelector as a dependency but I do not currently know
          //how to serve two interfaces with the same singleton. If there is no way to do it with the current
```
msevestre, line 57: remove comments

msevestre (changes requested):

src/OSPSuite.R/RRegister.cs
abdelr: It was working fine without Excluding the type. I have just explicitly exclude it but not sure if necessary.

src/OSPSuite.R/Services/DataImporterTask.cs
abdelr: Done

src/OSPSuite.R/RRegister.cs
msevestre: oh it's in R only? I don't think there is automatic registration tehre

msevestre: oh yes there is. Then it would have created an issue when getting  ICsvSeparatorSelector (at least in theory)

PR #1216, opened 2021-07-26
Fixes #1215 plot backgtround

### LINKED ISSUE #1215: Plots: Default BG color does not match
@Yuri05 commented on [Mon Jul 05 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1822)

tbf

---

@msevestre commented on [Mon Jul 26 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1822#issuecomment-886804778)

Probably a fix in core but potentially something to do in MoBi as well

---

@msevestre commented on [Mon Jul 26 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1822#issuecomment-886809489)

I actually cannot reproduce when I change the default color. So it has to be something related to the default chart color. Investigating

---

@msevestre commented on [Mon Jul 26 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1822#issuecomment-886812712)

yeah it's the default color that is not supported (Transparent). Changing in core to white

msevestre:


msevestre:
This color is not supported when exporting to the clipboard. Not sure why. (probably png, vs emf etc..) Anyways. White will do

PR #1219, opened 2021-07-27
#1217 fixed

### LINKED ISSUE #1217: Importer: second set of [unit] in header not automatically identified
The initial automatic detection of units, when the units are in the headers does not support headers where there are more than one set of [ ]. Fe, "Time [blabla] [h]" as a cilumn name would result in an unidentified unit, whereas it should result in unit= "h". The automatic detection of units in headers will be adjusted as follows:

"Time [h]" => h
"[min] Time [h]" =>h
"Time [blabla] [h]"=> h
"[Time after dose] [h]" => h
"[min] Time"=> InvalidUnit
"Time [semester]" => InvalidUnit

georgeDaskalakis:
closes #1217


src/OSPSuite.Infrastructure.Import/Core/DataFormat/MixColumnsDataFormat.cs
```diff
@@ -37,7 +37,8 @@ protected override string ExtractLloq(string description, IUnformattedData data,
       }
       protected override UnitDescription ExtractUnits(string description, IUnformattedData data, List<string> keys, ref double rank)
       {
-         var units = Regex.Match(description, @"\[.+\]").Value;
+         var units = Regex.Match(description, @"\[([^\]\[]*)\]$").Value;
```
msevestre, line 40: Should not we have one place with the logic to extra the unit from a text? Here it's the same code as DataFormatHeaderWithUnit. Why?

georgeDaskalakis: I have created two helper funtions in the base class to fix the code duplication that you are correctly pointing out. I did not move the whole code out, since the two different implementations vary a bit in the way the handle the incrementation of the ranking.

PR #1226, opened 2021-07-27
Fixes #1225 empty value crash

### LINKED ISSUE #1225: Get values for empty datacolumn should not crash
REquired for R

msevestre:


src/OSPSuite.Infrastructure.Import/Core/AbstractImporterException.cs
```diff
@@ -0,0 +1,11 @@
+﻿using OSPSuite.Utility.Exceptions;
+
+namespace OSPSuite.Infrastructure.Import.Core
+{
+   public abstract class AbstractImporterException : OSPSuiteException
```
msevestre, line 5: @abdelr @georgeDaskalakis I have renamed it to be SINGULAR (one exception). Also constructor on abstract class should be protected.

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
```diff
@@ -172,26 +174,14 @@ private void loadSheetsFromDataPresenter(object sender, ImportSheetsEventArgs ar
...
-               );
-               
-               foreach (var sheetName in args.Sheets.Keys)
-               {
-                  _importerDataPresenter.Sheets.Remove(sheetName);
-               }
-            }
+            var message = e is AbstractImporterException ? e.Message : Captions.Importer.UnexpectedExceptionWhenLoading;
+            _dialogCreator.MessageBoxError(message);
+            args.Sheets.Keys.Each(_importerDataPresenter.Sheets.Remove);
```
msevestre, line 184: It's good to use a more functional way of writing for loops in my option. I'd suggest to avoid foreach in general

PR #1227, opened 2021-07-27
1221 exclude empty mapping

### LINKED ISSUE #1221: Non-mapped meta data is added to the default naming
grafik

I would also suggest to remove `{Source}` from the default pattern.
(I always remove `{Source}.{Sheet}` from the naming pattern. While {Sheet} might be usefull for multisheet imports, I don't see any use cases where {Source} would be required. And if so, it still can be added manually

abdelr:
Fixes #1221 (Excluding empty mappings)

### DIFF
--- a/tests/OSPSuite.Presentation.Tests/Importer/Presenters/ImporterPresenterSpecs.cs
+++ b/tests/OSPSuite.Presentation.Tests/Importer/Presenters/ImporterPresenterSpecs.cs
@@ -421,11 +421,13 @@ namespace OSPSuite.Presentation.Importer.Presenters
 
    public class When_loading_configuration : concern_for_ImporterPresenter
    {
+      private Cache<string, DataSheet> _sheets;
 
-      protected override void Because()
+      protected override void Context()
       {
-         var sheets = new Cache<string, DataSheet>();
-         sheets.Add("sheet1", A.Fake<DataSheet>());
+         base.Context();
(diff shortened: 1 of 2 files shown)


tests/OSPSuite.Presentation.Tests/Importer/Presenters/ImporterPresenterSpecs.cs
```diff
@@ -415,4 +418,32 @@ public void must_validate_mapping()
          A.CallTo(() => _columnMappingPresenter.ValidateMapping()).MustHaveHappened();
       }
    }
+
+   public class When_loading_configuration : concern_for_ImporterPresenter
+   {
+
+      protected override void Because()
+      {
+         var sheets = new Cache<string, DataSheet>();
```
msevestre, line 427: This is more CONTEXT of the test. .I am having trouble seeing what causes the configuration load. Can you separate Context from action?

msevestre (changes requested):

tests/OSPSuite.Presentation.Tests/Importer/Presenters/ImporterPresenterSpecs.cs
abdelr: I hope it makes more sense now?

msevestre (commented):
perfect.

PR #1228, opened 2021-07-27
Code refactoring

msevestre:


msevestre (commented):
@abdelr Don't forget to run resharper formatting before committing files. 
Also I updated the code to remove the usage of  the magic string  "Dimensionsless"

PR #1231, opened 2021-07-28
1223 initial mapping

### LINKED ISSUE #1223: Wrong default mapping
When I import this file [011.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/6885049/011.xlsx), 
`Study Id` and `Subject Id` are interpreted as Measurement and Error in the defualt mapping.
grafik

And when I import this file [011_2.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/6885065/011_2.xlsx) (just changed the order of columns) - Measurement is properyl mapped per default, but Error is mapped to the Study Id 
(interestingly also assuming arithmetic error with the same unit as measurement).
grafik

abdelr:
Fixes #1223 (Correcting initial mapping)


Yuri05 (approved):

msevestre:
I love the tests! So refreshing to see the issue tested. Well done

PR #1236, opened 2021-07-28
Visual fixes

### LINKED ISSUE #1234: Importer: Table alignments
Preview table is a bit wider than the preview plot

image

abdelr:
Fixes #1234 (Fixing table alignment and hiding the row indicator panel)


abdelr:
It also hides the row indicator panel on the grid for the preview (requested by Pavel today) and the raw view of the excel which was not discussed today but should be handled in the same manner as the preview of the data repository in my opinion.

PR #1237, opened 2021-08-04
1167 integration tests

georgeDaskalakis:


tests/OSPSuite.Presentation.Tests/OSPSuite.Presentation.Tests.csproj
```diff
@@ -14,7 +14,7 @@
 
   <ItemGroup>
     <PackageReference Include="FakeItEasy" Version="6.2.1" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.1.5" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.1.0" />
```
Yuri05, line 17: @georgeDaskalakis Downgrading the package?

georgeDaskalakis: @Yuri05 there were errors because the same package in Core has version 3.1.0. I opted for changing the package version here, since it is just a test project, instead of updating the Core package. Of course if you think it is preferable, I can change update the version in all other projects in the solution to be 3.1.5

Yuri05: I think it's ok to keep it like this (would try to make as few changes to Core as possible until V10 release).
Just wanted to make sure it's not an "unintended change" :)

tests/OSPSuite.Core.IntegrationTests/OSPSuite.Core.IntegrationTests.csproj
```diff
@@ -49,7 +49,10 @@
   <ItemGroup>
     <ProjectReference Include="..\..\src\OSPSuite.Core\OSPSuite.Core.csproj" />
     <ProjectReference Include="..\..\src\OSPSuite.Infrastructure.Castle\OSPSuite.Infrastructure.Castle.csproj" />
+    <ProjectReference Include="..\..\src\OSPSuite.Infrastructure.Import\OSPSuite.Infrastructure.Import.csproj" />
     <ProjectReference Include="..\..\src\OSPSuite.Infrastructure\OSPSuite.Infrastructure.csproj" />
+    <ProjectReference Include="..\..\src\OSPSuite.Presentation\OSPSuite.Presentation.csproj" />
+    <ProjectReference Include="..\..\src\OSPSuite.UI\OSPSuite.UI.csproj" />
```
msevestre, line 55: why is the UI requiured?

tests/OSPSuite.Presentation.Tests/Services/DataImporterSpecs.cs
```diff
@@ -0,0 +1,291 @@
...
+      [Observation]
+      public void should_return_empty_on_invalid_file_name()
+      {
+         sut.ImportFromConfiguration(_importerConfiguration, _metaDataCategories, _columnInfos, _dataImporterSettings,
+            getFileFullName(
+               "")).Count.ShouldBeEqualTo(0);
+      }
+
+      [Observation]
+      public void should_return_null_on_non_existent_file_name()
```
msevestre, line 184: name does not match what's tested (null vs 0)

tests/OSPSuite.Presentation.Tests/Services/DataImporterSpecs.cs
```diff
@@ -0,0 +1,291 @@
...
+      [Observation]
+      public void should_import_simple_data_from_excel()
+      {
+         sut.ImportFromConfiguration(_importerConfiguration, _metaDataCategories, _columnInfos, _dataImporterSettings,
+            getFileFullName(
+               "IntegrationSample1.xlsx")).Count.ShouldBeEqualTo(1);
+      }
+
+      [Observation]
+      public void use_sheet_names_when_importing_simple_data_from_csv()
```
msevestre, line 168: Name does not seem to match what's tested

tests/OSPSuite.Presentation.Tests/Services/DataImporterSpecs.cs
```diff
@@ -0,0 +1,291 @@
...
+      public void should_throw_on_invalid_file_type()
+      {
+         sut.ImportFromConfiguration(_importerConfiguration, _metaDataCategories, _columnInfos, _dataImporterSettings,
+            getFileFullName(
+               "simple.pkml")).Count.ShouldBeEqualTo(0);
+         A.CallTo(() => _dialogCreator.MessageBoxError(Error.UnsupportedFileType)).MustHaveHappened();
+      }
+
+      [Observation]
+      public void should_return_null_on_invalid_file_format()
```
msevestre, line 201: same here

msevestre (changes requested):

tests/OSPSuite.Presentation.Tests/OSPSuite.Presentation.Tests.csproj
georgeDaskalakis: Is it then OK if I update the version in Core to 3.1.5? Locally there seem to be no problems with such a change.

msevestre: No. See comment below
The version in core should be 3.1.0. so you're changes are ok (downgrade is good)

tests/OSPSuite.Core.IntegrationTests/OSPSuite.Core.IntegrationTests.csproj
georgeDaskalakis: in order to register the ICsvSeparatorSelector implementation. Since we are testing end to end importing from files, the instantiation of the objects needs a ICsvSeparatorSelector registration.

msevestre: Why is the CsvSeparatorSelector in UI then? Integrationtests should never rely on UI

tests/OSPSuite.Presentation.Tests/Services/DataImporterSpecs.cs
georgeDaskalakis: fixed

tests/OSPSuite.Core.IntegrationTests/OSPSuite.Core.IntegrationTests.csproj
georgeDaskalakis: Actually you are right, I think I have to fake this anyway. I will change this.

tests/OSPSuite.Presentation.Tests/Services/DataImporterSpecs.cs
georgeDaskalakis: name and test functionality corrected

tests/OSPSuite.Presentation.Tests/OSPSuite.Presentation.Tests.csproj
georgeDaskalakis: OK I ll leave it that way. Sorry for the misunderstanding.

tests/OSPSuite.Presentation.Tests/Services/DataImporterSpecs.cs
```diff
@@ -0,0 +1,291 @@
...
+      public void should_return_empty_on_invalid_file_format()
+      {
+         sut.ImportFromConfiguration(_importerConfiguration, _metaDataCategories, _columnInfos, _dataImporterSettings,
+            getFileFullName(
+               "sample1.xlsx")).Count.ShouldBeEqualTo(0);
+      }
+
+      //currently not working
+      /*
+            [Observation]
```
Yuri05, line 210: should we uncomment the test and just add `[Ignore]` ?

georgeDaskalakis: I have ignored the test for now, but good that you saw that Yuri. It addresses an exception handling issue that we should talk about. I am going to create an issue about this.

PR #1248, opened 2021-08-09
fixes #1241

### LINKED ISSUE #1241: Importer: Wrong values when importing with LLOQ?
[CompiledDataSet.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/6945325/CompiledDataSet.xlsx)

When trying to import the first sheet of the attached file, the second data set contains entries <1 and <0.5 for measurement.

image

In the import preview, the LLOQ-line is correctly set at 1, but the imported values for the entries are 0.5 and 0.25?

image

georgeDaskalakis:
closes #1241

### DIFF
--- a/tests/OSPSuite.Infrastructure.Tests/Import/DataSetToDataRepositoryMapperSpecs.cs
+++ b/tests/OSPSuite.Infrastructure.Tests/Import/DataSetToDataRepositoryMapperSpecs.cs
@@ -237,6 +237,7 @@ namespace OSPSuite.Infrastructure.Import
    {
       protected override void Because()
       {
+         var temp = _dataSourceLLOQ.DataSetAt(0);
          _result = sut.ConvertImportDataSet(_dataSourceLLOQ.DataSetAt(0));
       }
 
@@ -250,6 +251,12 @@ namespace OSPSuite.Infrastructure.Import
       public void should_add_maximum_lloq_value_to_data_repository()
       {
          _result.DataRepository.ObservationColumns().First().DataInfo.LLOQ.ShouldBeEqualTo(2.0f);
+      }
+
+      [Observation]
+      public void values_should_also_be_assigned_the_maximum_lloq_value()
+      {
+         _result.DataRepository.ObservationColumns().First().Values.ToArray().ShouldBeEqualTo(new float[] { 10.0f, 1.0f, 1.0f });
       }
    }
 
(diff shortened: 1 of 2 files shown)


tests/OSPSuite.Infrastructure.Tests/Import/DataSetToDataRepositoryMapperSpecs.cs
```diff
@@ -251,6 +267,12 @@ public void should_add_maximum_lloq_value_to_data_repository()
       {
          _result.DataRepository.ObservationColumns().First().DataInfo.LLOQ.ShouldBeEqualTo(2.0f);
       }
+
+      [Observation]
+      public void values_should_also_be_assigned_the_maximum_lloq_value()
+      {
+         _result.DataRepository.ObservationColumns().First().Values.ToArray().ShouldBeEqualTo(new float[] { 10.0f, 1.0f, 1.0f });
```
msevestre, line 274: This is not what we had discussed. The value assigned in the repo should be LLOQ/2
/cc @PavelBal @Yuri05

msevestre (changes requested):
@georgeDaskalakis As far as I can tell, the LLOQ value is not set properly. It should be the HIGHEST LLOQ value in the column divided by 2
@Yuri05 Please confirm

tests/OSPSuite.Infrastructure.Tests/Import/DataSetToDataRepositoryMapperSpecs.cs
PavelBal: Yes.

georgeDaskalakis: Agreed, but it is assigned correctly no? You can see also in the previous test ( should_add_maximum_lloq_value_to_data_repository ) that the repository's LLOQ correctly gets set to 2.0f. That means that both measurements should then be truncated to 1.0f. Right? Or am I missing somenthing?

msevestre: Form the diff, it looks like the values went from 0.5 to 1. So I thought you were not dividing by two.
I need to look at the code then. The diff is not enough

georgeDaskalakis: I understand what you mean, it is not clear just from the diff. In thetests I have introduced _parsedDataSetInconsistentLLOQ, where I change one of the LLOQ values from 1.0 to 2.0. That way we have tests for 2 separate test cases: when there are 2 LLOQ in the same dataset (inconsistentLLOQ, where we also test if there is a warning message) and the normal test case with one LLOQ (When_mapping_a_data_repository.should_use_lloq_halve_two). It is tested that the LLOQ is set to the highest one of the two and that the values are set to the half of that LLOQ.

Issue #1245, opened 2021-08-09
Importer: Check OnEvent() used throughout importer UI


georgeDaskalakis:
Since we have had some occasions of totally unhandled exceptions that were breaking the application during the testing phase, it does make sense to chec that in the importer code OnEvent() is being used everywhere to avoid such behaviour. This is especially important to make sure the changes of #1240 do not create any such behaviour.

PR #1252, opened 2021-08-12
#1245 fixed

georgeDaskalakis:
Because of the nature of the changes I would say no unit tests are needed for this specific PR. Also ,to be honest, I am not sure where I should write a test to actually catch an exception that would go unhandled because of a missing OnEvent().

### DIFF
--- a/src/OSPSuite.UI/Views/Importer/ImporterDataView.cs
+++ b/src/OSPSuite.UI/Views/Importer/ImporterDataView.cs
@@ -60,17 +60,17 @@ namespace OSPSuite.UI.Views.Importer
 
          var customButton = new CustomHeaderButton(ButtonPredefines.Combo) { ToolTip = "Select Page" };
          importerTabControl.CustomHeaderButtons.Add(customButton);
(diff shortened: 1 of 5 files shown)


src/OSPSuite.UI/Views/Importer/ImporterDataView.cs
```diff
@@ -60,17 +60,17 @@ public override void InitializeResources()
...
 
-      private void onTabControlCustomHeaderButtonClick(object sender, CustomHeaderButtonEventArgs e)
+      private void onTabControlCustomHeaderButtonClick()
       {
          var popupMenu = new DXPopupMenu { MenuViewType = MenuViewType.Menu };
          foreach (XtraTabPage page in importerTabControl.TabPages)
          {
             var menuitem = new DXMenuItem(page.Text);
-            menuitem.Click += onPageListMenuItemClick;
+            menuitem.Click += (s, a) => OnEvent(onPageListMenuItemClick, s, a);
```
msevestre, line 73: typicall s, a are called o, e lol. Let's try to be consistent next time :)

msevestre (approved):

PR #1274, opened 2021-08-29
#1264 fixed

### LINKED ISSUE #1264: DataImporter: Dimension gets reset
[CompiledDataSet.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/7047539/CompiledDataSet.xlsx)

1) Start import process and select the attached file
2) Correct unit/dimension is recognized for "Measurement": 
image
3) Click on the unit edit field of the measurement - the first dimension in the list (for PK-Sim "Concentration (molar)") is selected:

image

georgeDaskalakis:
close #1264


georgeDaskalakis:
Here, the bugf waws coming because of the event in the view of the comboBox change in the Ui getting triggered before the complete initialization of the presenter with data - that's why no unit test, since because of the UI involved it would be kind of complicated to write one.

src/OSPSuite.Presentation/Presenters/Importer/UnitsEditorPresenter.cs
```diff
@@ -50,7 +50,9 @@ public void SelectDimension(string dimensionName)
                         _dimensions.FirstOrDefault() ??
                         Constants.Dimension.NO_DIMENSION;
 
-            _selectedUnit = Dimension.DefaultUnit.Name;
+            if (_selectedUnit == null || !Dimension.HasUnit(_selectedUnit))
```
msevestre, line 53: I think this would benefit from comments explaining how the unit can not be part of the dimension available units
Also there is a defaultUnitName method I think

msevestre (changes requested):
Can you add some comments explaining why the order matters? It will be too easy to mess it up again later

PR #1276, opened 2021-08-29
WIP 1262

georgeDaskalakis:
Just adding some null checks suggested by ReSharper, although not sure the error came form those. 

The way that error occured, it seems that by clicking from one field to another in the columnMappingView, the transformation to data repository was called with a null unit. I discussed with Abdel adding a check for that and triggering an error, but we decided against it: this should never really happen, and if it does it means something in the functionality before went terribly wrong. In that case it would actualy be better to get the exception, instead of an error message.


msevestre:
@georgeDaskalakis Do you want me to review this? Or what is the status here?

georgeDaskalakis:
@msevestre if you agree with the description, you can review and then we can merge and the issue can be coinsidered resolved. Otherwise we can discuss this further.

msevestre:
it was in draft mode...that's why I asked :)

PR #1278, opened 2021-08-30
mapping view is now wider

### LINKED ISSUE #1257: Make the MappingSettings a bit wider
After renaming some of the columns, they are now cut by default
image

Let's make it a bit wider
image

abdelr:
Fixes #1257 (wider mapping view)


msevestre:
Suggestion. Put screenshots before and after in PR. OTherwise, I can't seen how it has changed

PR #1282, opened 2021-08-30
CreateConfigurationFor accepts the sheetName

### LINKED ISSUE #1271: DataImporterTask.CreateConfigurationFor should accept sheet name
`CreateConfigurationFor(string dataPath)` should also accept sheet name for which the configuration should be created (instead of always using the first sheet only).

abdelr:
Fixes #1271 (Accepts sheetName when creating a configuration from a data file)


src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -163,7 +163,7 @@ protected virtual void ExtractNonQualifiedHeadings(List<string> keys, List<strin
       protected virtual void ExtractGeneralParameters(List<string> keys, IUnformattedData data, IReadOnlyList<MetaDataCategory> metaDataCategories, ref double rank)
       {
          var columnsCopy = keys.ToList();
-         foreach (var header in columnsCopy.Where(h => metaDataCategories.Any(c => c.Name == h)))
+         foreach (var header in columnsCopy.Where(h => metaDataCategories.Select(c => c.Name).FindHeader(h) != null))
```
abdelr, line 166: Additional code change to ensure that the search for the metadata is not case sensitive

src/OSPSuite.R/Services/DataImporter.cs
```diff
@@ -191,10 +191,17 @@ private ColumnInfo createErrorColumn(ColumnInfo timeColumn, ColumnInfo concentra
          return errorInfo;
       }
 
-      public ImporterConfiguration ConfigurationFromData(string dataPath, IReadOnlyList<ColumnInfo> columnInfos, IReadOnlyList<MetaDataCategory> metaDataCategories)
+      public ImporterConfiguration ConfigurationFromData(string dataPath, IReadOnlyList<ColumnInfo> columnInfos, IReadOnlyList<MetaDataCategory> metaDataCategories, string sheetName)
```
msevestre, line 194: I don't understand WHY we have exactly the same code between R services and Presentation Service.s Shouldn't that be a common service?🤔

src/OSPSuite.R/Services/DataImporterTask.cs
```diff
@@ -17,9 +17,9 @@ public interface IDataImporterTask
    {
       ImporterConfiguration CreateConfiguration();
       ImporterConfiguration GetConfiguration(string filePath);
-      ImporterConfiguration CreateConfigurationFor(string dataPath);
+      ImporterConfiguration CreateConfigurationFor(string dataPath, string sheetName = null);
       void SaveConfiguration(ImporterConfiguration configuration, string path);
-      IReadOnlyList<DataRepository> ImportExcelFromConfiguration(string configurationPath, string dataPath);
+      IReadOnlyList<DataRepository> ImportExcelFromConfiguration(string configurationPath, string dataPath = null);
```
msevestre, line 22: Why this change here?

msevestre (commented):
Some minor questions

src/OSPSuite.R/Services/DataImporter.cs
abdelr: You are right... Already created an abstract base class 🤔

src/OSPSuite.R/Services/DataImporterTask.cs
abdelr: So you can define the sheet to use as base for the configuration (similar to reset mappings based on current sheet on the UI) or leave it null and then use the first sheet (default behavior)

msevestre: Why is this called dataPath and not sheet?

msevestre:
@abdelr Going to merge but probably that parameter should be renamed

PR #1284, opened 2021-09-01
Fixes #1283 downgrad

### LINKED ISSUE #1283: Downgrad ISharp to 1.2.0

msevestre:
@PavelBal @georgeDaskalakis 
Going to depend expliicitely on the version 1.2.0 in core as well....just to see if we can load the damn stuff

### DIFF
--- a/src/OSPSuite.Infrastructure/OSPSuite.Infrastructure.csproj
+++ b/src/OSPSuite.Infrastructure/OSPSuite.Infrastructure.csproj
@@ -44,7 +44,7 @@
     <PackageReference Include="OSPSuite.Utility" Version="4.0.0.4" />
     <PackageReference Include="Serilog" Version="2.10.0" />
     <PackageReference Include="Serilog.Extensions.Logging.File" Version="2.0.0" />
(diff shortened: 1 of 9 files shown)


src/OSPSuite.Infrastructure/OSPSuite.Infrastructure.csproj
```diff
@@ -44,7 +44,7 @@
     <PackageReference Include="OSPSuite.Utility" Version="4.0.0.4" />
     <PackageReference Include="Serilog" Version="2.10.0" />
     <PackageReference Include="Serilog.Extensions.Logging.File" Version="2.0.0" />
-    <PackageReference Include="SharpZipLib" Version="1.3.1" />
+    <PackageReference Include="SharpZipLib" Version="1.2.0" />
```
msevestre, line 47: This is the change here

msevestre (commented):
@PavelBal For some reason, mpoi used by PKSim and MoBi is the net45 version and not the ,net standard version. That's why we have a difference. We need to make it work for the .netstandard version otherwise it won't work under linux

PR #1287, opened 2021-09-01
Null check

### LINKED ISSUE #1286: DataRepository editor crashes on null values for extended properties.
Extended properties should not have empty values on the first place so the source of the problem should also be solved but for the time being, in case a pkml file is corrupted, the editor should not crash.

abdelr:
Fixes #1286 (Should not crash when editing observed data with null values on extended properties)

### DIFF
--- a/src/OSPSuite.Core/Domain/Data/DataRespositoryExtensions.cs
+++ b/src/OSPSuite.Core/Domain/Data/DataRespositoryExtensions.cs
@@ -44,7 +44,7 @@ namespace OSPSuite.Core.Domain.Data
       private static string valueMapper(IEnumerable<IExtendedProperty> properties, IExtendedProperty value)
       {
          if (value.ValueAsObject == null)
-            return "";
+            return string.Empty;
          return valueMapper(properties, value.ValueAsObject.ToString());
       }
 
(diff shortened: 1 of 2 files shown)


abdelr:
@msevestre I did not create any test since it is just a null check. If you think we should still have the test, please, let me know.

src/OSPSuite.Core/Domain/Data/DataRespositoryExtensions.cs
```diff
@@ -43,6 +43,8 @@ public static DataColumn FirstDataColumn(this DataRepository dataRepository)
 
       private static string valueMapper(IEnumerable<IExtendedProperty> properties, IExtendedProperty value)
       {
+         if (value.ValueAsObject == null)
+            return "";
```
msevestre, line 47: string.empty instead of "" please

msevestre: and in fact, it might work to just use ? in `value.ValueAsObject?.ToString()`

Last, please add a test :)

msevestre (changes requested):

src/OSPSuite.Core/Domain/Data/DataRespositoryExtensions.cs
abdelr: I tried yesterday with value.ValueAsObject?.ToString() but did not work. Today, after writing the test I tried again and it still throws  so I just switch to string.Empty.

Issue #1293, opened 2021-09-09
Importer: update issue for the "Molecular weight" mapping


PavelBal:
If the default configuration has no column assigned to the field "Molecular weight":

![image](https://user-images.githubusercontent.com/25903040/132692751-4fd2512a-9cbc-49fe-a0df-465e91ae9493.png)

And a field is manually assigned, the "x"-button for clearing the mapping is grayed out until the "manual"-field is selected by the user.

![image](https://user-images.githubusercontent.com/25903040/132692873-12c47820-6fbd-435e-9434-921d9b474fc3.png)
It only becomes active when clicking on the field "Edit extra fields". This is misleading, as I first though this mapping cannot be cleared at all.

PR #1294, opened 2021-09-09
#1293 fixed

georgeDaskalakis:
The issue #1293 arises because the grid does not get refreshed after the cell value gets changed. The event subscription is in line 53 in the ColumnMappingView

columnMappingGridView.CellValueChanged += (o, e) => OnEvent(_presenter.ValidateMapping);

Incorporating the refreshing of the data in the validateMapping solves the issue, but I am not sure it is optimal design - although the do get called together most of the time and refreshing the data after validation does make sense. @msevestre do you think I should create a new function in the presenter for these two calls or keep the changes the way they are now?


msevestre:
I think this is ok. Maybe you could add the comment explaining why the refreshData is required ? I'll wait until tomorrow morning to merge this (my time)

georgeDaskalakis:
Comment added :)

PR #1299, opened 2021-09-14
Fixes #1298 find unit case sensitive

### LINKED ISSUE #1298: ospsuite::toBaseUnit should be case insensitive
@sfrechen commented on [Mon Sep 13 2021](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/614)

The input argument `unit` should be converted to lower before being processed.

Ottherwise, the function throws an error, if for example `µg/L` is given instead of `µg/l`.
However, both should work!

---

@msevestre commented on [Tue Sep 14 2021](https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/614#issuecomment-919256830)

This needs to be done in .NET

msevestre:


src/OSPSuite.Infrastructure.Import/Services/SimulationPKAnalysesImporter.cs
```diff
@@ -108,7 +108,7 @@ private double convertValueToCoreValue(IDimension dimension, double valueInUnit,
       {
          var unit = dimension.Unit(unitName);
          if (unit == null)
-            throw new OSPSuiteException(Error.UnitIsNotDefinedInDimension(dimension.Name, unitName));
+            throw new OSPSuiteException(Error.UnitIsNotDefinedInDimension(unitName, dimension.Name));
```
msevestre, line 111: nice twist here..

src/OSPSuite.R/Services/DimensionTask.cs
```diff
@@ -92,7 +94,11 @@ public DimensionTask(IDimensionFactory dimensionFactory)
          _dimensionFactory = dimensionFactory;
       }
 
-      public bool HasUnit(string dimensionName, string unit) => DimensionByName(dimensionName).HasUnit(unit);
+      public bool HasUnit(string dimensionName, string unit)
+      {
+         var dimension = DimensionByName(dimensionName);
+         return dimension.HasUnit(unit) || dimension.SupportsUnit(unit, ignoreCase: true);
```
msevestre: for R, I am enhancing this check to return also if there is a case sensitive counter part

Yuri05 (approved):

PR #1300, opened 2021-09-15
1292 solved

georgeDaskalakis:


msevestre:
@georgeDaskalakis make sure you refrence the issue with Fixes #xx etc.. otherwise the link is gone

georgeDaskalakis:
@msevestre you are right I forgot

PR #1311, opened 2021-09-24
1301 calculation method=variation

msevestre:
This should improve the starting time of a PI drastically.  a lot of stuff was done sequentially where I think it could have been done in parallel. Let's discuss to see if we want this in this version or not.
@PavelBal It would be cool if you could try this version of ospsuite.core on your super fast computer to see how the speed has improved

### DIFF
--- a/src/OSPSuite.Core/Domain/Services/ParameterIdentifications/CategorialParameterIdentificationRunInitializer.cs
+++ b/src/OSPSuite.Core/Domain/Services/ParameterIdentifications/CategorialParameterIdentificationRunInitializer.cs
@@ -1,23 +1,32 @@
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
-using OSPSuite.Utility.Extensions;
 using OSPSuite.Core.Domain.ParameterIdentifications;
+using OSPSuite.Utility.Extensions;
 
 namespace OSPSuite.Core.Domain.Services.ParameterIdentifications
 {
-   public class CategorialParameterIdentificationRunInitializer : ParameterIdentifcationRunInitializer
+   public class CategorialParameterIdentificationRunInitializer : ParameterIdentificationRunInitializer
    {
-      private readonly ICoreSimulationFactory _simulationFactory;
+      private readonly Utility.Container.IContainer _container;
       private readonly ICategorialParameterIdentificationDescriptionCreator _descriptionCreator;
+      private readonly ICoreUserSettings _coreUserSettings;
       private CalculationMethodCombination _calculationMethodCombination;
       private bool _isSingleCategory;
       private CategorialParameterIdentificationRunMode _runMode;
 
-      public CategorialParameterIdentificationRunInitializer(ICloneManagerForModel cloneManager, IParameterIdentificationRun parameterIdentificationRun, ICoreSimulationFactory simulationFactory, ICategorialParameterIdentificationDescriptionCreator descriptionCreator) : base(cloneManager, parameterIdentificationRun)
+      public CategorialParameterIdentificationRunInitializer(
+         ICloneManagerForModel cloneManager,
+         IParameterIdentificationRun parameterIdentificationRun,
+         Utility.Container.IContainer container,
+         ICategorialParameterIdentificationDescriptionCreator descriptionCreator,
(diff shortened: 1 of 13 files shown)


src/OSPSuite.Core/Domain/Services/ParameterIdentifications/CategorialParameterIdentificationRunInitializer.cs
```diff
@@ -28,24 +37,33 @@ public void Initialize(ParameterIdentification parameterIdentification, int runI
...
+         {
+            CancellationToken = cancellationToken,
+            MaxDegreeOfParallelism = _coreUserSettings.MaximumNumberOfCoresToUse
+         };
+
          return Task.Run(() =>
          {
             var newParameterIdentification = _cloneManager.Clone(ParameterIdentification);
-            newParameterIdentification.AllSimulations.ToList().Each(originalSimulation =>
+            Parallel.ForEach(newParameterIdentification.AllSimulations.ToList(), parallelOptions, originalSimulation =>
```
msevestre, line 51: Simulations are now cloned in parallel instead of one at a time. This is a very expensive operation.

src/OSPSuite.Core/Domain/Services/ParameterIdentifications/ParameterIdentificationRun.cs
```diff
@@ -82,18 +82,25 @@ public class ParameterIdentificationRun : IParameterIdentificationRun
...
          RunResult.Description = _parameterIdentification.Description;
          _optimizationAlgorithm = _optimizationAlgorithmMapper.MapFrom(_parameterIdentification.AlgorithmProperties);
          _residualCalculator = _residualCalculatorFactory.CreateFor(_parameterIdentification.Configuration);
-         _parameterIdentification.AllSimulations.Each(s => _allSimModelBatches.Add(s, createSimModelBatch(s)));
+         var parallelOptions = createParallelOptions(_cancellationToken);
+
+         Parallel.ForEach(_parameterIdentification.AllSimulations, parallelOptions, simulation =>
+         {
+            parallelOptions.CancellationToken.ThrowIfCancellationRequested();
+            _allSimModelBatches.TryAdd(simulation, createSimModelBatch(simulation));
```
msevestre, line 101: Also initializing simModel in paralellel. When optimizing multiple simulations, this will be a massive gain in startup time

src/OSPSuite.Core/Domain/Services/ParameterIdentifications/ParameterIdentificationRun.cs
```diff
@@ -220,29 +227,32 @@ private void updateRunResult(OptimizationRunResult currentResult)
       private IReadOnlyList<SimulationRunResults> runSimulations()
       {
          var simulationResults = new ConcurrentBag<SimulationRunResults>();
-         var parallelOptions = new ParallelOptions
-         {
-            CancellationToken = _cancellationToken,
-            MaxDegreeOfParallelism = _parameterIdentification.IsSingleRun ? _coreUserSettings.MaximumNumberOfCoresToUse : 1
```
msevestre, line 226: This was the actual problem here. We were only running on one core because we assume that we would achieve max // from the categorial variation (calculation method). Using Parallel.ForEach should not create too much parallelism

msevestre:
image
monster PI

Yuri05 (approved):

msevestre:
@All: Merging so that I can build a new PKSim setup to try

PavelBal:
Well-well!
image

msevestre:
Is this...good?

PavelBal:
Yepp! But then four out of five runs (I selected all 5 available calculation methods) failed, I am trying to reproduce.

PavelBal:
Cannot reproduce. But lol, it consumes 33GB RAM.

msevestre:
Using your PI example (17 simulation) and all those variations, we create A LOT of simulations in memory I am not surprised
With 64Cores, I hope the machine has >100GB RAM

PavelBal:
@msevestre OK, I can reproduce, and it is non-deterministic - this time two out of five runs failed, and it looks like this:

image

Yuri05:
what is the error message of the faulted PIs?

msevestre:
@PavelBal you should be able to see it by going over the icon maybe? (tooltip)

PR #1320, opened 2021-09-29
1319 pi issue

msevestre:


src/OSPSuite.Core/Domain/Services/ParameterIdentifications/CategorialParameterIdentificationRunInitializer.cs
```diff
@@ -45,14 +46,22 @@ public override Task<ParameterIdentification> InitializeRun(CancellationToken ca
...
                parallelOptions.CancellationToken.ThrowIfCancellationRequested();
                var newSimulation = createNewSimulationFrom(originalSimulation, _calculationMethodCombination.CalculationMethods);
-               updateReferencesInParameterIdentification(newParameterIdentification, newSimulation, originalSimulation);
+               concurrentDictionary.TryAdd(originalSimulation, newSimulation);
+            });
+
+            concurrentDictionary.Each(kv =>
+            {
+               //Key is the old simulation, value is the corresponding updated new simulation
+               newParameterIdentification.SwapSimulations(kv.Key, kv.Value);
```
msevestre, line 64: This was changing the list of simulation in a non thread safe manner. This is now done outside of the Parallel loop

PR #1327, opened 2021-10-05
#1326 fixed

### LINKED ISSUE #1326: DataImporter: Units for Molecular weight
I think this is an important point - the "Molecular Weight" is obviously treated in base units - though I cannot imagine that any user will ever consider entering the MW values in the excel as kg/µmol!

@msevestre @Yuri05

georgeDaskalakis:
closes #1326

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Services/Importer.cs
+++ b/src/OSPSuite.Infrastructure.Import/Services/Importer.cs
@@ -6,6 +6,7 @@ using OSPSuite.Assets;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Data;
 using OSPSuite.Core.Domain.Services;
+using OSPSuite.Core.Domain.UnitSystem;
 using OSPSuite.Core.Import;
 using OSPSuite.Infrastructure.Import.Core;
 using OSPSuite.Infrastructure.Import.Core.Mappers;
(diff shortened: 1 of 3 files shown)


src/OSPSuite.Infrastructure.Import/Services/Importer.cs
```diff
@@ -210,11 +214,13 @@ public MappingProblem CheckWhetherAllDataColumnsAreMapped(IReadOnlyList<ColumnIn
...
 
-            if (!string.IsNullOrEmpty(molecularWeightDescription))
+            if (!molecularWeightValueAsString.IsNullOrEmpty())
             {
-               if (double.TryParse(molecularWeightDescription, out var molWeight))
+               if (double.TryParse(molecularWeightValueAsString, out var molWeight))
                {
-                  dataRepo.AllButBaseGrid().Each(x => x.DataInfo.MolWeight = molWeight);
+                  //we are assuming that the MW coming from the column in excel is always in g/mol, that's why we need this conversion here, 
+                  //otherwise it would be saved in kg/μmol
```
msevestre, line 222: The otherwise is weird. 
suggestion
```
 //we are assuming that the MW coming from the column in excel is always in g/mol, that's why we need the conversion to core unit to ensure that the value is saved properly
```

src/OSPSuite.Infrastructure.Import/Services/Importer.cs
```diff
@@ -226,6 +232,11 @@ public MappingProblem CheckWhetherAllDataColumnsAreMapped(IReadOnlyList<ColumnIn
          return dataRepositories;
       }
 
+      private double molWeightValueInCoreUnit(double valueInDisplayUnit)
+      {
+         return _molWeightDimension.UnitValueToBaseUnitValue(_molWeightDimension.DefaultUnit, valueInDisplayUnit);
```
msevestre, line 237: This is correct. Converting from "DefaultUnit" (g/ml) to baseUnit.

msevestre (commented):
Update the comment. Rest looks good to me

src/OSPSuite.Infrastructure.Import/Services/Importer.cs
msevestre: Yes lol g/mol is the default unit

PR #1329, opened 2021-10-05
Fixes #1328 default meta data category

### LINKED ISSUE #1328: DefaultMetaDataCategories should return IReadOnlyList

msevestre:


src/OSPSuite.Presentation/Services/DataImporter.cs
```diff
@@ -29,57 +29,6 @@ IApplicationController applicationController
          _applicationController = applicationController;
       }
 
-      public override IList<MetaDataCategory> DefaultMetaDataCategories()
```
msevestre, line 32: @abdelr @georgeDaskalakis Not sure why this was not promoted to the base class .. This is the exact same code for both R and Presentation implementation

tests/OSPSuite.Starter/Presenters/ImporterTestPresenter.cs
```diff
@@ -162,12 +162,12 @@ public void StartWithPKSimSettings()
          var dataImporterSettings = new DataImporterSettings();
          dataImporterSettings.AddNamingPatternMetaData(Constants.FILE, Constants.SHEET);
          dataImporterSettings.AddNamingPatternMetaData(Constants.FILE, Constants.SHEET, "Species");
-         var metaDataCategories = _dataImporter.DefaultMetaDataCategories();
+         var metaDataCategories = _dataImporter.DefaultMetaDataCategories().ToList();
          _dataGenerator.AddMoleculeValuesToMetaDataList(metaDataCategories);
          _dataGenerator.AddOrganValuesToMetaDataList(metaDataCategories);
          StartImporterExcelView
          (
-            (IReadOnlyList<MetaDataCategory>)metaDataCategories,
```
msevestre, line 170: The goal is that we don't have to do something like this everywhere

Issue #1344, opened 2021-10-14
runSimulation and runSimulationBatches yield different output values


abdullahhamadeh:
[simbatch_example.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/7348673/simbatch_example.zip)

First result set generated by runSimulationBatches, the second by runSimulation:

![image](https://user-images.githubusercontent.com/18409098/137379585-2cf6c78a-1269-42fb-b626-56405050c3f1.png)

abdullahhamadeh:
[debug-simbatch6.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/7350202/debug-simbatch6.zip)

Another example with a much more simplified model composed of two physical compartments only.

Two parameters are set in the simulation, but we set them in different order:

![image](https://user-images.githubusercontent.com/18409098/137416336-59db7e67-8534-42da-8879-4cdbef656705.png)

In one order, the results from runSimulationBatches/runSimulation agree, in the other order they don't.

With parSet1 the runSimulationBatches/runSimulation results are inconsistent:
![image](https://user-images.githubusercontent.com/18409098/137416893-bd42faac-7609-464a-9d51-6709377124ba.png)


With parSet2 the runSimulationBatches/runSimulation results are consistent:
![image](https://user-images.githubusercontent.com/18409098/137416924-d85943f5-0396-4623-90fa-852958fed008.png)

PR #1347, opened 2021-10-18
Reproduce bug #1344

msevestre:


msevestre:
Very interesting

This is how I set my parameters
       "DERMAL_APPLICATION_AREA|skin_compartment|Hydrated SC",
                  "DERMAL_APPLICATION_AREA|skin_compartment|SC_skin_sublayer|SC_total_thickness",

This is how it's being read 

image

in SimModel

 
image

Somehow the order is inverted when coming from SimModel!!

PR #1351, opened 2021-10-19
Using ParallelForEach on ConcurrentSimulationRunner

### LINKED ISSUE #1348: Use Parallel.ForEach when running simulations in parallel from R instead of ConcurrencyManager

abdelr:
Fixes #1348 (Using Parallel.ForEach on ConcurrentSimulationRunner)


abdelr:
> It seems that some of the code is the same as in #1345 I'd suggest in those case to make a PR onto the feature branch instead of on develop. Otherwise, it's very hard to review I am wondering why the concurrenymanager abstraction was removed. It was creating the results in a consistent way, it could ow do the Task.Run and Parallel,ForEachm making sure the token is cancelled properly, try catch etc.. There is now many instances where we do exactly the same and potentially more in the future. I would consider adding this one back to make the usage consistent across the code
> 
> Thoughts?

I have actually used the idea of merging into the feature branch in the past but I was not sure if we were synced on it. I would like to use this protocol in the future. It is the best I can think of. Still, for this time, I would merge one PR merge back from develop into the second PR, resolve conflicts...

msevestre:
>I have actually used the idea of merging into the feature branch in the past but I was not sure if we were synced on it. I would like to use this protocol in the future. It is the best I can think of

Yes this is what we had discussed. For bigger feature, this should be the norm

msevestre:
I am hesitant to merge into develop now. You can still create a feature branch PI off develop and update all your PR to have this feature branch as target. 
There are a few questions in this PR that I raised. I'll look at the other PR until they are answered

msevestre (commented):
ok

PR #1352, opened 2021-10-20
Using Parallel.ForEach on RunPopulation

### LINKED ISSUE #1349: When running a population (PopulationRunner.cs) we should use Parallel.ForEach and avoid using the PopulationDataSplitter

abdelr:
Fixes #1349 (Using Parallel.ForEach on RunPopulation)


abdelr:
Juri is right. It will slow down all threads. I a closing the PR.

PR #1359, opened 2021-11-02
1337 chart color grouping

georgeDaskalakis:


src/OSPSuite.Presentation/Binders/ObservedDataDragDropBinder.cs
```diff
@@ -24,6 +25,25 @@ public IReadOnlyList<DataRepository> DroppedObservedDataFrom(IDragEvent e)
...
       }
 
+      public Cache<string, List<DataRepository>> DroppedObservedDataWithFolderPathFrom(IDragEvent e)
+      {
+         var dataNodes = e.Data<IEnumerable<ITreeNode>>();
+
+         if (dataNodes == null)
+            return new Cache<string, List<DataRepository>>();
+
+         var treeNodes = dataNodes as IList<ITreeNode> ?? dataNodes.ToList();
```
msevestre, line 35: Can you explain why you don't just do ToList?

src/OSPSuite.Presentation/Binders/ObservedDataDragDropBinder.cs
```diff
@@ -24,6 +25,25 @@ public IReadOnlyList<DataRepository> DroppedObservedDataFrom(IDragEvent e)
...
+
+         if (areAllObservedDataNodes(treeNodes))
+            return observedDataNodesWithFolderPathFromObservedDataNodes(treeNodes);
+
+         if (areAllObservedDataClassificationNodes(treeNodes) || areAllRootNodes(treeNodes))
```
msevestre, line 40: Very nice and easy to read

src/OSPSuite.Presentation/IPresentationUserSettings.cs
```diff
@@ -55,6 +55,12 @@ public interface IPresentationUserSettings
       /// </summary>
       Color ChartDiagramBackColor { get; set; }
 
+      /// <summary>
+      ///    When dragging and dropping observed data folders to a chart
+      ///    assign the same color to curves coming from the same folder
+      /// </summary>
+      bool ColorGroupObservedDataFromSameFolder { get; set; }
```
msevestre, line 62: Type should probably be bool? 
So that you know if it was asked already or not. E.g. you need 3 states

src/OSPSuite.Presentation/Binders/ObservedDataDragDropBinder.cs
```diff
@@ -80,6 +100,38 @@ private IReadOnlyList<ObservedDataNode> observedDataNodesFromClassificationNodes
...
+            });
+         });
+         return observedDataWithFolderAddressCache;
+      }
+
+      //when dragging and dropping just a selection of observed data and not folders, each one should get a different color
+      private static Cache<string, List<DataRepository>> observedDataNodesWithFolderPathFromObservedDataNodes(IEnumerable<ITreeNode> treeNodes)
+      {
+         var observedDataWithFolderAddressCache = new Cache<string, List<DataRepository>>();
+         treeNodes.Each(node =>
```
msevestre, line 124: treeNode.OfType<ObservedDataNode>(). Each....
That way you don't have to cast for type

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -474,6 +482,28 @@ public void SetCurveNameDefinition(Func<DataColumn, string> curveNameDefinition)
          _curveSettingsPresenter.CurveNameDefinition = curveNameDefinition;
       }
 
+      public void AddCurvesWithSameColorForColumn(IReadOnlyList<DataColumn> dataColumnList, CurveOptions defaultCurveOptions = null)
```
msevestre, line 485: That looks a lot like the code below. Can we harmonize?

src/OSPSuite.Presentation/Binders/ObservedDataDragDropBinder.cs
```diff
@@ -24,6 +25,25 @@ public IReadOnlyList<DataRepository> DroppedObservedDataFrom(IDragEvent e)
          return getObservedDataNodesFrom(e).Select(observedDataNode => observedDataNode.Tag.Subject).ToList();
       }
 
+      public Cache<string, List<DataRepository>> DroppedObservedDataWithFolderPathFrom(IDragEvent e)
```
msevestre, line 28: What's with the signature? Why do you need a cache with an Id? What does it represent? How is this being used?
Isn't it really a list of list?

src/OSPSuite.Presentation/Binders/ObservedDataDragDropBinder.cs
```diff
@@ -24,6 +25,25 @@ public IReadOnlyList<DataRepository> DroppedObservedDataFrom(IDragEvent e)
...
+      {
+         var dataNodes = e.Data<IEnumerable<ITreeNode>>();
+
+         if (dataNodes == null)
+            return new Cache<string, List<DataRepository>>();
+
+         var treeNodes = dataNodes as IList<ITreeNode> ?? dataNodes.ToList();
+
+         if (areAllObservedDataNodes(treeNodes))
```
msevestre, line 37: Can you add a comment here? Something like "This is the case when dropping one or more observed data from multi select" etc..

src/OSPSuite.Presentation/Binders/ObservedDataDragDropBinder.cs
```diff
@@ -80,6 +100,38 @@ private IReadOnlyList<ObservedDataNode> observedDataNodesFromClassificationNodes
          return observedDataList;
       }
 
+      private Cache<string, List<DataRepository>> observedDataNodesWithFolderPathFromClassificationNodes(IEnumerable<ITreeNode> treeNodes)
```
msevestre, line 103: oh maybe you need the cache internally right? But it's not being used outside?

src/OSPSuite.Presentation/Binders/ObservedDataDragDropBinder.cs
msevestre, line 28: I am having trouble to see where this method is being used?

src/OSPSuite.Presentation/Binders/ObservedDataDragDropBinder.cs
georgeDaskalakis: Hahahaha, true true. Renamed it to, observedDataNodesFromNodes().

src/OSPSuite.Presentation/IPresentationUserSettings.cs
georgeDaskalakis: After discussing with Pavel we thought that if the color grouping is initially disabled and you get the option to toggle it on and off both through the observed data context manu and the settings, we could skip the dialog. I have aslso updated this in the description of the issue, where thie drag and dropping part of the feature is currently fully documented.

src/OSPSuite.Presentation/Binders/ObservedDataDragDropBinder.cs
georgeDaskalakis: way simpler, thanks :) Corrected

src/OSPSuite.Presentation/Binders/ObservedDataDragDropBinder.cs
georgeDaskalakis: comment added for both cases

src/OSPSuite.Presentation/Binders/ObservedDataDragDropBinder.cs
georgeDaskalakis: In PK-Sim :D The PK-Sim ChartPresenter has a _observedDataDragDropBinder as private member, and responds to a drag and drop action accordingly. I have finished the PK-Sim branch, but I wanted to first have the Core part review finalized to push that PR too.

src/OSPSuite.Presentation/Binders/ObservedDataDragDropBinder.cs
georgeDaskalakis: Tbh I have not changed this part of code (or even the naming underneath for that matter). We were already differentiating between only folders and only data sets, because different methods were being used to retrieve the data already, all that changed was the signature(Cache) and now the naming to better reflect wehat the methods are doing. 

But I totally agree, I also see no reason why not just a  .ToList() is not being used. If you have nothing against this, I will change this to .ToList() only.

georgeDaskalakis: Actually let us better briefly talk about this

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
georgeDaskalakis: refactored it using a new private function

src/OSPSuite.Presentation/Binders/ObservedDataDragDropBinder.cs
georgeDaskalakis: It is the ID of the parent node for internal use, but yes you are right, we do not need to give it as a Cache from the public function. I corrected this

src/OSPSuite.Presentation/IPresentationUserSettings.cs
msevestre: I somewhat disagree with this. Why not tell the user that this is a feature available? This is also what we had discussed

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -474,23 +482,52 @@ public void SetCurveNameDefinition(Func<DataColumn, string> curveNameDefinition)
...
+
+
       public Curve AddCurveForColumn(DataColumn dataColumn, CurveOptions defaultCurveOptions = null)
       {
-         var curve = Chart.CreateCurve(dataColumn.BaseGrid, dataColumn, _curveNameDefinition(dataColumn), _dimensionFactory);
+         var (exists, curve) = createAndConfigureCurve(dataColumn, defaultCurveOptions);
 
-         if (Chart.HasCurve(curve.Id))
-            return Chart.CurveBy(curve.Id);
+         if (exists) return curve;
```
msevestre, line 509: not on one line

msevestre (commented):
I am going to merge this.
I am in favor however of implementing it as we had discussed initially. This has the advantage that the users are made aware of this new feature. Otherwise, it will be hidden

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
msevestre: @georgeDaskalakis you can change this in the next PR

src/OSPSuite.Presentation/Binders/ObservedDataDragDropBinder.cs
```diff
@@ -24,6 +25,26 @@ public IReadOnlyList<DataRepository> DroppedObservedDataFrom(IDragEvent e)
          return getObservedDataNodesFrom(e).Select(observedDataNode => observedDataNode.Tag.Subject).ToList();
       }
 
+      public List<List<DataRepository>> DroppedObservedDataWithFolderPathFrom(IDragEvent e)
```
msevestre, line 28: @georgeDaskalakis I am still confused as to where this is used?DroppedObservedDataWithFolderPathFrom
Why do you need List? IReadOnly is probably even better from an API point of view

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
georgeDaskalakis: done

PR #1365, opened 2021-11-03
Using FindUnit instead of HasUnit

### LINKED ISSUE #1364: Bug in Import Observed Data: "Import unit from a column" is case sensitive
@sfrechen commented on [Fri Oct 29 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1931)

Should be case insensitive.
Currently, an error is thrown in case of `µg/L` that this is an unknown unit.

---

@Yuri05 commented on [Fri Oct 29 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1931#issuecomment-954972870)

OSP9 or OSP10? in 10 this should be fixed: https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1298

---

@msevestre commented on [Mon Nov 01 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1931#issuecomment-956333237)

This feature is only in v10. So it's a bug

---

@msevestre commented on [Mon Nov 01 2021](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/1931#issuecomment-956333433)

@georgeDaskalakis and @abdelr can you check this out?

abdelr:
Fixes #1364 (Import unit from a column" should be case insensitive)

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
@@ -213,7 +213,8 @@ namespace OSPSuite.Infrastructure.Import.Core
                   //if unit comes from a column
                   if (column.Key.Column.Dimension == null)
                   {
-                     var dimensionOfFirstUnit = columnInfo.SupportedDimensions.FirstOrDefault(x => x.HasUnit(column.Value.ElementAt(0).Unit));
+                     var firstNonEmptyUnit = column.Value.FirstOrDefault(x => !string.IsNullOrEmpty(x.Unit));
+                     var dimensionOfFirstUnit = columnInfo.SupportedDimensions.FirstOrDefault(x => x.FindUnit(firstNonEmptyUnit.Unit, true) != null);
 
                      for (var i = 0; i < column.Value.Count(); i++)
                      {
@@ -221,11 +222,11 @@ namespace OSPSuite.Infrastructure.Import.Core
                            continue;
 
                         //if the unit specified does not belong to one of the supported dimensions of the mapping
-                        if (!columnInfo.SupportedDimensions.Any(x => x.HasUnit(column.Value.ElementAt(i).Unit)))
+                        if (!columnInfo.SupportedDimensions.Any(x => x.FindUnit(column.Value.ElementAt(i).Unit, true) != null))
                            throw new InvalidDimensionException(column.Value.ElementAt(i).Unit, columnInfo.DisplayName);
 
                         //if the unit specified is not of the same dimension as the other units of the same data set
-                        if (columnInfo.SupportedDimensions.First(x => x.HasUnit(column.Value.ElementAt(i).Unit)) != dimensionOfFirstUnit)
+                        if (columnInfo.SupportedDimensions.First(x => x.FindUnit(column.Value.ElementAt(i).Unit, true) != null) != dimensionOfFirstUnit)
                            throw new InconsistentDimensionBetweenUnitsException(columnInfo.DisplayName);
                      }
                   }
--- a/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
@@ -174,8 +174,9 @@ namespace OSPSuite.Infrastructure.Import.Core
                            if (double.IsNaN(errorColumn.Value.ElementAt(i).Measurement))
                               continue;
 
-                           if (column.SupportedDimensions.FirstOrDefault(x => x.HasUnit(measurementColumn.Value.ElementAt(i).Unit)) !=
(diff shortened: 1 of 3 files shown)


src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -213,19 +213,20 @@ private void validateUnitsSupportedAndSameDimension(IReadOnlyList<ColumnInfo> co
                   //if unit comes from a column
                   if (column.Key.Column.Dimension == null)
                   {
-                     var dimensionOfFirstUnit = columnInfo.SupportedDimensions.FirstOrDefault(x => x.HasUnit(column.Value.ElementAt(0).Unit));
+                     var firstNonEmptyUnit = column.Value.FirstOrDefault(x => !string.IsNullOrEmpty(x.Unit));
+                     var dimensionOfFirstUnit = columnInfo.SupportedDimensions.FirstOrDefault(x => x.FindUnit(firstNonEmptyUnit.Unit, true) != null);
```
msevestre, line 217: can you add , ignoreCase: true instead of true? It makes the code much easier to understand
see https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Domain/UnitSystem/DimensionFactory.cs#L125 in code base

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -213,19 +213,20 @@ private void validateUnitsSupportedAndSameDimension(IReadOnlyList<ColumnInfo> co
...
 
                      for (var i = 0; i < column.Value.Count(); i++)
                      {
                         if (double.IsNaN(column.Value.ElementAt(i).Measurement))
                            continue;
 
                         //if the unit specified does not belong to one of the supported dimensions of the mapping
-                        if (!columnInfo.SupportedDimensions.Any(x => x.HasUnit(column.Value.ElementAt(i).Unit)))
+                        if (!columnInfo.SupportedDimensions.Any(x => x.FindUnit(column.Value.ElementAt(i).Unit, true) != null))
```
msevestre, line 225: Since you are checkign if there is one and then retrieving it, what about you do something like this
var dimension = columnInfo.SupportedDimensions.FirstOrDefault(x => x.FindUnit(column.Value.ElementAt(i).Unit, ignoreCase: true)
if(dimension==null)
....
and later
if(dimension!=dimensionOfFirstUnit)

also column.Value.ElementAt(i) could be defined as a variable, for readability

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
```diff
@@ -68,9 +68,9 @@ public DataSetToDataRepositoryMappingResult ConvertImportDataSet(ImportedDataSet
       private bool convertParsedDataColumnAndReturnWarningFlag(DataRepository dataRepository, KeyValuePair<ExtendedColumn, IList<SimulationPoint>> columnAndData, string fileName)
       {
          DataColumn dataColumn;
-         var unit = columnAndData.Value.First().Unit;
+         var unit = columnAndData.Value.FirstOrDefault(x => !string.IsNullOrEmpty(x.Unit)).Unit;
          var warningFlag = false;
-         var dimension = columnAndData.Key.Column.Dimension ?? columnAndData.Key.ColumnInfo.SupportedDimensions.FirstOrDefault(x => x.HasUnit(unit));
+         var dimension = columnAndData.Key.Column.Dimension ?? columnAndData.Key.ColumnInfo.SupportedDimensions.FirstOrDefault(x => x.FindUnit(unit, true) != null);
```
msevestre, line 73: same here, SupportsUnit instead of FindUnit . And name the optional parameter

tests/OSPSuite.Infrastructure.Tests/Import/DataSourceSpecs.cs
```diff
@@ -357,6 +357,133 @@ public void geometric_error_does_not_check_units()
...
+         };
+         _dataSet = new DataSet();
+         _dataSet.AddData(new List<ParsedDataSet>() { { new ParsedDataSet(new List<(string, IList<string>)>(), A.Fake<IUnformattedData>(), new List<UnformattedRow>(), parsedData) } });
+         sut.DataSets.Clear();
+
+      }
+
+      protected override void Because()
+      {
+         base.Because();
```
msevestre, line 476: remove base.becasue. What we are testing is always explicit and base.Because should always be empty

msevestre (changes requested):
small code update for readability. Logic seems ok

msevestre:
Also the build is failing

msevestre:
Looking great

PR #1370, opened 2021-11-05
Fixes #1367 try get value

### LINKED ISSUE #1367: Add a method to evaluate the value of the aprameter and return false if it's not possible TryGetValue

msevestre:


msevestre:
@Yuri05 This is to ensure that we do not catch all the time in debug mode in PKSim to speed up the debug process

src/OSPSuite.Core/Domain/Quantity.cs
```diff
@@ -38,6 +40,12 @@ public interface IQuantity : IFormulaUsable, IUsingFormula, IWithValueOrigin
       /// </summary>
       double ValueInDisplayUnit { get; set; }
 
+      /// <summary>
+      ///    The value in the displayed unit
+      /// </summary>
+      bool TryGetValueInDisplayUnit(out double result);
```
Yuri05, line 46: Should `TryGetValueInDisplayUnit `be used **only** for tests? Then: why not defining it as Extension Method?

msevestre: No not for test. I want to use it in real code when the debugger is running, catching the error all the time creates a lot of overhead. It makes working with the tool while debugging very hard (initial concentration formula are all invalid and the whole UI freezes). This is a problem in debug mode only that I wanted to fix for a while. Now the pain is too big so it's required

msevestre: But we can define it as extension method. I don't see a problem

msevestre: This will be used in ParameterDTO in pksim

Yuri05: ok, if not for tests onle, then it's OK I think

Yuri05 (approved):

msevestre:
Well it did not improve performance so I'll revert this and implement a fix in pksim directly

PR #1381, opened 2021-11-17
naming fix

georgeDaskalakis:


georgeDaskalakis:
converted to draft mainly to investigate the possibility of writing a unit test for this. Also to make 100% sure that we need the molecule in the groupByParameters at all.

Yuri05 (approved):

msevestre:
@georgeDaskalakis tests are broken

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -173,11 +173,12 @@ protected virtual void ExtractGeneralParameters(List<string> keys, IUnformattedD
       public IEnumerable<ParsedDataSet> Parse(IUnformattedData data,
          IReadOnlyList<ColumnInfo> columnInfos)
       {
-         var groupByParams =
+         var groupingCriteria =
             Parameters
-               .Where(p => p is GroupByDataFormatParameter || p is MetaDataFormatParameter)
+               .Where(p => p is GroupByDataFormatParameter || p is MetaDataFormatParameter && p.ColumnName != null)//this last check makes sure that a MetaDataFormatParameter with manual input (eg molecule)
```
msevestre, line 178: put the comment above the line.. I don't want to have to scroll to read the command that makes the come clearer

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -173,11 +173,12 @@ protected virtual void ExtractGeneralParameters(List<string> keys, IUnformattedD
...
+                                                                                                                   // does not get added to the groupByParams if it is not set
                .Select(p => (p.ColumnName, p is GroupByDataFormatParameter || (p as MetaDataFormatParameter).IsColumn ? (data.GetColumnDescription(p.ColumnName).ExistingValues) as IList<string> : new List<string>() { p.ColumnName }));
```
msevestre, line 180: I think this line deserves a rewrite.....

msevestre (changes requested):

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
georgeDaskalakis: done

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
georgeDaskalakis: I spread it out to multiple lines, I think it way more understandable now.

georgeDaskalakis:
Just as a note: on the other PR I am going to commit I have fixed the issue with the null ColumnName in a MetaData of the configuration, so that the old tests would still work even without specifying the molecule.

Still I sugest we do keep the molecule specified in the tests here, since it does make more sense.

msevestre:
Going to merge this

I think that
```
 .Where(p => p is GroupByDataFormatParameter || p is MetaDataFormatParameter && p.ColumnName != null)
               .Select(p =>
                  (p.ColumnName,
                     p is GroupByDataFormatParameter || (p as MetaDataFormatParameter).IsColumn
                        ? (data.GetColumnDescription(p.ColumnName).ExistingValues) as IList<string>
```

Is still way too complicated. Something is missing here, and I believe it is a the P level (whatever that is). Tell don't ask. Where we keep asking are you a parameter, are you a formatted, do you have a column.. That code knows way too much of the internal and it should be abstracted in simple methods with a clear name

msevestre:
Thsi castign allt he time as IList xx is also a big no no. And a clear break of encapsulation. you should NOT know that this is an IList and therefore, I am going to wait because this should be fixed before merging.

georgeDaskalakis:
@msevestre you are right about the casting, the reason we are supplying an ReadOnly to the outside is that it should actually stay a ReadOnly, Fixed this

georgeDaskalakis:
@msevestre about the .Where(...).Select(...) expression, I am not sure I understand what you mean about the p level. I grant you it is complicated, mainly I think because we are mixing elements that are of different type and should behave wildly different ( GroupByDataFormatParameter,  MetaDataFormatParameter, with manual input, without manual input). Lets discuss this briefly in the meeting tomorrow.

msevestre:
@georgeDaskalakis There is a conflict. can you fix this? then I'll merge

georgeDaskalakis:
@msevestre I have created abstract member functions in the base class to help identify if the subclasses fulfill certain necessary criteria. Could you please give me a quick feedback on this last commit before merging?

msevestre (approved):
I think this is so much better already. Well done. Where we should discuss , is usafe of methods vs operty get.. I think all of these could be {get;} but this is more of a personal pref. Merging

PR #1382, opened 2021-11-18
1374 create configuration for should store dimension

### LINKED ISSUE #1374: CreateConfigurationFor(string dataPath, string sheetName = null) does not store dimension of columns
[CompiledDataSet_oneSheet.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/7504241/CompiledDataSet_oneSheet.xlsx)

If creating a configuration for this uploaded sheet using `CreateConfigurationFor` and saving the configuration, the columns for "Time", "Measurement", and  "Error" do not contain the `dimension` property:

```
    <MappingDataFormatParameter columnName="Concentration (mass)[ng/ml]">
      <MappedColumn name="Measurement">
        <Unit selectedUnit="ng/ml" />
      </MappedColumn>
    </MappingDataFormatParameter>
```

If the same file is loaded in PK-Sim and the configuration is saved, the dimension property is set:

```
    <MappingDataFormatParameter columnName="Concentration (mass)[ng/ml]">
      <MappedColumn name="Measurement" dimension="Concentration (mass)">
        <Unit selectedUnit="ng/ml" />
      </MappedColumn>
    </MappingDataFormatParameter>
```

abdelr:
Fixes #1374 (CreateConfigurationFor should store dimension)


src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -47,9 +47,45 @@ private double setParameters(IUnformattedData data, IReadOnlyList<ColumnInfo> co
...
+      private void setDimensionsForMappings(IReadOnlyList<ColumnInfo> columnInfos)
+      {
+         foreach (var parameter in _parameters.OfType<MappingDataFormatParameter>())
+         {
+            var mappingColumn = parameter.MappedColumn;
+
+            if (mappingColumn?.Unit == null || mappingColumn.Dimension != null)
+               continue;
+
+            var concreteColumnInfo = columnInfos.FirstOrDefault(x => x.Name == parameter.MappedColumn.Name);
```
msevestre, line 63: used mappingColumn instead of parameter.MappedColumn

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -47,9 +47,45 @@ private double setParameters(IUnformattedData data, IReadOnlyList<ColumnInfo> co
...
          return totalRank;
       }
 
+      private void setDimensionsForMappings(IReadOnlyList<ColumnInfo> columnInfos)
+      {
+         foreach (var parameter in _parameters.OfType<MappingDataFormatParameter>())
+         {
+            var mappingColumn = parameter.MappedColumn;
+
+            if (mappingColumn?.Unit == null || mappingColumn.Dimension != null)
```
msevestre, line 60: why mappingColumn? in the first part of the if but not in the second? this is not logical. Either you fear that mappingColumn is null or you don't fear. But it should always be the same behavior

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
msevestre, line 63: FirstOrDefault means thius cam be null. Or you do not check for null. Therefor either a check is missing or it should be First

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -47,9 +47,45 @@ private double setParameters(IUnformattedData data, IReadOnlyList<ColumnInfo> co
...
+               continue;
+
+            var concreteColumnInfo = columnInfos.FirstOrDefault(x => x.Name == parameter.MappedColumn.Name);
+            //initial settings for fraction dimension
+            if (concreteColumnInfo.DefaultDimension?.Name == Constants.Dimension.FRACTION &&
+                mappingColumn.Unit.ColumnName.IsNullOrEmpty() &&
+                mappingColumn.Unit.SelectedUnit == UnitDescription.InvalidUnit)
+            {
+               mappingColumn.Dimension = concreteColumnInfo.DefaultDimension;
+               mappingColumn.Unit = new UnitDescription(mappingColumn.Dimension.BaseUnit.Name);
```
msevestre, line 70: Is this supposed to be the baseUnit or the DefaultUnit? We also have a DEfaultUnitName property directly

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -109,53 +109,9 @@ private void setDataFormat(IList<DataFormatParameter> formatParameters)
          View.SetMappingSource(_mappings);
          ValidateMapping();
          InitializeErrorUnit();
-         setDimensionsForMappings();
+         //setDimensionsForMappings();
```
msevestre, line 112: remove

msevestre (changes requested):

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
abdelr: If mappingColumn (I have just renamed it to mappedColumn, I think it makes more sense) is null, then the first part will be true and then the second part will not be evaluated. But I agree it is not fully logic to have it only on the first part. I will add it to the second part even though it is totally useless.

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
abdelr: True, I changed it to First since null will mean the configuration is wrong and it should actually throw.

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
abdelr: Right, DefaultUnitName should be the one.

msevestre (commented):
ok

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -60,7 +60,7 @@ private void setDimensionsForMappings(IReadOnlyList<ColumnInfo> columnInfos)
             if (mappedColumn?.Unit == null || mappedColumn?.Dimension != null)
                continue;
 
-            var concreteColumnInfo = columnInfos.First(x => x.Name == mappedColumn.Name);
+            var concreteColumnInfo = columnInfos.First(x => x.DisplayName == mappedColumn.Name);
```
msevestre, line 63: What's the displayName of a column Info? Do we have instances where it's not the same as Name? I think this makes sense to use DisplayName but I don't know why we have this in the first place

PR #1392, opened 2021-12-01
1378 color group by metadata

### LINKED ISSUE #1378: Color group curves in chart according to their metadata
In order to further extend the color grouping functionality for the charts, the following new feature is suggested:

A new button will be introduced in the Chart Editor called “Color Group Curves”. 

image

Clicking the button opens the following dialog where all the common metadata of the observed data shown in the curve are shown.
 
image 

The user can select one or more metadata, according to which the lines will be grouped and given the same color.  Although the colors assigned to the specific lines should be saved, the whole configuration will not be saved, but will only be applied once.

### LINKED ISSUE #1386: Color Grouping in Chart: Tab instead of button
For the new color grouping feature described here:

https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1378

the UI element should be a tab instead of a button, in order to keep things in the UI more consistent. The tab should contain the same list of checkboxes and an Apply button that would work exactly in the same way as the "OK" of the dialog. The selected filters will not be saved or serialized in any way.

georgeDaskalakis:
closes #1378 
closes #1386


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1244,6 +1244,12 @@ public static class FontAndSizeSettings
             public static readonly string FontSizeWatermark = "Font Size Watermark";
             public static readonly string IncludeOriginData = "Include Origin Data";
          }
+
+         public static class ColorGrouping
+         {
+            public static string ColorGroupingDialogDescription = "Choose the metadata according to which the data will be grouped to the same color";
```
msevestre, line 1250: First interation
"Select metadata used to group data by color:

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -240,15 +247,80 @@ public class ChartEditorPresenter : AbstractCommandCollectorPresenter<IChartEdit
...
          _view.SetDataBrowserView(dataBrowserPresenter.View);
       }
 
+      //gets all the common metaData of the observed data that correspond to active curves 
+      private IReadOnlyList<string> getCommonMetaDataOfCurves(IReadOnlyList<DataRepository> allObservedData)
+      {
+         var activeDataColumns = _dataBrowserPresenter.GetAllUsedDataColumns();
+         _activeObservedData = allObservedData.Where(x => activeDataColumns.Any(column => column.Repository.Name == x.Name)).ToList();
+
+         if (_activeObservedData == null || !_activeObservedData.Any())
```
msevestre, line 268: _activeObservedData  cannot be null

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -192,11 +196,13 @@ public class ChartEditorPresenter : AbstractCommandCollectorPresenter<IChartEdit
       private readonly IChartSettingsPresenter _chartSettingsPresenter;
       private readonly IChartExportSettingsPresenter _chartExportSettingsPresenter;
       private readonly ICurveSettingsPresenter _curveSettingsPresenter;
+      private readonly ICurveColorGroupingPresenter _curveColorGroupingPresenter;
       private readonly IDataBrowserPresenter _dataBrowserPresenter;
       private readonly IChartTemplateMenuPresenter _chartTemplateMenuPresenter;
       private readonly IChartUpdater _chartUpdater;
       private readonly IEventPublisher _eventPublisher;
       private readonly IDimensionFactory _dimensionFactory;
+      private List<DataRepository> _activeObservedData;
```
msevestre, line 205: I don't understand what the active observed Data is?

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -240,15 +247,80 @@ public class ChartEditorPresenter : AbstractCommandCollectorPresenter<IChartEdit
...
+         {
+            var tempGroupedList = new List<List<DataRepository>>();
+            foreach (var existingGroup in groupedDataRepositories)
+            {
+               var dataReposGroupedBySingleMetaData =
+                  existingGroup.GroupBy(x => x.ExtendedProperties[groupingMetaData].ValueAsObject).Select(group => @group.ToList());
+               tempGroupedList.AddRange(dataReposGroupedBySingleMetaData);
+            }
+
+            groupedDataRepositories = tempGroupedList;
```
msevestre, line 313: I don't quite understand this. Aren't you overriding this list EVERY time you itereate int eh first for Loop?

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -240,15 +247,80 @@ public class ChartEditorPresenter : AbstractCommandCollectorPresenter<IChartEdit
...
+
+         var firstCurveMetaData = _activeObservedData.First().ExtendedProperties.Keys.ToList();
+         var commonMetaData = firstCurveMetaData.Where(x => _activeObservedData.All(observedData => observedData.ExtendedProperties.Keys.Contains(x))).ToList();
+
+         return commonMetaData;
+      }
+
+      private void onApplyColorGrouping(IReadOnlyList<string> eSelectedMetaData)
+      {
+         if (_activeObservedData == null)
```
msevestre, line 279: If you defined this list as a readonly list, yopu never have to check for null. AT most you clear, and you AddRange when you update it. Last, call it _activeOBservedDataList because it looks like it's only one from the naming

msevestre (changes requested):

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
georgeDaskalakis: True, I changed this

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
georgeDaskalakis: made it into a readonlylist and removed this

src/OSPSuite.Assets/UIConstants.cs
georgeDaskalakis: what about "Select metadata as criteria for color grouping" ?

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
georgeDaskalakis: Yeah I think this is worth some discussing. This is was my best idea algorithmically to take advantage of .GroupBy() to group by multiple criteria:

I group by the first criterion and I get a IEnumerable<List<DataRepository>> containing the groups. If I were to apply .GroupBy() again to each group I would get IEnumerable<IEnumerable<List<DataRepository>>> and so on. So at the end of each grouping I need to "flatten" the structure, since we only care that they are separate groups, we do not really care what the "parent" group is. To do that I am using this temp list.

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
georgeDaskalakis: this is a list that contains all the observed data that correspond to curves active in the chart. This is only in case the chart has metadata, and it initially gets set at the initialization of the PK-Sim ChartPresenter ( through SetAvailableColorGroupingMetaData(), that gets called from PK-Sim). This is done because the ChartPresenter in Core has no knowledge of the datarepositories corresponding to the curves.

msevestre: hum...but curve as xData and yData with a REpositoru method in it to access the repo...

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
msevestre: lol I still don't get it. And I am trying. I think I need to see it in action. I am going to get the code and run it.
How can I get the code to activate in the test program?

msevestre:
@georgeDaskalakis Can we chat about this when you have time. I think I understand what is going on now but I am wondering why we need this activeList at all. The ChardEditorPresenter should be able to get the activeRepository from one of it's subpresenter isntead of holdign a refernvce that WILL get out of date as I do not see the code where it's being updated when I select deselect a curve etc..

msevestre:
I am available until 8pm European time today or after 10pm

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -120,7 +120,7 @@ public interface IChartEditorPresenter : IPresenter<IChartEditorView>,
       /// <summary>
       ///    Adds a curves to the chart for all dataColumns in <paramref name="dataColumnList" /> if they do not already exist.
       ///    Curve options will be applied if they are specified and a new curve is created.All the created curves will have
-      ///    the same color. 
+      ///    the same color.
```
msevestre, line 123: @georgeDaskalakis Please review my changes and if you agree please merge.
I would love if you could explain the algorithm IN the code for grouping as we discussed on the call yesterday because it is not obvious

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -278,20 +283,20 @@ private void assignSameColorToGroupedCurves(List<IReadOnlyList<DataRepository>>
...
          foreach (var group in groupedDataRepositories)
          {
-            var groupColor = Chart.SelectNewColor();
             var curvesInGroup = findCurvesCorrespondingToRepositoryList(group);
 
             if (!curvesInGroup.Any())
                continue;
-            
+
+            var groupColor = Chart.SelectNewColor();
```
msevestre, line 291: I think the color should be used only if there are curves. Otherwise you will skip a color

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -434,6 +440,9 @@ private void updateUsedColumns()
       {
          if (Chart == null) return;
          _dataBrowserPresenter.InitializeIsUsedForDataColumns(Chart.UsedColumns);
+
+         //Once used columns are updated, we need to make sure we update the meta data related to used column as well
+         refreshColorGroupingPresenter();
```
msevestre, line 445: I have created a method here. When we changed the selected columns, we need to refresh somehow

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -636,11 +644,15 @@ public void Handle(ChartUpdatedEvent chartUpdatedEvent)
          if (!canHandle(chartUpdatedEvent))
             return;
 
-         _curveColorGroupingPresenter.SetMetadata(getCommonMetaDataOfCurves());
```
msevestre: this should be called in the refresh I believe. Why do it BEFORE refreshing all the other presenters with data. It should be done last.. Since updateUsedColumns is called last, it should be good

msevestre (commented):
@georgeDaskalakis Please let me know. 
Once you have added the comment for the algo and if you agree with my changes let me know and I will merge it (or if you do want to merge it, please use SQUASH commit )

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
georgeDaskalakis: agree

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
georgeDaskalakis: thanks for the changes @msevestre ! I totally agree with them. Sorry for taking so long to respond, I did not notice your comments at all - I thought this PR was over and did not have it in  mind. I will now merge with squash.

PR #1398, opened 2021-12-10
Fixes #1397 make is change by create individual writable

### LINKED ISSUE #1397: Make IsCreatedByIndividual writable as it needs to be changed via code

msevestre:


src/OSPSuite.Core/Domain/ParameterInfo.cs
```diff
@@ -12,6 +12,7 @@ public enum ParameterFlag
       MinIsAllowed = 2 << 3,
       MaxIsAllowed = 2 << 4,
       CanBeVariedInPopulation = 2 << 5,
+      IsChangedByCreateIndividual = 2 << 6,
```
msevestre, line 15: @Yuri05 Some disease state  algo are varying parameters that are not ohter standard parameters. We need to update this dynamically instead of the way it is done now. I will add a new column in the DB as well to read the information from the DB instead of the weird way we are doing it now

msevestre (commented):
I will merge as soon as the build is ready

Yuri05 (approved):

PR #1400, opened 2021-12-15
#1396 multiple curve options

### LINKED ISSUE #1396: Change Options of multiple curves in Chart
Editing options of multiple curves

In the “Curves and Axis Options” Tab, a new context menu item will be added called “Edit options for selected”. It will only be visible when the user has selected at least two rows.
 
image

This option will open a new dialog that lists common option of the curves. Initially the values for the options are set to "<Current value>"–  meaning that the selected curves will retain for that option the value they had before opening the dialog.

When another value is selected though, all the options of the selected rows will be set to this value.
 
image

georgeDaskalakis:
closes #1396


src/OSPSuite.Presentation/Presenters/Charts/CurveMultiItemEditorPresenter.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System.Drawing;
+using OSPSuite.Core.Chart;
+using OSPSuite.Presentation.Views.Charts;
+
+namespace OSPSuite.Presentation.Presenters.Charts
+{
+   public enum YesNoValues
```
georgeDaskalakis, line 7: @msevestre no particularly proud of this solution here....still I preferred Yes/No instead of true/false and the enum keeps things consistent with the other comboboxes.

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1253,6 +1255,16 @@ public static class ColorGrouping
             public static string ColorGroupingDialogDescription = "Select metadata as criteria for color grouping";
             public static string ApplyColorGroupingButton = "Apply Color Grouping";
          }
+
+         public static class MultiCurveOptions
+         {
+            public static string CurrentValue = "<Current value>";
+            public static string Color = "Color";
```
msevestre, line 1262: We have those constants somewhere. Please reuse. They clearly mean the same

src/OSPSuite.Presentation/Presenters/Charts/CurveMultiItemEditorPresenter.cs
```diff
@@ -0,0 +1,50 @@
...
+      void Show();
+      bool Canceled();
+      SelectedCurveValues GetSelectedValues();
+   }
+
+   public class SelectedCurveValues
+   {
+      public Color? Color { get; set; }
+      public string Style { get; set; }
+      public string Symbol { get; set; }
```
msevestre, line 24: Why is everything designed as string?we have specific enums etc..this is very suspicious

src/OSPSuite.Presentation/Presenters/Charts/CurveSettingsPresenter.cs
```diff
@@ -203,5 +208,32 @@ public void UpdateColorForCurve(Curve curve, Color color)
...
+            multiEditorPresenter.Show();
+
+            if (multiEditorPresenter.Canceled())
+               return;
+
+            SelectedCurveValues selectedValues = multiEditorPresenter.GetSelectedValues();
+            foreach (var curveDTO in selectedCurveDTOs)
+            {
+               if (selectedValues.Color != null) curveDTO.Color = (Color)selectedValues.Color;
+               if (Enum.TryParse(selectedValues.Style, out LineStyles selectedStyle))
```
msevestre, line 225: Lol..yes this was very suspicious. 
Just set the value using real types (maybe nullable) instead or all those cast. 
Unless there is something that I don't understand? In that case I need some explanations

src/OSPSuite.Presentation/Presenters/Charts/CurveSettingsPresenter.cs
```diff
@@ -203,5 +208,32 @@ public void UpdateColorForCurve(Curve curve, Color color)
          var curveDTO = findCurveDTOFor(curve);
          UpdateCurveColor(curveDTO, color);
       }
+
+      public void EditProperties(List<CurveDTO> selectedCurveDTOs)
+      {
+         using (var multiEditorPresenter = _applicationController.Start<ICurveMultiItemEditorPresenter>())
+         {
+            multiEditorPresenter.Show();
```
msevestre, line 216: Just encapsulate this behaviour. You are calling 3 presenter methods in a row. 
Call only one and put the logic in the presenter

src/OSPSuite.Presentation/Presenters/Charts/CurveMultiItemEditorPresenter.cs
msevestre, line 7: Why do we need this? Why is our yes no not enough?

msevestre (changes requested):
Lots of things that I do not understand:
* Usage of a new enum type for yes/no when we have our types already
* Usage of string as type for everything! From boolean to enums.

Let's chat

src/OSPSuite.Assets/UIConstants.cs
georgeDaskalakis: done

src/OSPSuite.Presentation/Presenters/Charts/CurveMultiItemEditorPresenter.cs
georgeDaskalakis: @msevestre let us briefly talk about this...I used to have them as nullable enum, but was not very fond of that behaviour either.

georgeDaskalakis:
@msevestre I changed this and implemented with the screenBinder as we discussed. Please take a look of the changes and if you have no objections feel free to merge :)

src/OSPSuite.Presentation/Presenters/Charts/CurveMultiItemEditorPresenter.cs
```diff
@@ -0,0 +1,57 @@
...
+      public Symbols? Symbol { get; set; }
+      public bool? Visible { get; set; }
+      public bool? VisibleInLegend { get; set; }
+      public bool IsColorSet { get; set; } = false;
+   }
+
+   public class CurveMultiItemEditorPresenter : AbstractDisposablePresenter<ICurveMultiItemEditorView, ICurveMultiItemEditorPresenter>,
+      ICurveMultiItemEditorPresenter
+   {
+      private SelectedCurveValues _selectedCurveValues;
```
msevestre, line 33: Run resharper on this file. It's clearly not formatter properly. Let's run resharper on all files before we submit a PR

src/OSPSuite.Presentation/Formatters/MultiCurveOptionsFormatter.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace OSPSuite.Presentation.Formatters
+{
+   public class BooleanYesNoFormatter : IFormatter<bool?>
+   {
+      public string Format(bool? valueToFormat)
+      {
+         if (valueToFormat == null)
+            return Captions.Chart.MultiCurveOptions.CurrentValue;
+
+         return  valueToFormat.Value ? "Yes" : "No";
```
msevestre, line 14: @georgeDaskalakis no need to cast. use value

src/OSPSuite.Presentation/Presenters/Charts/CurveMultiItemEditorPresenter.cs
```diff
@@ -0,0 +1,59 @@
...
+      private readonly SelectedCurveValues _selectedCurveValues;
+
+      public CurveMultiItemEditorPresenter(ICurveMultiItemEditorView view) : base(view)
+      {
+         _selectedCurveValues = new SelectedCurveValues();
+      }
+
+      public IEnumerable<bool?> AllBooleanOptions { get; } = Constants.MultiCurveOptions.AllBooleanOptions;
+
+      public IEnumerable<LineStyles?> AllLineStyles { get; } = new List<LineStyles?>() {null}.Union(EnumHelper.AllValuesFor<LineStyles>().Cast<LineStyles?>());
```
msevestre, line 42: no need to evaluate the propertie all the time. Define a single property

src/OSPSuite.Presentation/Presenters/Charts/CurveMultiItemEditorPresenter.cs
```diff
@@ -0,0 +1,59 @@
...
+
+      public IEnumerable<Symbols?> AllSymbols { get; } = new List<Symbols?>() {null}.Union(EnumHelper.AllValuesFor<Symbols>().Cast<Symbols?>());
+
+      public void SetColorChangedFlag()
+      {
+         _selectedCurveValues.IsColorSet = true;
```
msevestre, line 48: I don't get this IsColorSet property. Why not Color?

src/OSPSuite.Presentation/Presenters/Charts/CurveSettingsPresenter.cs
```diff
@@ -203,5 +208,26 @@ public void UpdateColorForCurve(Curve curve, Color color)
...
+         using (var multiEditorPresenter = _applicationController.Start<ICurveMultiItemEditorPresenter>())
+         {
+            var selectedValues = multiEditorPresenter.GetSelectedValues();
+
+            foreach (var curveDTO in selectedCurveDTOs)
+            {
+               if (selectedValues.IsColorSet)
+                  curveDTO.Color = selectedValues.Color;
+
+               curveDTO.LineStyle = selectedValues.Style.GetValueOrDefault(curveDTO.LineStyle);
```
msevestre, line 223: use the properties of nullable

src/OSPSuite.UI/Views/Charts/CurveMultiItemEditorView.cs
```diff
@@ -0,0 +1,84 @@
...
+
+namespace OSPSuite.UI.Views.Charts
+{
+   public partial class CurveMultiItemEditorView : BaseModalView, ICurveMultiItemEditorView
+   {
+      private ICurveMultiItemEditorPresenter _presenter;
+      private readonly IFormatter<bool?> _boolFormatter = new BooleanYesNoFormatter();
+      private readonly IFormatter<LineStyles?> _lineStylesFormatter = new LineStylesFormatter();
+      private readonly IFormatter<Symbols?> _symbolsFormatter = new SymbolsFormatter();
+      private readonly ScreenBinder<SelectedCurveValues> _screenBinder = new ScreenBinder<SelectedCurveValues> {BindingMode = BindingMode.TwoWay};
```
msevestre, line 20: define as readonly in cosntructor or direct init

src/OSPSuite.UI/Views/Charts/CurveMultiItemEditorView.Designer.cs
```diff
@@ -0,0 +1,266 @@
...
+      /// Clean up any resources being used.
+      /// </summary>
+      /// <param name="disposing">true if managed resources should be disposed; otherwise, false.</param>
+      protected override void Dispose(bool disposing)
+      {
+         if (disposing && (components != null))
+         {
+            components.Dispose();
+         }
+         _screenBinder.Dispose();
```
msevestre, line 22: Very important to dispoise of binder otherwise you have a leak

src/OSPSuite.Presentation/Presenters/Charts/CurveMultiItemEditorPresenter.cs
```diff
@@ -0,0 +1,59 @@
...
+   }
+
+   public class SelectedCurveValues
+   {
+      public Color Color { get; set; }
+      public LineStyles? Style { get; set; }
+      public Symbols? Symbol { get; set; }
+      public bool? Visible { get; set; }
+      public bool? VisibleInLegend { get; set; }
+      public bool IsColorSet { get; set; } = false;
```
msevestre, line 27: @georgeDaskalakis I do not understand this flag

msevestre (commented):
I have updated the code to match the coding standard and ran resharper. Let's try to do it before submitting the PR
Also: Why do we need the IsColorSet flag? Color? not enough i guess but I dont understand why

src/OSPSuite.Presentation/Presenters/Charts/CurveMultiItemEditorPresenter.cs
georgeDaskalakis: good point, I do go through the changes, but just by looking at the diff it is easy to lose something.

src/OSPSuite.Presentation/Presenters/Charts/CurveMultiItemEditorPresenter.cs
georgeDaskalakis: good point

src/OSPSuite.Presentation/Presenters/Charts/CurveMultiItemEditorPresenter.cs
georgeDaskalakis: so the Color? did not work with binding to the ColorEdit Control. Also not with the formatter, so the best workaround I could think of was to use the flag.

src/OSPSuite.Presentation/Presenters/Charts/CurveMultiItemEditorPresenter.cs
georgeDaskalakis: @msevestre thanks for all the other points and corrections! Again this flag has to do with not being able to bind the Color? property to the ColorEdit Control.

msevestre: I see. Probably we could add support for Color? in DevExpressBinder. Merging for now

PR #1403, opened 2021-12-17
Should be possible to switch between different parameter identification feedback views 

### LINKED ISSUE #1372: Should be possible to switch ParameterIdentification from Feedback view

abdelr:
Fixes #1372 (Should be possible to switch between different parameter identification feedback views )


src/OSPSuite.Presentation/Presenters/ParameterIdentifications/MultipleParameterIdentificationFeedbackPresentersManager.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+   public interface IMultipleParameterIdentificationFeedbackPresentersManager : 
+      IListener<ParameterIdentificationStartedEvent>,
+      IListener<ParameterIdentificationTerminatedEvent>,
+      IListener<ProjectClosedEvent>
+   {
+      IParameterIdentificationFeedbackPresenter FeedbackPresenterFor(ParameterIdentification parameterIdentification);
+   }
+
+   public class MultipleParameterIdentificationFeedbackPresentersManager : IMultipleParameterIdentificationFeedbackPresentersManager
```
msevestre, line 17: Why is it called `MultipleParameterIdentificationFeedbackPresentersManager` and not `ParameterIdentificationFeedbackPresentersManager`

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/MultipleParameterIdentificationFeedbackPresentersManager.cs
```diff
@@ -0,0 +1,52 @@
...
+   {
+      static private IContainer _container;
```
msevestre, line 19: Why is this static? remove

msevestre: and use resharped to create it as readonly etc..

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/MultipleParameterIdentificationFeedbackPresentersManager.cs
```diff
@@ -0,0 +1,52 @@
...
+      private Cache<ParameterIdentification, IParameterIdentificationFeedbackPresenter> _currentParameterIdentifications = new Cache<ParameterIdentification, IParameterIdentificationFeedbackPresenter>();
```
msevestre, line 20: this should also be readonly. Please run resharper to do this stuff for yoy automatically

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/MultipleParameterIdentificationFeedbackPresentersManager.cs
```diff
@@ -0,0 +1,52 @@
...
+
+      public MultipleParameterIdentificationFeedbackPresentersManager(IContainer container)
+      {
+         _container = container;
+      }
+
+      public IParameterIdentificationFeedbackPresenter FeedbackPresenterFor(ParameterIdentification parameterIdentification)
+      {
+         if (_currentParameterIdentifications.Contains(parameterIdentification))
```
msevestre, line 29: I fear that this might be called from multiple threads no? Should it be thread safe?

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationFeedbackPresenter.cs
```diff
@@ -25,13 +25,12 @@ public interface IParameterIdentificationRunFeedbackPresenter : IParameterIdenti
...
    public interface IParameterIdentificationFeedbackPresenter : IPresenter<IParameterIdentificationFeedbackView>,
       IToogleablePresenter,
-      IListener<ParameterIdentificationStartedEvent>,
-      IListener<ParameterIdentificationSelectedEvent>,
-      IListener<ParameterIdentificationTerminatedEvent>,
       IListener<ParameterIdentificationIntermediateResultsUpdatedEvent>,
       IListener<ProjectClosedEvent>
    {
       bool ShouldRefreshFeedback { get; set; }
+      void OnParameterIdentificationStarted(ParameterIdentification parameterIdentification);
```
msevestre, line 32: OnXxx looks like an event handler but it not. Why not just call it ParmaeterIdentificationStarted?

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationFeedbackPresenter.cs
```diff
@@ -142,6 +130,7 @@ public void Handle(ProjectClosedEvent eventToHandle)
          clearFeedbackReferences();
          resetFeedback();
          _view.NoFeedbackAvailable();
+         _view.Hide();
```
msevestre, line 133: Why do you had the view? the user won't see that the PI is finished or even see the optimal results etc... or am i missing something?

msevestre (changes requested):
Looks good. A few questions and reshaper should be run to format the file properly

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationFeedbackPresenter.cs
abdelr: If you close the project, should we keep the view open? To me, closing the project means you are finished with everything on it, including checking if the PI has finished or not. Actually, the references will be cleared meaning you should not be able to keep the view opened, right?

abdelr:
> Looks good. A few questions and reshaper should be run to format the file properly

I have just fixed the problem with resharper so I hope there will not be any further problem with it anymore ;)

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationFeedbackPresenter.cs
msevestre: yeah ok. makes sense. What was the current behaviour? It would just stay open?

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationFeedbackPresentersManager.cs
```diff
@@ -0,0 +1,59 @@
...
+      public void Handle(ParameterIdentificationTerminatedEvent eventToHandle)
+      {
+         FeedbackPresenterFor(eventToHandle.ParameterIdentification).ParameterIdentificationTerminated(eventToHandle.ParameterIdentification);
+      }
+
+      public void Handle(ProjectClosedEvent eventToHandle)
+      {
+         lock (_locker)
+         {
+            _currentParameterIdentifications.Clear();
```
msevestre, line 55: You are holding a reference all the `IParameterIdentificationFeedbackPresenter`s here.
They need to be `ReleaseFrom` explicitly as they are registered in the EventPublisher on construction. This is a memory leak 

so something like 
`_currentParameterIdentifications.Each(x=>x.ReleaseFrom(_eventPublisher));` where _eventPublisher is injected into the constructor of the Manager

src/OSPSuite.Presentation/UICommands/ParameterIdentificationFeedbackViewVisibilityUICommand.cs
```diff
@@ -1,19 +1,28 @@
...
    {
-      private readonly IParameterIdentificationFeedbackPresenter _feedbackPresenter;
+      private readonly IParameterIdentificationFeedbackPresentersManager _multipleParameterIdentificationFeedbackPresentersManager;
+      private ParameterIdentification _parameterIdentification;
 
-      public ParameterIdentificationFeedbackViewVisibilityUICommand(IParameterIdentificationFeedbackPresenter feedbackPresenter)
+      public ParameterIdentificationFeedbackViewVisibilityUICommand(IParameterIdentificationFeedbackPresentersManager multipleParameterIdentificationFeedbackPresentersManager)
       {
-         _feedbackPresenter = feedbackPresenter;
+         _multipleParameterIdentificationFeedbackPresentersManager = multipleParameterIdentificationFeedbackPresentersManager;
```
msevestre, line 15: rename this variable to match type

msevestre (changes requested):
There is a leak that I missed on first pass

msevestre:
The build is failing for some reasons...

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationFeedbackPresentersManager.cs
abdelr: I am wondering if ApplicationController is not doing the work already?

```
      public virtual ISingleStartPresenter Open<TSubject>(TSubject subject, ICommandCollector commandCollector)
      {
         var presenterToOpen = PresenterFor(subject);
         if (presenterToOpen == null)
         {
            presenterToOpen = CreatePresenterForSubject(subject);
            presenterToOpen.Closing += removePresenter;
            presenterToOpen.InitializeWith(commandCollector);
            _allOpenedPresenters.Add(presenterToOpen);
         }
         return presenterToOpen;
      }
```

msevestre: You are not resolving the presenter usting the applicationController but you are using the container. 

But this is in fact a good point. I am wondering if we need this ParameterIdentifiactionManager at all since we already have everything in the applicationController. Why aren't you using this?

msevestre: Yeah the ApplictionController does exactly what you need I feel for almost everything. Let's talk about it during the call later today

abdelr:
@msevestre There was already a ISingleStartPresenter registered to edit a ParameterIdentification so I needed to create a wrapper (ParameterIdentificationFeedback) for the ParameterIdentificationFeedbackView/Presenter. Additionally I needed a manager which will always return the same object for the same ParameterIdentification and add some "registrations". I actually do not like fully the code in MoBiApplicationController with the long if else for resolving a presenter for a subject. I would prefer to use the self aware pattern we discussed recently (we ask the presenters if they know how to handle the subject instead of having a long list with all the bindings, but I do not want to draw my attention now on this direction. Maybe for the future? In any case, I am still solving some unforeseen issues.

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationFeedbackManager.cs
```diff
@@ -0,0 +1,67 @@
...
+         lock (_locker)
+         {
+            _cache.Clear();
+         }
+      }
+
+      public void Handle(ParameterIdentificationStartedEvent eventToHandle)
+      {
+         var parameterIdentification = eventToHandle.ParameterIdentification;
+         lock (_locker)
```
msevestre, line 55: Call GetFeedback  method here (or to follow nomenclature rename it to FeedbackFor) instead of duplicating the code.

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationFeedbackPresenter.cs
```diff
@@ -165,6 +167,57 @@ private void updateFeedback(ParameterIdentificationRunState state)
...
+      }
+
+      public void Handle(RenamedEvent eventToHandle)
+      {
+         //Nothing to do here
+      }
+
+      public void Edit(ParameterIdentificationFeedback objectToEdit)
+      {
+         _key = objectToEdit;
```
msevestre, line 208: I don't understand why you need this. Can't you save the _parameterIndentification as member?

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationFeedbackPresenter.cs
```diff
@@ -165,6 +167,57 @@ private void updateFeedback(ParameterIdentificationRunState state)
...
+         if (objectToEdit.Running)
+            Handle(new ParameterIdentificationStartedEvent(objectToEdit.ParameterIdentification));
```
msevestre, line 210: This is also very weird.  THis is not how the publisher works. Want to have a call?

msevestre (changes requested):
I don't understand why you need this new Object structure. Let's have a chat

abdelr:
I will call you in about 20 minutes.

msevestre (commented):
You mentioned a leak when clsoing. It is still missing right?

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationFeedbackPresenter.cs
```diff
@@ -165,6 +178,67 @@ private void updateFeedback(ParameterIdentificationRunState state)
...
+      {
+         _parameterIdentificationFeedback = objectToEdit;
+
+         switch (objectToEdit.RunStatus)
+         {
+            case RunStatusId.Running:
+               setParameterIdentificationToStarted(objectToEdit.ParameterIdentification);
+               break;
+            case RunStatusId.Canceled:
+               setParameterIdentificationToStarted(objectToEdit.ParameterIdentification);
```
msevestre, line 227: This is weird. Why start and then cancel. I am sure there is a reason but I would comment this in the code

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationFeedbackPresenter.cs
```diff
@@ -73,21 +76,18 @@ protected override void SaveFormLayout(Point location, Size size)
 
       public void Handle(ParameterIdentificationStartedEvent eventToHandle)
       {
-         _currentParameterIdentifications.Add(eventToHandle.ParameterIdentification);
-         _parameterIdentification = eventToHandle.ParameterIdentification;
-         _view.Caption = Captions.ParameterIdentification.FeedbackViewFor(_parameterIdentification.Name);
-         showParameterIdentificationFeedback();
+         if (eventToHandle.ParameterIdentification != _parameterIdentificationFeedback.ParameterIdentification)
```
msevestre, line 79: Much better. You even have a _parameterIdentification member so this could be
if (!Equals(_parameterIdentification , eventToHandle.ParameterIdentification))
  return;

Same later.  on  public void Handle(ParameterIdentificationTerminatedEvent eventToHandle)

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationFeedbackPresenter.cs
```diff
@@ -165,6 +178,67 @@ private void updateFeedback(ParameterIdentificationRunState state)
...
+      {
+         //Nothing to do here
+      }
+
+      public void Handle(RenamedEvent eventToHandle)
+      {
+         //Nothing to do here
+      }
+
+      public void Edit(ParameterIdentificationFeedback objectToEdit)
```
msevestre, line 217: rename objectToEdit to parameterIdentificationFeedback

msevestre (commented):
Minor changes. Looks very good. The onky really required is the comment to explain why you call the start on cancel. The rest is more for consistency purposes. Fix what you see fit and please merge

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationFeedbackPresenter.cs
```diff
@@ -165,6 +178,72 @@ private void updateFeedback(ParameterIdentificationRunState state)
...
+         {
+            case RunStatusId.Running:
+               setParameterIdentificationToStarted(parameterIdentificationFeedback.ParameterIdentification);
+               break;
+            case RunStatusId.Canceled:
+               // Presenter was opened after the events were triggered but actually a terminated
+               // event is triggered after a started event was triggered and the feedback view
+               // should evolve through both of them. If terminated is invoked directly, then the
+               // view would should its initial state asking the user to start the parameter identification
+               // instead of showing the terminated parameter identification and its last state.
```
msevestre, line 231: yes!!

msevestre (approved):

msevestre:
The build if failing @abdelr Can you check?

PR #1413, opened 2021-12-31
Fixes #1395 try parse

### LINKED ISSUE #1395: Use latest funcparser to ensure that we can use TryParse

msevestre:


src/OSPSuite.Core/Domain/Formulas/BlackBoxFormula.cs
```diff
@@ -8,7 +8,7 @@ namespace OSPSuite.Core.Domain.Formulas
    /// </summary>
    public class BlackBoxFormula : Formula
    {
-      protected override double CalculateFor(IEnumerable<IObjectReference> usedObjects, IUsingFormula dependentObject)
+      protected override double CalculateFor(IReadOnlyList<IObjectReference> usedObjects, IUsingFormula dependentObject)
```
msevestre, line 11: I changed this internal interface so that we do not have to always call ToList

src/OSPSuite.Core/Domain/Quantity.cs
```diff
@@ -43,7 +43,7 @@ public interface IQuantity : IFormulaUsable, IUsingFormula, IWithValueOrigin
       /// <summary>
       ///    The value in the displayed unit
       /// </summary>
-      bool TryGetValueInDisplayUnit(out double result);
+      (double value, bool success) TryGetValueInDisplayUnit();
```
msevestre, line 46: This is a better API. Since it was not used before, I have no problem changing it

src/OSPSuite.Core/Services/ExplicitFormulaParser.cs
```diff
@@ -71,6 +71,23 @@ public void Parse()
             throw new FuncParserException($"Unable to parse expression: {ex.Message}", ex);
          }
       }
+
+      public (double value, bool success) TryCompute(double[] variableValues, double[] parameterValues)
+      {
+         var success = _parserFunc.TryParse();
```
msevestre, line 77: maybe we do need a TryCompute in FuncParser...This is good for now I think but this seems to be a bit backward

src/OSPSuite.Core/Domain/Quantity.cs
```diff
@@ -147,19 +147,29 @@ public double ValueInDisplayUnit
...
+
+         var (value, success) = Formula.TryCalculate(this);
+         if (success)
          {
-           //Formula could not be parsed. In that case, we save the value as double.NaN so that next time, the value will be NaN
-           result = double.NaN;
-           return false;
+            _cachedValue = value;
+            //Cached value is only valid if the Formula has updated it's references
+            _cachedValueValid = Formula.AreReferencesResolved;
```
Yuri05, line 163: I forgot all the details about cached value, but: does it make sense to set _cachedValue if _cachedValueValid is false?
In this case _cachedValue will not be used anyway, right ?

src/OSPSuite.Core/Services/ExplicitFormulaParser.cs
Yuri05: Can do. TryCompute would behave exactly as your current C# implementation: call TryParse and return false if TrParse fails; otherwise return true and the result of the calculation (which could also be Nan, Infinity etc.)

src/OSPSuite.Core/Domain/Quantity.cs
msevestre: This is the same code as for Value {set;} . I did not want to change the logic here...You are probably right

msevestre: going to check this out

msevestre: I am going to leave the code as is. Otherwise we need to add some more if checks for nothing.

PR #1415, opened 2022-01-03
Fixes #1414 fixes unit test not running

### LINKED ISSUE #1414: Unit tests not running
It is currently not possible to ru the unit tests because of the removal of the nuget package dependencies in the solution.

msevestre:


msevestre:
@georgeDaskalakis @abdelr 
I am merging this into  develop. I checked. Tests were running locally BEFORE and AFTER the change so I have no way of checking if this fixes your issues
How are you running the tests?

PR #1417, opened 2022-01-04
Fixes #1416 autofac change revert

### LINKED ISSUE #1416: Revert autofac changes creating issue in R

msevestre:


src/OSPSuite.Infrastructure.Autofac/OSPSuite.Infrastructure.Autofac.csproj
```diff
@@ -32,7 +32,7 @@
 
 
   <ItemGroup>
-    <PackageReference Include="Autofac" Version="5.2.0" />
+    <PackageReference Include="Autofac" Version="4.9.4" />
```
msevestre, line 35: @Yuri05 @abdelr @georgeDaskalakis 
As long as we are not working with .NET 5 or higher, we will have those issues where we cannot upgrade. We need to move away from Mono asap. All our problems are due to the fact that we are stuck with .NET standard 2.0

PR #1423, opened 2022-01-06
Importer: Better message and avoid null access exception

### LINKED ISSUE #1410: Importer: Exception on non-monotonical data
When a column mapping change results in the time column being non-monotonical and there are already imported data sets, the following exception is thrown:

image

Also the description of the error (in the red under the column mapping) could be better if it started "please update the mapping..."

abdelr:
Fixes #1410

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -1276,7 +1276,7 @@ namespace OSPSuite.Assets
       public static readonly string UnsupportedFileType = "The type of file that you are trying to open is not currently supported";
       public static readonly string CannotRemoveBaseGridColumnStillInUse = "Cannot remove base grid column still used by other columns";
 
-      public static string ErrorWhenPlottingDataRepository(int sheetName, string exceptionMessage) => $"Plotting data set number:{sheetName} produced the following error: {exceptionMessage}";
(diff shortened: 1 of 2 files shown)


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1276,7 +1276,7 @@ public static class Error
       public static readonly string UnsupportedFileType = "The type of file that you are trying to open is not currently supported";
       public static readonly string CannotRemoveBaseGridColumnStillInUse = "Cannot remove base grid column still used by other columns";
 
-      public static string ErrorWhenPlottingDataRepository(int sheetName, string exceptionMessage) => $"Plotting data set number:{sheetName} produced the following error: {exceptionMessage}";
+      public static string ErrorWhenPlottingDataRepository(int sheetName, string exceptionMessage) => $"It was not possible to plot the data sets. Please, update your mappings to correct it. An error occur while plotting data set number:{sheetName + 1} produced the following error: {exceptionMessage}";
```
msevestre, line 1279: was is it? in update your mapping to correct it?

abdelr: The reason for this error is normally that you are missing a grouping by mapping. So, in 95% of the cases the error is that there is a missing mapping (update your mappings to correct the problem). In the other 5% well, there is a real problem in the data and the time is just not monotonical I would say but I still think it makes sense to draw the attention of the user to the mappings just to check if there are still missing mappings.

msevestre: >Please, update your mappings to correct it. 

I don't think it's english. It needs to be rephrase. What does "it" mean

PR #1434, opened 2022-01-10
Setting dimension with unit

### LINKED ISSUE #1405: New data import cannot handle correctly arithmetic error of a fraction given as percentage
Trying to import the data from the attached xlsx-file: 
1. Choosing "fraction" as dimension and "%" as unit for the DATAvalues - assigned to Measurement
2. Assigning "Arithmetic Standard Deviation" and selecting unit "%" for the Error data column 

After importing, the error bars are too high - cf. attached screenshot.

[test_2.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/7736755/test_2.xlsx)

errorBars

abdelr:
Fixes #1405

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
@@ -181,8 +181,11 @@ namespace OSPSuite.Presentation.Presenters.Importer
          }
          else
          {
-            column.Unit = _mappingParameterEditorPresenter.Unit;  
-            column.Dimension = _mappingParameterEditorPresenter.Dimension;
+            column.Unit = _mappingParameterEditorPresenter.Unit;
+            column.Dimension = _columnInfos
+               .First(x => x.DisplayName == model.MappingName)
+               .SupportedDimensions
+               .FirstOrDefault(x => dimensionContainsUnit(x, column.Unit));
          }
 
          if (model.ColumnInfo.IsBase())
@@ -210,6 +213,11 @@ namespace OSPSuite.Presentation.Presenters.Importer
 
          ValidateMapping();
          _view.CloseEditor();
+      }
+
+      private bool dimensionContainsUnit(IDimension dimension, UnitDescription unit)
+      {
+         return dimension.Units.Select(x => x.Name).Contains(unit.SelectedUnit);
       }
 
       public bool ShouldManualInputOnMetaDataBeEnabled(ColumnMappingDTO model)
--- a/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
@@ -185,7 +185,7 @@ namespace OSPSuite.Presentation.Presenters.Importer
             column.Dimension = _columnInfos
                .First(x => x.DisplayName == model.MappingName)
                .SupportedDimensions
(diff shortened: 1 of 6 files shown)


src/OSPSuite.Presentation/Presenters/Importer/UnitsEditorPresenter.cs
```diff
@@ -18,7 +18,7 @@ public class UnitsEditorPresenter : AbstractDisposablePresenter<IUnitsEditorView
 
       private bool _columnMapping;
 
-      public UnitDescription Unit => new UnitDescription(_selectedUnit, _selectedColumn);
+      public UnitDescription Unit => new UnitDescription(_view.SelectedUnit, _selectedColumn);
```
msevestre, line 21: Why not use _selectedUnit. Why are they out of sync? This smells. Remove the _selectedUnit if it's not used

src/OSPSuite.UI/Views/Importer/UnitsEditorView.cs
```diff
@@ -36,6 +36,11 @@ public UnitsEditorView()
          _columnComboBox.EditValueChanged += (s, a) => OnEvent(onColumnComboBoxTextChanged);
       }
 
+      public string SelectedUnit 
+      { 
+         get => _unitComboBox.EditValue as string;
```
msevestre, line 41: Why isn't it bound to an object like we always do?

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -181,8 +181,11 @@ public void UpdateDescriptionForModel(MappingDataFormatParameter mappingSource)
          }
          else
          {
-            column.Unit = _mappingParameterEditorPresenter.Unit;  
-            column.Dimension = _mappingParameterEditorPresenter.Dimension;
+            column.Unit = _mappingParameterEditorPresenter.Unit;
+            column.Dimension = _columnInfos
```
msevestre, line 185: THis set needs to be commented. too much nested info that makes it very hard to follow. 
Specifically, why is ` _mappingParameterEditorPresenter.Dimension;` not enough
also you are using, FIrst, are you sure to find one?.
Last when comparing , we use string.equals which deals with null properly (I may be wrong but I think == does not) . To verify :)

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -212,6 +215,11 @@ public void UpdateDescriptionForModel(MappingDataFormatParameter mappingSource)
          _view.CloseEditor();
       }
 
+      private bool dimensionContainsUnit(IDimension dimension, UnitDescription unit)
```
msevestre, line 218: we have a HasUnit.. why do you need to rewrite it >

msevestre (changes requested):

src/OSPSuite.UI/Views/Importer/UnitsEditorView.cs
```diff
@@ -36,6 +36,12 @@ public UnitsEditorView()
          _columnComboBox.EditValueChanged += (s, a) => OnEvent(onColumnComboBoxTextChanged);
       }
 
+      public string SelectedUnit
```
msevestre, line 39: I do not know why we are using something like this instead of usual binding to an object. You are breaking the pattern we are using everywhere. I am going to merge because this PR has been open for too long It would be good to try to address the comments in PR (either to explain why this is done like this or implement the requested/proposed changes) a bit quicker if possible

PR #1437, opened 2022-01-11
1409 importer handling of imvalid files

### LINKED ISSUE #1409: Importer: Handling of "invalid" files
Currently in the importer, if the first sheet of a file is identified as invalid (meaning it does not comply to one of the existing formats), it is not allowed to load the file. This can create a problem in case a user has a file where the first sheet is a description or chart fe. Also we are currently not allowing string values in the numerical columns, meaning we are rejecting files that can in the importer UI actually be configured to be imported.

Concerning this there are 3 suggestion to discuss:
a) allow string values in the numerical column (this actually is certainly a bug, since we are filtering them out later on)
b) allow files that have at least one valid sheet ( not necessarily the first one )
c) should we actually display even the non-valid sheets?

abdelr:
Fixes #1409


src/OSPSuite.Infrastructure.Import/Services/Importer.cs
```diff
@@ -113,10 +119,6 @@ public void CalculateFormat(IDataSourceFile dataSource, IReadOnlyList<ColumnInfo
             throw new UnsupportedFormatException(dataSource.Path);
 
          dataSource.AvailableFormats = AvailableFormats(dataSource.DataSheets[sheetName].RawData, columnInfos, metaDataCategories).ToList();
```
msevestre, line 121: Looking at this code, we are modifying the dataSource given as parameter. Can I suggest to change this methods so that it does not have side effect and returns the availableFormats instead? Line 108 for instance . It is not clear at all that the format will be set in this method. Thougts?

msevestre (commented):
One suggestion to avoid side effects

src/OSPSuite.Infrastructure.Import/Services/Importer.cs
msevestre: @abdelr not sure if you saw this

abdelr: > @abdelr not sure if you saw this

Sure, but this is still a work in progress since the last discussion on markers for the sheets

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1275,6 +1277,9 @@ public static class Error
       public static readonly string NaNOnData = "Data contains NaN values at imported columns. Select a different action for NaN values or clean your data.";
       public static readonly string UnsupportedFileType = "The type of file that you are trying to open is not currently supported";
       public static readonly string CannotRemoveBaseGridColumnStillInUse = "Cannot remove base grid column still used by other columns";
+      public static readonly string SimpleParseErrorMessage = "There were errors while parsing your data. Navigate to the sheets to read the concrete error.";
+
+      public static string ParseErrorMessage(IEnumerable<string> errors) => "There were errors while parsing your data: " + string.Join(". ", errors);
```
msevestre, line 1282: here you could use string interpolation and not the old school +

msevestre (commented):
This is a big PR with some interesting changes. I need some more time to review

msevestre (changes requested):
The change in ospsuite.Dimensions is suspicious

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -16,15 +19,15 @@ public interface IDataSource
    {
       void SetDataFormat(IDataFormat dataFormat);
       void SetNamingConvention(string namingConvention);
-      void AddSheets(Cache<string, DataSheet> dataSheets, IReadOnlyList<ColumnInfo> columnInfos, string filter);
+      Cache<IDataSet, List<ParseErrorDescription>> AddSheets(Cache<string, DataSheet> dataSheets, IReadOnlyList<ColumnInfo> columnInfos, string filter);
```
msevestre, line 22: This is an unexpected signature. What does this returned cache represent? .Can you comment the interface so that we know what's going on?
In general, returning a cache is not commun weird because you are exposing the API to the consumer but it might be required

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -16,15 +19,15 @@ public interface IDataSource
...
       void SetMappings(string fileName, IEnumerable<MetaDataMappingConverter> mappings);
       ImporterConfiguration GetImporterConfiguration();
       IEnumerable<MetaDataMappingConverter> GetMappings();
       Cache<string, IDataSet> DataSets { get; }
       IEnumerable<string> NamesFromConvention();
       NanSettings NanSettings { get; set; }
       ImportedDataSet DataSetAt(int index);
-      void ValidateDataSourceUnits(IReadOnlyList<ColumnInfo> columnInfos);
+      Cache<IDataSet, List<ParseErrorDescription>> ValidateDataSourceUnits(IReadOnlyList<ColumnInfo> columnInfos);
```
msevestre, line 30: same with this one.

src/OSPSuite.Infrastructure.Import/Core/Exceptions/ParseErrorDescription.cs
```diff
@@ -0,0 +1,66 @@
...
+
+   public class NaNParseErrorDescription : ParseErrorDescription
+   {
+      public NaNParseErrorDescription()
+      {
+         Message = Error.NaNOnData;
+      }
+   }
+
+   public class MismatchingArrayLengthsParseErrorDescription : ParseErrorDescription
```
msevestre, line 27: Why do we have all those classes instead of using the string? associated with the message?

src/OSPSuite.Infrastructure.Import/Core/Exceptions/ParseErrorDescription.cs
```diff
@@ -0,0 +1,66 @@
...
+      {
+         Message = Error.InvalidErrorDimension;
+      }
+   }
+
+   public class EmptyDataSetsParseErrorDescription : ParseErrorDescription
+   {
+      public EmptyDataSetsParseErrorDescription(IEnumerable<string> dataSetNames)
+      {
+         Message = Error.EmptyDataSet($"{string.Join(", ", dataSetNames)}");
```
msevestre, line 63: use dataSetNames.ToString(", ")

src/OSPSuite.Infrastructure.Import/Core/Helpers/CachedListHelpers.cs
```diff
@@ -0,0 +1,24 @@
...
+      public static void Add<T1, T2>(Cache<T1, List<T2>> cache, T1 key, T2 x)
+      {
+         if (cache.Contains(key))
+            cache[key].Add(x);
+         else
+            cache.Add(key, new List<T2>() { x });
+      }
+
+      public static void Add<T1, T2>(Cache<T1, List<T2>> cache, T1 key, IEnumerable<T2> list)
+      {
```
msevestre, line 17: Make in an extension method (Call this CachedListExtensions) and the first one should call the other instead of duplicating.

src/OSPSuite.Infrastructure.Import/Core/IDataSet.cs
```diff
@@ -34,10 +35,9 @@ public void AddData(IEnumerable<ParsedDataSet> range)
          _data.AddRange(range);
       }
 
-      public void ThrowsOnNan(double indicator)
+      public bool NanValuesExist(double indicator)
       {
-         if (_data.Any(dataSet => dataSet.Data.Any(pair => pair.Key.ColumnInfo.IsMandatory && pair.Value.Any(point => point.Measurement == indicator || double.IsNaN(point.Measurement)))))
-            throw new NanException();
+         return _data.Any(dataSet => dataSet.Data.Any(pair => pair.Key.ColumnInfo.IsMandatory && pair.Value.Any(point => point.Measurement == indicator || double.IsNaN(point.Measurement))));
```
msevestre, line 40: can we comment a bit on this line ...please. And maybe we can break the logic a bit here because it's very complicated to decipher
Also run resharper on this file

src/OSPSuite.Infrastructure.Import/Services/Importer.cs
```diff
@@ -102,22 +102,23 @@ public IDataSourceFile LoadFile(IReadOnlyList<ColumnInfo> columnInfos, string fi
 
          if (dataSource.DataSheets == null) return null;
 
-         CalculateFormat(dataSource, columnInfos, metaDataCategories, dataSource.DataSheets.Keys.FirstOrDefault());
 
-         return dataSource;
+         foreach(var x in dataSource.DataSheets.Keys)
+         {
+            dataSource.AvailableFormats = CalculateFormat(dataSource, columnInfos, metaDataCategories, x).ToList();
+            if (dataSource.AvailableFormats?.Count > 0)
```
msevestre, line 109: Any()

msevestre: How can it be nulll if you call ToList before?

src/OSPSuite.Infrastructure.Import/Services/Importer.cs
```diff
@@ -102,22 +102,23 @@ public IDataSourceFile LoadFile(IReadOnlyList<ColumnInfo> columnInfos, string fi
...
          if (sheetName == null)
             throw new UnsupportedFormatException(dataSource.Path);
 
-         dataSource.AvailableFormats = AvailableFormats(dataSource.DataSheets[sheetName].RawData, columnInfos, metaDataCategories).ToList();
-
-         if (dataSource.AvailableFormats.Count == 0)
-            throw new UnsupportedFormatException(dataSource.Path);
-
-         dataSource.Format = dataSource.AvailableFormats.FirstOrDefault();
+         return AvailableFormats(dataSource.DataSheets[sheetName].RawData, columnInfos, metaDataCategories);
```
msevestre, line 121: This could return a IReadOnly so that you don't have to TOList earlier.

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -16,15 +19,15 @@ public interface IDataSource
    {
       void SetDataFormat(IDataFormat dataFormat);
       void SetNamingConvention(string namingConvention);
-      void AddSheets(Cache<string, DataSheet> dataSheets, IReadOnlyList<ColumnInfo> columnInfos, string filter);
+      Cache<IDataSet, List<ParseErrorDescription>> AddSheets(Cache<string, DataSheet> dataSheets, IReadOnlyList<ColumnInfo> columnInfos, string filter);
       void SetMappings(string fileName, IEnumerable<MetaDataMappingConverter> mappings);
       ImporterConfiguration GetImporterConfiguration();
       IEnumerable<MetaDataMappingConverter> GetMappings();
       Cache<string, IDataSet> DataSets { get; }
```
msevestre, line 26: I don't like this as a cache exposed. DataSet are added outside of the DataSource as it stands

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -72,16 +75,18 @@ public void SetNamingConvention(string namingConvention)
...
 
-      public void AddSheets(Cache<string, DataSheet> dataSheets, IReadOnlyList<ColumnInfo> columnInfos, string filter)
-      {         
+      public Cache<IDataSet, List<ParseErrorDescription>> AddSheets(Cache<string, DataSheet> dataSheets, IReadOnlyList<ColumnInfo> columnInfos, string filter)
+      {
          _importer.AddFromFile(_configuration.Format, filterSheets(dataSheets, filter), columnInfos, this);
          if (NanSettings == null || !double.TryParse(NanSettings.Indicator, out var indicator))
             indicator = double.NaN;
+         var errors = new Cache<IDataSet, List<ParseErrorDescription>>();
          foreach (var dataSet in DataSets.KeyValues)
```
msevestre, line 84: why are you iterating over the keyValues? You are never using the Key

msevestre (changes requested):
I see wha't's happening and this is ok. We are breaking a lot of encapsulation but passing Cache back and forth .
Cache is great because you can define a way to retrieve keys automatically by specifing a lambda. We are never using it. WE also never know what the key is. Is it the sheetName sometimes? what's a DataSet compared to a Sheet. Why are we adding DataSet with the sheet key as key..
Lots of questions for me...also coming from older code. We need to run resharper on those files because some basic optimization and code clean up are not done

abdelr:
@msevestre I encapsulated the errors on a new class. I hope it makes more sense now?

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -227,22 +277,30 @@ private void validateUnitsSupportedAndSameDimension(IReadOnlyList<ColumnInfo> co
...
                      }
                   }
                }
             }
          }
+         return errors;
       }
 
-      void IDataSource.ValidateDataSourceUnits(IReadOnlyList<ColumnInfo> columnInfos)
+      ParseErrors IDataSource.ValidateDataSourceUnits(IReadOnlyList<ColumnInfo> columnInfos)
```
msevestre, line 299: Why the IDataSource. here?

src/OSPSuite.Infrastructure.Import/Core/Helpers/CachedListHelpers.cs
```diff
@@ -0,0 +1,24 @@
+﻿using OSPSuite.Utility.Collections;
+using System.Collections.Generic;
+
+namespace OSPSuite.Infrastructure.Import.Core.Helpers
+{
+   static public class CachedListHelpers
```
msevestre, line 6: This needs to be removed right?

src/OSPSuite.Infrastructure.Import/Core/IDataSet.cs
```diff
@@ -34,10 +35,21 @@ public void AddData(IEnumerable<ParsedDataSet> range)
          _data.AddRange(range);
       }
 
-      public void ThrowsOnNan(double indicator)
+      public bool NanValuesExist(double indicator)
       {
-         if (_data.Any(dataSet => dataSet.Data.Any(pair => pair.Key.ColumnInfo.IsMandatory && pair.Value.Any(point => point.Measurement == indicator || double.IsNaN(point.Measurement)))))
-            throw new NanException();
+         //returns true if for any ParsedDataSet 
+         return _data.Any(
```
msevestre, line 41: YES!!!!

src/OSPSuite.UI/Views/Importer/ImporterDataView.cs
```diff
@@ -222,13 +228,12 @@ public string GetActiveFilterCriteria()
 
       public void AddTabs(List<string> sheetNames)
       {
-         //we should seek an alternative
+         importerTabControl.Images = imageCollection1;
```
msevestre, line 231: rename this control

src/OSPSuite.UI/Views/Importer/ImporterDataView.resx
```diff
@@ -117,8 +117,50 @@
...
     <value>17, 17</value>
+  </metadata>
+  <data name="useForImportCheckEdit.ToolTip" xml:space="preserve">
+    <value>When selected, the filter will apply to the data during the import process. When deselected, the filter only affects this view. Check documentation for more information on defining filters: &lt;href=https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering&gt;https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering&lt;/href&gt;</value>
+  </data>
+  <metadata name="imageCollection1.TrayLocation" type="System.Drawing.Point, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
+    <value>163, 17</value>
+  </metadata>
+  <assembly alias="DevExpress.Utils.v21.2" name="DevExpress.Utils.v21.2, Version=21.2.3.0, Culture=neutral, PublicKeyToken=b88d1754d700e49a" />
+  <data name="imageCollection1.ImageStream" type="DevExpress.Utils.ImageCollectionStreamer, DevExpress.Utils.v21.2" mimetype="application/x-microsoft.net.object.bytearray.base64">
```
msevestre, line 130: We do not use images like that at all. Icons are added to the ImageList and are loaded all dynamically

msevestre (changes requested):
The domain codes is SO Much better. Thanks
The UI code needs to be updated :
Some controls have default naming with 1 at the end
Not proper usage of icons

src/OSPSuite.UI/Views/Importer/ImporterDataView.cs
```diff
@@ -286,10 +291,55 @@ public void SetFilter(string filter)
...
+      }
       public string GetFilter()
       {
          return DevExpress.Data.Filtering.CriteriaToWhereClauseHelper.GetDataSetWhere(dataViewingGridView.ActiveFilterCriteria);
       }
+
+      private void refreshErrorMessage()
+      {
+
+         if (_lastLoadedDataSets.Contains(SelectedTab) && _lastErrors.Contains(_lastLoadedDataSets[SelectedTab]))
```
msevestre, line 311: somehow this feels like this code belongs in the presenter.

src/OSPSuite.UI/Views/Importer/ImporterDataView.cs
```diff
@@ -286,10 +291,55 @@ public void SetFilter(string filter)
...
+            labelControlError.Text = "";
+            layoutControlItem2.Visibility = DevExpress.XtraLayout.Utils.LayoutVisibility.Never;
+         }
+      }
+
+      private void refreshErrorMarks()
+      {
+         importerTabControl.TabPages.Each(x =>
+         {
+            if (_lastLoadedDataSets.Contains(x.Text))
```
msevestre, line 327: this codes for which image to show also belongs in the presenter 
And hopefully, we won't have those magin number anymore 1 and 0 once we use ApplicationIcons

src/OSPSuite.UI/Views/Importer/ImporterDataView.resx
```diff
@@ -117,8 +117,50 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
-  <assembly alias="System.Drawing" name="System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
-  <data name="errorProvider.TrayLocation" type="System.Drawing.Point, System.Drawing">
+  <metadata name="errorProvider.TrayLocation" type="System.Drawing.Point, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
     <value>17, 17</value>
+  </metadata>
+  <data name="useForImportCheckEdit.ToolTip" xml:space="preserve">
+    <value>When selected, the filter will apply to the data during the import process. When deselected, the filter only affects this view. Check documentation for more information on defining filters: &lt;href=https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering&gt;https://docs.open-systems-pharmacology.org/shared-tools-and-example-workflows/features-of-tables#filtering&lt;/href&gt;</value>
```
msevestre, line 124: Tootip needs to be put into constants!

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -155,7 +155,26 @@ public IEnumerable<string> NamesFromConvention()
...
 
-      public ImportedDataSet DataSetAt(int index)
+      public IDataSet DataSetAt(int index)
+      {
+         var sheetIndex = 0;
+         var sheet = DataSets.GetEnumerator();
+         var accumulatedIndexes = 0;
+         while (sheet.MoveNext() && index >= 0)
+         {
+            if (sheet.Current.Data.Count() > index)
```
msevestre, line 165: (sheet.Current.Data.Count() 3 times in the same while loop. Maybe cache it?

abdelr:
@msevestre Only open PR. Just adding you to the comment to make sure you know this is ready for revision.

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -117,7 +155,27 @@ public IEnumerable<string> NamesFromConvention()
          return _importer.NamesFromConvention(_configuration.NamingConventions, _configuration.FileName, DataSets, _mappings);
       }
 
-      public ImportedDataSet DataSetAt(int index)
+      public IDataSet DataSetAt(int index)
```
msevestre, line 158: Let's talk about this in the meeting. I don't understand what this is doing.

msevestre:
I think this is working as expected. I have my concerns regarding the deeply nested structure in DataSource.cs
I am going to merge this however and this can be addressed at some other time. I will create an issue for this

PR #1440, opened 2022-01-11
Fixes #1433 maximal number of outputs

### LINKED ISSUE #1433: Maximal number of output points
The number of output points (in the OutputSchema) is restricted to 5000. This value is a great limitation when simulating months (or even years). I think this value should be increased (or even exposed as an input parameter).

msevestre:


src/OSPSuite.Core/Serialization/SimModel/DTO/OutputSchemaExport.cs
```diff
@@ -29,9 +29,7 @@ public int NumberOfTimePoints
             if (StartTime == EndTime)
                return Constants.MIN_NUMBER_OF_POINTS_PER_INTERVAL; //just a single point
 
-            return Math.Min(
-               Math.Max(Constants.MIN_NUMBER_OF_POINTS_PER_INTERVAL, (int) Math.Round(Resolution * (EndTime - StartTime)) + 1),
-               Constants.MAX_NUMBER_OF_POINTS_PER_INTERVAL);
+            return Math.Max(Constants.MIN_NUMBER_OF_POINTS_PER_INTERVAL, (int) Math.Round(Resolution * (EndTime - StartTime)) + 1);
```
msevestre, line 32: @Yuri05 Any idea why we have + 1 ?
If I have 2 hours (start time 0, end time 2) , and resolution 1000 pts/h , I am expecting 2000 and not 2001 point. WHY??

Yuri05: Actually I expect 2001 points indeed, the last one at t=2h :)
Easier to see with few points: e.g. if I have 2 points/h and simulate 2 hours, I expect 5 points, at t=0, 30min, 1h, 1h30 and 2h.

Yuri05 (approved):

msevestre:
All good then;-)

PR #1443, opened 2022-01-13
1404 empty column error fixed

### LINKED ISSUE #1404: New data import misaligns values/column header
Attached is a xlsx-file as generated from a plot scanning tool. 
After importing to MoBi, the values and column headers are misaligned - see attched screenshot (it seems due to the empty column present in the Excel table).

[test.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/7736688/test.xlsx)

Preview_Source_Column

georgeDaskalakis:
fixes #1404


msevestre:
The build is failing?!?

PR #1453, opened 2022-01-21
Fixes #1452

### LINKED ISSUE #1452: TryGetValues does crash! and should not

msevestre:


msevestre (commented):
@Yuri05 One important test missing

PR #1462, opened 2022-01-28
Do not prompt for close on import

### LINKED ISSUE #1460: Importing a file: Are you sure you want to cancel
Clicking on Import does the following

image

WHAT?

abdelr:
Fixes #1460 (Do not prompt for close on import)

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
@@ -33,6 +33,7 @@ namespace OSPSuite.Presentation.Presenters.Importer
    {
       private readonly IImporterPresenter _importerPresenter;
       private readonly IDialogCreator _dialogCreator;
+      private bool _shouldPromptBeforeClose = false;
 
       public ModalImporterPresenter(IModalImporterView view, IImporterPresenter importerPresenter, IDialogCreator dialogCreator) : base(view)
       {
@@ -41,12 +42,11 @@ namespace OSPSuite.Presentation.Presenters.Importer
          AddSubPresenters(importerPresenter);
       }
 
-      private bool _shouldPromtBeforeClose = false;
       public override bool ShouldClose
       {
          get
          {
-            if (!_shouldPromtBeforeClose)
+            if (!_shouldPromptBeforeClose)
                return true;
             var shouldCancel = _dialogCreator.MessageBoxYesNo(Captions.ReallyCancel);
             return shouldCancel == ViewResult.Yes;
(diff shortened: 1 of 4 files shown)


msevestre:
What does the view.CloseOnImport do? Why is the presenter asking the view to trigger the close? This feels weird.

Also needs test

msevestre (changes requested):
Test missing and wondering about flow between view and presenter. Going to check the code

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
```diff
@@ -41,10 +41,13 @@ public ModalImporterPresenter(IModalImporterView view, IImporterPresenter import
          AddSubPresenters(importerPresenter);
       }
 
+      private bool _shouldPromtBeforeClose = false;
```
msevestre, line 44: move the definition of this flag at the top (use resharper) and correct typo (Prompt)

src/OSPSuite.Presentation/Presenters/Importer/IModalImporterPresenter.cs
```diff
@@ -106,7 +109,12 @@ ImporterConfiguration configuration
             results.Each(r => r.ConfigurationId = configurationId);
          }
 
-         _importerPresenter.OnTriggerImport += (s, d) => { _view.CloseOnImport(); };
+         _importerPresenter.OnTriggerImport += (s, d) => 
+         {
+            _shouldPromtBeforeClose = false;
+            _view.CloseOnImport();
```
msevestre, line 115: Oh see. This methods already exists and it is called CloseView. 
From the name it looked it was triggering some event.. (Onxxx) .Naming is really everything in programming :)

IndrajeetPatil: > "There are only two hard things in Computer Science: cache invalidation and naming things"?
> Phil Karlton

😅

abdelr:
@msevestre the test is not valid because the view.Display() method returns immediately. So, this is a blocking scenario. I have some ideas on how to implement this but I wonder if we already have some support for it. We could talk later about it if you want.

msevestre:
A few remarks:
First,why are you registering the event outside of the constructor?
Second, why are you registering to the event twice?
Last we never have this type of boolean hanging around to decide if we should close or not. Any idea why this is required here?

msevestre (commented):
Is the test ok now? @abdelr

abdelr:
We still do not have a cancel button, but I am not sure if we want to change this. In any case, if we do, it will be just adding the button which will invoke form close but keep in mind that we already have some buttons changing captions and could be a bit confusing. We have the X button anyways.

PR #1465, opened 2022-02-01
try parse unit from description

### LINKED ISSUE #1427: Importer should try to parse dimension/unit from header after selecting a column for "Measurement"
[TestData.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/7828854/TestData.xlsx)

In MoBi:

- Import the sheet "Mitrakou_1990" for import. In the default mapping, the column "MW" is selected for measurement (which complies with the specifications)

image

- Map "Measurement" to the column "Concentration (molar) [µmol/l]"

The "Units" are not updated. I would expect the importer to try to guess the units and set them accordingly.

abdelr:
Fixes #1427 (Try to parse the unit from the column description)

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
@@ -110,6 +110,12 @@ namespace OSPSuite.Infrastructure.Import.Core.DataFormat
 
       protected abstract UnitDescription ExtractUnits(string description, IUnformattedData data, List<string> keys, IReadOnlyList<IDimension> supportedDimensions,  ref double rank);
 
+      public UnitDescription ExtractUnits(string description, IReadOnlyList<IDimension> supportedDimensions)
+      {
+         var rank = 0.0;
+         return ExtractUnits(description, null, null, supportedDimensions, ref rank);
+      }
+
       protected virtual void ExtractQualifiedHeadings(List<string> keys, List<string> missingKeys, IReadOnlyList<ColumnInfo> columnInfos, IUnformattedData data, ref double rank)
       {
          foreach (var header in columnInfos)
--- a/src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
@@ -110,10 +110,10 @@ namespace OSPSuite.Infrastructure.Import.Core.DataFormat
 
       protected abstract UnitDescription ExtractUnits(string description, IUnformattedData data, List<string> keys, IReadOnlyList<IDimension> supportedDimensions,  ref double rank);
 
-      public UnitDescription ExtractUnits(string description, IReadOnlyList<IDimension> supportedDimensions)
+      public UnitDescription ExtractUnitDescriptions(string description, IReadOnlyList<IDimension> supportedDimensions)
       {
          var rank = 0.0;
-         return ExtractUnits(description, null, null, supportedDimensions, ref rank);
+         return ExtractUnits(description, data: null, keys: null, supportedDimensions, ref rank);
       }
 
       protected virtual void ExtractQualifiedHeadings(List<string> keys, List<string> missingKeys, IReadOnlyList<ColumnInfo> columnInfos, IUnformattedData data, ref double rank)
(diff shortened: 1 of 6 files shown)


src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -110,6 +110,12 @@ private void setSecondaryColumnUnit(IReadOnlyList<ColumnInfo> columnInfos)
 
       protected abstract UnitDescription ExtractUnits(string description, IUnformattedData data, List<string> keys, IReadOnlyList<IDimension> supportedDimensions,  ref double rank);
 
+      public UnitDescription ExtractUnits(string description, IReadOnlyList<IDimension> supportedDimensions)
+      {
+         var rank = 0.0;
```
msevestre, line 115: what is this rank again? and should not the rank be an int?

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -110,6 +110,12 @@ private void setSecondaryColumnUnit(IReadOnlyList<ColumnInfo> columnInfos)
...
+         return ExtractUnits(description, null, null, supportedDimensions, ref rank);
```
msevestre, line 116: name parameter so we know what is null here ?

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -526,6 +526,9 @@ private void setDescriptionForRow(ColumnMappingDTO model, bool isColumn)
                   break;
                case ColumnMappingDTO.ColumnType.Mapping:
                   model.Source = new MappingDataFormatParameter(model.ExcelColumn, new Column() {Name = model.MappingName, Unit = new UnitDescription(UnitDescription.InvalidUnit)});
+                  var unit = _format.ExtractUnits(model.ExcelColumn, _columnInfos.First(x => x.DisplayName == model.MappingName).SupportedDimensions);
```
msevestre, line 529: Very hard for me to review here. Are you sure here that you have one column with the displayName == MAppingName?

src/OSPSuite.Infrastructure.Import/Core/IDataFormat.cs
```diff
@@ -18,5 +19,6 @@ public interface IDataFormat
       void CopyParametersFromConfiguration(OSPSuite.Core.Import.ImporterConfiguration configuration);
       IList<string> ExcelColumnNames { get; }
       IEnumerable<ParsedDataSet> Parse(IUnformattedData data, IReadOnlyList<ColumnInfo> columnInfos);
+      UnitDescription ExtractUnits(string description, IReadOnlyList<IDimension> supportedDimensions);
```
msevestre, line 22: Should it be ExtractUnit since you return a UnitDescription?

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
abdelr: rank is a number that will indicate how good a match is a format for a given file. So, if there are two formats (right now we have only one but we discussed about having more at some point) that are valid for a file, the rank parameter will tell how much of a format specific feature was detected so the format with the higher rank should be selected. If you have a file with units on the headers, it should also be possible to use the format with the units on the columns but the one with units on the headers should be the first option since the file has this specific feature. I hope it makes more sense now but we could off course discuss it.

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
abdelr: I am creating a temporal variable for a cleaner code. It should not be the case that the First throws because we are getting the supported dimensions from a valid mapping column. So, if the MappingName does not exist on the columnInfos collection, it means this is not a valid mapping. I could create a specific error for it although this is not an error for the user but for the developer. An error here means the code is wrong and not anything done by the user.

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
msevestre: I see. Not sure why rank is double and not an integer but this works

msevestre (approved):

PR #1466, opened 2022-02-01
Prompting user when resetting mapping based on current sheet

### LINKED ISSUE #1463: Importer: When selecting "Reset configuration based on current sheet", show hint
"This action will result in dropping all loaded sheets. Do you want to proceed?", as it actually clears the imported mapping.

abdelr:
Fixes #1463 (Prompting user when resetting mapping based on current sheet)


msevestre:
Why is this closed?

abdelr:
Was implemented already. Explanation on the issue #1463

PR #1467, opened 2022-02-01
Reset mapping before loading configuration

### LINKED ISSUE #1459: Importer: Confusing behavior when loading the configuration valid for the selected but not for the first sheet
An excel file with two sheets - `Sheet A` and `Sheet B` with different structures. A configuration stored as xlm that is valid for `Sheet B`.

- Add observed data -> Select the file.
- Select `Sheet B`
- Load configuration ->Error 
 

```
PK-Sim®

The following excel columns do not exist in the current file. 
 
'Dose [unit]
Concentration (mass) [ng/ml]' 

The corresponding mappings from the configuration will be lost. Do you want to continue?

```

I actually first have to select "Reset configuration based on selected sheet" and then load the configuration... This is not the expected behavior. I think the "Reset configuration based on selected sheet" should be performed in background when loading a configuration.

abdelr:
Fixes #1459 (Reset mapping before loading configuration)

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
@@ -89,7 +89,7 @@ namespace OSPSuite.Presentation.Presenters.Importer
          _importerDataPresenter.OnTabChanged += onTabChanged;
          _importerDataPresenter.OnDataChanged += onImporterDataChanged;
          _columnMappingPresenter.OnMissingMapping += onMissingMapping;
-         _columnMappingPresenter.OnResetMappingBasedOnCurrentSheet += onResetMappingBasedOnCurrentSheet;
+         _columnMappingPresenter.OnResetMappingBasedOnCurrentSheet += (o, e) => onResetMappingBasedOnCurrentSheet();
          _columnMappingPresenter.OnMappingCompleted += onCompletedMapping;
          View.DisableConfirmationView();
       }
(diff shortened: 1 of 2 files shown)


src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
```diff
@@ -404,6 +404,8 @@ public void LoadConfigurationWithoutImporting()
          if (confirmDroppingOfLoadedSheets())
                return;
 
+         onResetMappingBasedOnCurrentSheet(this, null);
```
msevestre, line 407: onResetMappingBasedOnCurrentSheet does not use either e nor null. Can I suggest to remove the argument in this mehtod, call in resetMappingBasedOnCurrentSheet and in the event registration, do (o, e)=> eesetMappingBasedOnCurrentSheet()

tests/OSPSuite.Presentation.Tests/Importer/Presenters/ImporterPresenterSpecs.cs
```diff
@@ -429,4 +442,19 @@ public void must_filter_empty_mappings()
...
+      protected override void Because()
+      {
+         (sut as ImporterPresenterForTest).OnResetMappingBasedOnCurrentSheetInvoked = false;
+         sut.LoadConfigurationWithoutImporting();
+      }
+
+      [Observation]
+      public void must_reset_format_based_on_current_sheet()
+      {
+         (sut as ImporterPresenterForTest).OnResetMappingBasedOnCurrentSheetInvoked.ShouldBeTrue();
```
msevestre, line 457: neat.

msevestre (changes requested):
suggestion so that we do not call a method that is an event handler with weird aguments

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
abdelr: You are right

msevestre (approved):

PR #1468, opened 2022-02-01
Fixes #1464 throw when running the same twice

### LINKED ISSUE #1464: Throw exception when running the same simulation with runSimulations
as we are caching with id

msevestre:


src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
```diff
@@ -38,56 +38,99 @@ public interface IConcurrencyManager
...
+      /// </summary>
+      /// <typeparam name="TData">Data type to consume by the worker function</typeparam>
+      /// <param name="data">List of data to consume by the workers</param>
+      /// <param name="action">
+      ///    An action running on each worker on each piece of data.
+      /// </param>
+      /// <param name="cancellationToken">Cancellation token to cancel the threads</param>
+      /// <param name="numberOfCoresToUse">Number of cores to use. Use null to take all cores</param>
+      /// <returns>Dictionary binding a result for each input data after running the action on it</returns>
+      Task RunAsync<TData>(
```
msevestre, line 64: @abdelr As discussed. Thanks for the suggestion

src/OSPSuite.R/Domain/ConcurrentRunSimulationBatch.cs
```diff
@@ -41,12 +41,9 @@ internal SimulationBatch AddNewBatch()
 
       public IReadOnlyCollection<SimulationBatch> SimulationBatches => _simulationBatches;
 
-      public string AddSimulationBatchRunValues(SimulationBatchRunValues simulationBatchRunValues)
+      public void AddSimulationBatchRunValues(SimulationBatchRunValues simulationBatchRunValues)
       {
-         var id = generateId();
-         simulationBatchRunValues.Id = id;
```
msevestre, line 47: I am moving the creation of the Id to the object itselft...

src/OSPSuite.R/Services/ConcurrentSimulationRunner.cs
```diff
@@ -98,22 +98,23 @@ public void Clear()
...
-               settings => Enumerable.Range(0, settings.MissingBatchesCount).Select(_ => settings)
-            ).ToList(),
-            (settings, ct) => settings.AddNewBatch(),
+            batch => Enumerable.Range(0, batch.MissingBatchesCount).Select(_ => batch)
+         ).ToList();
+
+         return _concurrencyManager.RunAsync(
+            batches,
+            //Use {} to specify using the overload without return value
+            (batch, ct) => { batch.AddNewBatch(); },
```
msevestre, line 115: This is the only downside of using the same name... I need to use the {} to that it picks the right overload

tests/OSPSuite.Presentation.Tests/Services/DataImporterSpecs.cs
```diff
@@ -186,19 +189,30 @@ public void should_import_simple_data_from_csv()
       }
 
       [Observation]
-      public void should_return_empty_on_invalid_file_name()
+      [TestCase("IntegrationSample1.csv", 2)]
+      [TestCase("sample_non_existent.csv", 2)]
```
msevestre, line 193: oups... this does not belong here!

tests/OSPSuite.R.Tests/Services/ConcurrentSimulationRunnerSpecs.cs
```diff
@@ -235,11 +238,18 @@ public override void GlobalContext()
...
+         {
+            ParameterValues = new[] {3.5, 0.53}
+         };
+         _parValues3 = new SimulationBatchRunValues
+         {
+            ParameterValues = new[] {3.5, 0.53}
+         };
 
          _simulationBatch1.AddSimulationBatchRunValues(_parValues1);
-         _simulationBatch2.AddSimulationBatchRunValues(_parValues1);
```
msevestre, line 241: Adding the same object multiple time is not allowed anymore... unique ids

msevestre (commented):
@abdelr Ready to review

src/OSPSuite.Core/Domain/Services/ConcurrencyManager.cs
```diff
@@ -38,56 +38,99 @@ public interface IConcurrencyManager
...
                {
-                  cancellationToken.ThrowIfCancellationRequested();
-                  try
-                  {
-                     results.TryAdd(
-                        datum, 
-                        new ConcurrencyManagerResult<TResult>(datum.Id, action(datum, cancellationToken))
-                     );
-                  }
-                  catch (Exception e)
```
abdelr, line 79: You removed the try/catch. This was here to make sure that if one run fails (running one simulation throws an exception for example) the rest are still recorded. Like this I think you would not be able to return anything at all. Right?

msevestre: yep. But here we have no way of knowing which one failed as we have no identifier.

abdelr: Fair enough.

abdelr (approved):

PR #1475, opened 2022-02-01
1474 column without name fixed

### LINKED ISSUE #1474: Import parameter values from XLS: error a column without name is present
Error description:

https://github.com/Open-Systems-Pharmacology/MoBi/issues/679

georgeDaskalakis:
closes #1474


src/OSPSuite.Infrastructure.Import/Services/IImportFromExcelTask.cs
```diff
@@ -100,13 +100,23 @@ private static IEnumerable<DataTable> dataTables(string fileName, string sheetNa
 
             while (reader.MoveToNextRow())
             {
-               dataTable.AddRowToDataTable(reader.CurrentRow);
+               var rowList = reader.CurrentRow;
```
msevestre, line 103: i am surprised about this naming rowList...when you have one row?

msevestre: isn't it more rowValues?

src/OSPSuite.Infrastructure.Import/Services/IImportFromExcelTask.cs
```diff
@@ -100,13 +100,23 @@ private static IEnumerable<DataTable> dataTables(string fileName, string sheetNa
...
 
             dataTablesList.Add(dataTable);
          }
 
          return dataTablesList;
       }
+
+      private static List<string> removeColumnsWithoutHeader(List<string> rowList, int headerCount)
+      {
+         if (rowList.Count > headerCount)
```
msevestre, line 117: What if the empty header is between columns. This would not work correct?

src/OSPSuite.Infrastructure.Import/Services/IImportFromExcelTask.cs
```diff
@@ -100,13 +100,23 @@ private static IEnumerable<DataTable> dataTables(string fileName, string sheetNa
...
+            rowList = rowList.GetRange(0, headerCount);
```
msevestre, line 118: If I am understanding properly, you are stopping at the first header empty. Is this the expected behavior

msevestre (commented):
Some questions

src/OSPSuite.Infrastructure.Import/Services/IImportFromExcelTask.cs
georgeDaskalakis: true. corrected

src/OSPSuite.Infrastructure.Import/Services/IImportFromExcelTask.cs
georgeDaskalakis: Not the the first empty header, we stop at the first element of a row that exceeds the number of headers

src/OSPSuite.Infrastructure.Import/Services/IImportFromExcelTask.cs
georgeDaskalakis: So this would not work, but not for this reason. The dataTable is being read correctly, because we first read the headers, and if there is an empty header inbetween existing ones we read "" (fe rowValues = ["e1", "", "e3"]). This results in the creation of an empty column, which seems to be not a problem since for the StartValues where this is used we do not even show the columns to the user.

The problem is that in MoBi the index of the columns (Unit, value etc) are hard coded. So even if the columns come with the wrong order, the start values cannot be imported from the file. Thanks for this comment, because otherwise I would not have tested with an empty column inbetween and would not have found out ;) 

I will do a separate issue for this on MoBi, since there exists the problematic code. The code here should remain the same. 

As a note: ideally the whole code should be rafactored to use our own data structures (as in the Observed Data Importer) instead of the DataTable that it currently does. I am not sure now is the right time for this, but we could also do a low-prio issue for this if you agree.

msevestre: >The problem is that in MoBi the index of the columns (Unit, value etc) are hard coded. So even if the columns come with the wrong order, the start values cannot be imported from the file. Thanks for this comment, because otherwise I would not have tested with an empty column inbetween and would not have found out ;)

Yeah we did not want to be that flexible at the time. Not a big deal for now. YEs creete a separate issue for this but what we are using is NOT DataRepository , no base gird etc so I am not sure how this would fit in the whole importer flow.

PR #1483, opened 2022-02-07
Fixes #1457 gc for R

### LINKED ISSUE #1457: GC for R
Export GC function in OSPSuite.R

msevestre:


src/OSPSuite.Core/Services/GarbageCollectionTask.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace OSPSuite.Core.Services
+{
+   public static class GarbageCollectionTask
+   {
+      /// <summary>
+      ///    Force GC Collection compacting the Large Object Heap. Implementation very similar to what DotMemory
+      ///    is doing according to one of their post
+      /// </summary>
+      public static void ForceGC()
```
msevestre, line 12: I think this is a better way of clearing the memory. Let see...

src/OSPSuite.Core/Workspace.cs
```diff
@@ -48,7 +49,7 @@ public void UpdateJournalPathRelativeTo(string projectFileFullPath)
       protected void ReleaseMemory()
       {
          //This might free some C++ objects still  being held into memory
-         this.GCCollectAndCompact();
```
msevestre, line 51: Having it as extension method was stupid. This is better as a simple static method

src/OSPSuite.R/Api.cs
```diff
@@ -65,6 +65,14 @@ public static void InitializeOnce(ApiConfig apiConfig)
 
       public static IDataImporterTask GetDataImporterTask() => resolveTask<IDataImporterTask>();
 
+      /// <summary>
+      /// Forces the Garbage collection 
+      /// </summary>
+      public static void ForceGC()
```
msevestre, line 71: Availalbe for R

msevestre (commented):
ready

src/OSPSuite.Core/Services/GarbageCollectionTask.cs
```diff
@@ -0,0 +1,22 @@
...
+      {
+         GCSettings.LargeObjectHeapCompactionMode = GCLargeObjectHeapCompactionMode.CompactOnce;
+         for (int i = 0; i < 4; i++)
+         {
+            GC.Collect(2, GCCollectionMode.Forced, blocking: true);
+            GC.WaitForPendingFinalizers();
+         }
```
Yuri05, line 19: where does the 4 in the GC-loop actually come from?

msevestre: no idea. Empirical value apparently that mimics so behavior. There is always stuff left if you do it only 3 times or less

Yuri05: hmm, ok :)

Yuri05 (approved):

src/OSPSuite.Core/Services/GarbageCollectionTask.cs
msevestre: I have added  a link to the discussion in code
https://dotnettools-support.jetbrains.com/hc/en-us/community/posts/360000088690-How-reproduce-DotMemory-s-Force-GC-button-s-behaviour-on-code-with-c-

PR #1485, opened 2022-02-07
Fixes #1484 value origin

### LINKED ISSUE #1484: ValueOrigin: Do not save when exiting view
This will be a breaking change but this makes sense. As discussed, OK button will be visible and chnages will only be accpeted if ok is pressed

msevestre:


src/OSPSuite.UI/Binders/ValueOriginBinder.cs
```diff
@@ -73,7 +74,7 @@ private void onToolTipControllerGetActiveObjectInfo(object sender, ToolTipContro
 
       private void closeUp(CloseUpEventArgs e)
       {
-         var cancel = e.CloseMode == PopupCloseMode.Cancel;
+         var cancel = e.CloseMode != PopupCloseMode.Normal;
```
msevestre, line 77: WE now have different mode of closing. Normal means OK is clicked (going to add this to the code as comment in fact)

src/OSPSuite.UI/Controls/UxRepositoryItemPopupContainerEditOKButton.cs
```diff
@@ -0,0 +1,123 @@
...
+using DevExpress.XtraEditors.Registrator;
+using DevExpress.XtraEditors.Repository;
+using DevExpress.XtraEditors.ViewInfo;
+
+namespace OSPSuite.UI.Controls
+{
+   //adapted from
+   //https://supportcenter.devexpress.com/ticket/details/e3666/how-to-display-the-default-ok-button-in-the-popupcontaineredit-s-popup-window#PopupContainerEditOKButton.cs
+
+   public class UxRepositoryItemPopupContainerEditOKButton : RepositoryItemPopupContainerEdit
```
msevestre, line 14: @abdelr @georgeDaskalakis @Yuri05 Can't believe there is so much code to write. But it looks with this code like this
image

Yuri05 (approved):

src/OSPSuite.UI/Controls/UxRepositoryItemPopupContainerEditOKButton.cs
Yuri05: Very nice! Will make some users very happy :)

msevestre: Yep. We should use this guy for all instances of REposityPopupContainer

PR #1486, opened 2022-02-08
Update ParameterIdentificationFeedbackViewVisibilityUICommand.cs

### LINKED ISSUE #1372: Should be possible to switch ParameterIdentification from Feedback view

abdelr:
Fixes #1372 (Using ActiveSubjectRetriever to retrieve the active parameter identification instead of listening to events)

### DIFF
--- a/src/OSPSuite.Presentation/UICommands/ParameterIdentificationFeedbackViewVisibilityUICommand.cs
+++ b/src/OSPSuite.Presentation/UICommands/ParameterIdentificationFeedbackViewVisibilityUICommand.cs
@@ -5,7 +5,7 @@ using OSPSuite.Presentation.Services;
 
 namespace OSPSuite.Presentation.UICommands
 {
-   public class ParameterIdentificationFeedbackViewVisibilityUICommand : ActiveObjectUICommand<IParameterIdentificationFeedbackPresenter>
+   public class ParameterIdentificationFeedbackViewVisibilityUICommand : ActiveObjectUICommand<ParameterIdentification>
    {
       private readonly ISingleStartPresenterTask _singleStartPresenterTask;
       private readonly IParameterIdentificationFeedbackManager _parameterIdentificationFeedbackManager;
@@ -18,7 +18,7 @@ namespace OSPSuite.Presentation.UICommands
 
       protected override void PerformExecute()
       {
-         _singleStartPresenterTask.StartForSubject(_parameterIdentificationFeedbackManager.GetFeedbackFor(Subject.ParameterIdentification));
+         _singleStartPresenterTask.StartForSubject(_parameterIdentificationFeedbackManager.GetFeedbackFor(Subject));
       }
    }
 }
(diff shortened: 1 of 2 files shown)


msevestre:
yes much better. This would have created. a leak the way it was before because the PI was not cleared anyways

src/OSPSuite.Presentation/UICommands/ParameterIdentificationFeedbackViewVisibilityUICommand.cs
```diff
@@ -1,31 +1,26 @@
...
-using OSPSuite.Core.Events;
+using OSPSuite.Core.Services;
 using OSPSuite.Presentation.Presenters.ParameterIdentifications;
 using OSPSuite.Presentation.Services;
-using OSPSuite.Utility.Events;
 
 namespace OSPSuite.Presentation.UICommands
 {
-   public class ParameterIdentificationFeedbackViewVisibilityUICommand : ObjectUICommand<IParameterIdentificationFeedbackPresenter>, IListener<ParameterIdentificationSelectedEvent>
+   public class ParameterIdentificationFeedbackViewVisibilityUICommand : ObjectUICommand<IParameterIdentificationFeedbackPresenter>
```
msevestre, line 8: there is in fact a better UI command for this.  ActveSubjectUIComand. Can you use this insteaD?

msevestre (changes requested):
see comment

src/OSPSuite.Presentation/UICommands/ParameterIdentificationFeedbackViewVisibilityUICommand.cs
abdelr: > ActveSubjectUIComand

I did not find ActveSubjectUIComand but instead I found ActiveObjectUICommand. The problem is that we need the command to implement ObjectUICommand<IParameterIdentificationFeedbackPresenter> but retrieve a ParameterIdentification as the active subject. So if I would extend ActiveObjectUICommand<IParameterIdentificationFeedbackPresenter> the Subject will be of type IParameterIdentificationFeedbackPresenter which is not what I need. To avoid this, I would need to change IParameterIdentificationFeedbackManager so it uses the cache based on the presenter, and not the object. This will imply to make the presenter to have a public getter for the object so we can access the object from the Subject (presenter). Do you like these changes?

msevestre: I don't understand. Why can't you be ActiveObjectUICommand<ParameterIdentificatoin>?

abdelr: It was about making the ParameterIdentification public on the presenter interface (IParameterIdentificationFeedbackPresenter). Sorry the description was not that clear. Check changes on the code for easier understanding.

msevestre: I am sorry but I don't understand how this works. Can we have a call?

abdelr: > I am sorry but I don't understand how this works. Can we have a call?

@msevestre Just call any time

msevestre (commented):
Very good

PR #1489, opened 2022-02-09
Generate Id for ParameterIdentificationFeedback object so it can be serialized

### LINKED ISSUE #1488: Saving a project in PKSim with Parameter Identification => Crash
@msevestre commented on [Fri Jan 14 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2050)

Application:
PK-Sim® 11.0.0

Unable to cast object of type 'OSPSuite.Presentation.Presenters.ParameterIdentifications.ParameterIdentificationFeedback' to type 'OSPSuite.Core.Domain.IWithId'.

Stack trace:
```
at OSPSuite.Utility.Extensions.ConversionExtensions.DowncastTo[T](Object objectToCast)
   at PKSim.Presentation.Services.WorkspaceLayoutUpdater.createNewWorkspaceLayoutWithOpenPresenters() in C:\dev\PK-Sim\src\PKSim.Presentation\Services\WorkspaceLayoutUpdater.cs:line 70
   at PKSim.Presentation.Services.WorkspaceLayoutUpdater.SaveCurrentLayout() in C:\dev\PK-Sim\src\PKSim.Presentation\Services\WorkspaceLayoutUpdater.cs:line 50
   at PKSim.Presentation.Services.ProjectTask.saveProjectToFile(String fileName) in C:\dev\PK-Sim\src\PKSim.Presentation\Services\ProjectTask.cs:line 181
   at PKSim.Presentation.Services.ProjectTask.SaveCurrentProject() in C:\dev\PK-Sim\src\PKSim.Presentation\Services\ProjectTask.cs:line 130
   at PKSim.Presentation.UICommands.SaveProjectCommand.Execute() in C:\dev\PK-Sim\src\PKSim.Presentation\UICommands\SaveProjectCommand.cs:line 17
   at OSPSuite.Presentation.UICommands.UICommandExtensions.ExecuteWithinExceptionHandler(IUICommand uiCommand, IEventPublisher eventPublisher, IChangePropagator changePropagator)
```

---

@msevestre commented on [Fri Jan 14 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2050#issuecomment-1013206338)

Due to new addition of PIFeedback object.. should probably not be saved in layout.

---

@msevestre commented on [Wed Feb 09 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2050#issuecomment-1033779751)

@abdelr This needs to be fixed to

abdelr:
Fixes #1488 (Saving project with parameter identification feedback view => crash)

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationFeedbackManager.cs
+++ b/src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationFeedbackManager.cs
@@ -1,15 +1,20 @@
-﻿using OSPSuite.Core.Domain.ParameterIdentifications;
+﻿using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.ParameterIdentifications;
 using OSPSuite.Core.Events;
+using OSPSuite.Utility;
 using OSPSuite.Utility.Collections;
 using OSPSuite.Utility.Events;
 
 namespace OSPSuite.Presentation.Presenters.ParameterIdentifications
 {
-   public class ParameterIdentificationFeedback
+   public class ParameterIdentificationFeedback : IWithId
    {
+      public string Id { get; set; }
+
       public ParameterIdentificationFeedback(ParameterIdentification parameterIdentification)
       {
          ParameterIdentification = parameterIdentification;
+         Id = ShortGuid.NewGuid();
       }
 
       public ParameterIdentification ParameterIdentification { get; private set; }


msevestre:
This line in PK-Sim should be changes
https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/develop/src/PKSim.Presentation/Services/WorkspaceLayoutUpdater.cs#L70

Instead of creating an Id that will not be registered anywhere. CLosing this PR as this will not work

abdelr:
> This line in PK-Sim should be changes https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/develop/src/PKSim.Presentation/Services/WorkspaceLayoutUpdater.cs#L70
> 
> Instead of creating an Id that will not be registered anywhere. CLosing this PR as this will not work

@msevestre Well, it did not crash anymore. The idea of saving the object was inspired by the fact that you mentioned that there is a way to save the project so you could resume it later in the exact state it was when it closed. So, for example, all dialogs that were opened should remain as they were. In such a case, we do need to have the object saved into the project, right?

msevestre:
For this the object needs to be registered. and available. By creating the ID by hand, it defeats the purpose of registration.

abdelr:
Ok so, should we register it and make it available or should we ignore it as you propose modifying the WorkspaceLayoutUpater? Also, I was considering some more changes to turn the subject into a IWithId, otherwise the downcast is missing some logic to return null in case it is not implementing the inteface.

msevestre:
no downacast. Jsut use as if it's null return? or what a I missing?

abdelr:
Could we talk?

abdelr:
The code on WorkspaceLayoutUpater already deal with null. The problem is that [DowncastTo](https://github.com/Open-Systems-Pharmacology/OSPSuite.Utility/blob/master/src/OSPSuite.Utility/Extensions/ConversionExtensions.cs#L20) should do something like:
    return objectToCast as T;
So it will return null if the object does not implement T. This is what I mean.

msevestre:
Yes don't use downcast. 
Instead as

PR #1508, opened 2022-02-22
Select the sheet on which the configuration is based on

### LINKED ISSUE #1409: Importer: Handling of "invalid" files

abdelr:
Fixes #1409 (Select the sheet on which the configuration is based on)

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Core/DataSourceFile.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataSourceFile.cs
@@ -12,10 +12,9 @@ namespace OSPSuite.Infrastructure.Import.Core
    {
       Cache<string, DataSheet> DataSheets { get; }
 	   string Path { get; set; }
-
       IDataFormat Format { get; set; }
-
       IList<IDataFormat> AvailableFormats { get; set; }
+      string FormatCalculatedFrom { get; set; }
    }
 
    public abstract class DataSourceFile : IDataSourceFile
@@ -34,6 +33,8 @@ namespace OSPSuite.Infrastructure.Import.Core
             Format = value.FirstOrDefault();
          }
       }
+
+      public string FormatCalculatedFrom { get; set; }
 
       protected DataSourceFile(IImportLogger logger)
       {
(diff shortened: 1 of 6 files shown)


src/OSPSuite.Infrastructure.Import/Core/DataSourceFile.cs
```diff
@@ -35,6 +34,8 @@ public IList<IDataFormat> AvailableFormats
          }
       }
 
+      public string FormatCalculatedFrom { get; set; }
```
msevestre, line 37: Can you add a comment explaining what this is (e.g. name of the xxx)

src/OSPSuite.Infrastructure.Import/Services/Importer.cs
```diff
@@ -107,7 +107,10 @@ public IDataSourceFile LoadFile(IReadOnlyList<ColumnInfo> columnInfos, string fi
          {
             dataSource.AvailableFormats = CalculateFormat(dataSource, columnInfos, metaDataCategories, x).ToList();
             if (dataSource.AvailableFormats.Any())
+            {
+               dataSource.FormatCalculatedFrom = x;
```
msevestre, line 111: Let's name that variable something else. x is fine in small lambda stuff. But we are in a foreach loop here. I want a real name. THis is what I mean with constant refactoring. This should have been caught before and I missed it. But it's never to late:)

msevestre (changes requested):
Looks good to me. Just small things and then it can be merged

msevestre (approved):

PR #1509, opened 2022-02-22
Dimension must also be inferred from headers

### LINKED ISSUE #1427: Importer should try to parse dimension/unit from header after selecting a column for "Measurement"

abdelr:
Fixes #1427 (Dimension should also be inferred from headers)


src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -515,6 +515,18 @@ public ToolTipDescription ToolTipDescriptionFor(int index)
          };
       }
 
+      private void setUnitAndDimension(ColumnMappingDTO model)
+      {
+         var supportedDimensions = _columnInfos.First(x => x.DisplayName == model.MappingName).SupportedDimensions;
```
msevestre, line 520: I feel that I have seens this code
`_columnInfos.First(x => x.DisplayName == model.MappingName` e.g. finding a column by display name so many times, that I am believe it needs to be refactored.
What about you create  cache with DisplayName keys instead of IREadOnlyList. That way you can simply do columnInfos[model.MappingName].

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -515,6 +515,18 @@ public ToolTipDescription ToolTipDescriptionFor(int index)
...
+         var unit = _format.ExtractUnitDescriptions(model.ExcelColumn, supportedDimensions);
+         if (unit.SelectedUnit != UnitDescription.InvalidUnit)
```
msevestre, line 522: if equal, return. Exit early

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -515,6 +515,18 @@ public ToolTipDescription ToolTipDescriptionFor(int index)
...
+         {
+            var mappingDataFormatParameter = (model.Source as MappingDataFormatParameter);
```
msevestre, line 524: are we sure this is a MappingDataFormatParameter. If yes, explain why as comment
I am seeing this type of casting with as also a lot in the code. This is very unusual to always know what to cast to and I believe there is a need for refactoring here as well

tests/OSPSuite.Presentation.Tests/Importer/Presenters/ColumnMappingPresenterSpecs.cs
```diff
@@ -445,7 +446,10 @@ public void the_unit_is_set_properly(string oldUnitDescription, string newUnitDe
          sut.SetDescriptionForRow(_model);
 
          //Assert
-         (_model.Source as MappingDataFormatParameter).MappedColumn.Unit.SelectedUnit.ShouldBeEqualTo(shouldUpdate ? newUnitDescription : oldUnitDescription);
+         var mappingDataFormat = (_model.Source as MappingDataFormatParameter);
+         mappingDataFormat.MappedColumn.Unit.SelectedUnit.ShouldBeEqualTo(shouldUpdate ? newUnitDescription : oldUnitDescription);
+         if (shouldUpdate && newUnitDescription != "?")
```
msevestre, line 451: what's with the `?` Don't we have a constant for it

tests/OSPSuite.Presentation.Tests/Importer/Presenters/ColumnMappingPresenterSpecs.cs
```diff
@@ -445,7 +446,10 @@ public void the_unit_is_set_properly(string oldUnitDescription, string newUnitDe
...
+            mappingDataFormat.MappedColumn.Dimension.HasUnit(mappingDataFormat.MappedColumn.Unit.SelectedUnit).ShouldBeTrue();
```
msevestre, line 452: mappingDataFormat.MappedColumn : create a variable and use it. It's referenced 3 times. This is going to make the test easier to understand

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
msevestre: (but not for this PR...it's a general observation)

abdelr: I think you are right. We definitely need a refactoring. The thing is DataFormatParameters have a lot of common use but some special features are concrete for mappings (time, measurement and error). I would say there could be a refactoring missing but I would like to postpone it for now.

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
msevestre: Cache is a ienumerable by default. It will be the same code everywhere except where you can simplify it drastically

abdelr:
@msevestre, @georgeDaskalakis ColumnInfos is now a `Cache<string, ColumnInfo>` and it is always created as `new Cache<string, ColumnInfo>(getKey: : x => x.DisplayName)`. This way we do not write anymore `_columnInfos.First(x => x.DisplayName == "Time")` but `_columnInfos["Time"]`. Notice there are several files touched so please, take a look and merge as soon as possible if the change is approved so we avoid collisions later.

msevestre:
I meant in the presenter. The columnInfo otherwise could have stayed as is.

tests/OSPSuite.Infrastructure.Tests/Import/DataSourceSpecs.cs
```diff
@@ -186,7 +186,7 @@ protected override void Context()
                      Name = "Time",
                      Unit = new UnitDescription("s")
                   },
-                  ColumnInfo = _columnInfos[0]
+                  ColumnInfo = _columnInfos.ElementAt(0)
```
msevestre, line 189: I think those indexes here should use keys instead. Also what is the displaName of the column info defined on line 175. It it name ? You are using it as key and it's not defined

tests/OSPSuite.Presentation.Tests/Importer/Presenters/ImporterPresenterSpecs.cs
```diff
@@ -169,7 +169,7 @@ protected static MetaDataCategory createMetaDataCategory<T>(string descriptiveNa
 
    public class When_setting_settings : concern_for_ImporterPresenter
    {
-      protected IReadOnlyList<ColumnInfo> _columnInfos = A.Fake<IReadOnlyList<ColumnInfo>>();
+      protected Cache<string, ColumnInfo> _columnInfos = A.Fake<Cache<string, ColumnInfo>>();
```
msevestre, line 172: Why a fake here? This does not make sense.

tests/OSPSuite.Starter/Tasks/ImporterConfigurationDataGenerator.cs
```diff
@@ -130,9 +131,9 @@ public IReadOnlyList<MetaDataCategory> DefaultGroupByTestMetaDataCategories()
          return categories;
       }
 
-      public IReadOnlyList<ColumnInfo> DefaultGroupByConcentrationImportConfiguration()
+      public Cache<string, ColumnInfo> DefaultGroupByConcentrationImportConfiguration()
       {
-         var columns = new List<ColumnInfo>();
+         var columns = new Cache<string, ColumnInfo>(getKey: x => x.DisplayName);
```
msevestre, line 136: Why do we have name and display name in columnInfo. I am getting a bit confused

tests/OSPSuite.UI.Tests/ColumnMappingControlSpecs.cs
```diff
@@ -709,7 +709,7 @@
 //             base.GlobalContext();
 //
 //             var dataRepositoryMapper = A.Fake<IImportDataTableToDataRepositoryMapper>();
-//             var columnInfos = A.Fake<IReadOnlyList<ColumnInfo>>();
+//             var columnInfos = A.Fake<Cache<string, ColumnInfo>>();
```
msevestre, line 712: Why are those tests commented out?

abdelr: Test file was left, removing it.

tests/OSPSuite.Starter/Tasks/ImporterConfigurationDataGenerator.cs
abdelr: I think they are not both needed since we always have them assigned to the same value. So we could create an issue to remove one of them (DisplayName I would say)

PR #1514, opened 2022-02-23
fix

georgeDaskalakis:
@PavelBal can you please test the behaviour with the dataset you had? Now the view stays enabled, but you cannot select datasets anymore (the last one remains plotted) and the import button becomes disabled. Also the error is displayed in both tabs.


georgeDaskalakis:
@msevestre you can check the behaviour in the Starter of Core - since it does not change anything about the behaviour after the data is imported. I fyou need any help on this just call me :)

src/OSPSuite.UI/Views/Importer/ImportConfirmationView.cs
```diff
@@ -26,6 +28,21 @@ public string SelectedSeparator
          get => separatorComboBoxEdit.SelectedText;
       }
 
+      public void DisableSelectingDataSets(string errorMessage)
```
msevestre, line 31: I don't like this. You are mixing two properties and that's why you are duplicating the call (look the first 3 lines are the same in the Enable and Disable)

You may need to methods, but it should be " bool SelectingDataSetsEnable {set}" (potentially get but you probably don't need it aside from tests)

And then you need a method "SetErrorMessage(string message)"

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
```diff
@@ -104,10 +104,10 @@ private void plotDataSet(object sender, DataSetSelectedEventArgs e)
          }
          catch (InvalidArgumentException invalidException)
          {
-            _view.DisableConfirmationView();
             var errors = new ParseErrors();
             errors.Add(_dataSource.DataSetAt(e.Index), new NonMonotonicalTimeParseErrorDescription(Error.ErrorWhenPlottingDataRepository(e.Index, invalidException.Message)));
```
msevestre, line 108: I know it does not come from this commit...but WHY is there this explicit `NonMonotonicalTimeParseErrorDescription` coming out of nowhere??

msevestre (changes requested):
ALso it feels that this "behavior" should be covered by tests at the presenter level at least

src/OSPSuite.UI/Views/Importer/ImportConfirmationView.cs
georgeDaskalakis: @msevestre  I like your suggestion way better! But then you mean listen to the propertyChanged to actually set the properties? If at some point you get 5 min time let's discuss this on a call

msevestre: No. I'll hope on a call with you later today

georgeDaskalakis: @msevestre call anytime :)

msevestre: I am on slack.

georgeDaskalakis:
@msevestre I have done the changes we discussed (also did add a getter for the test), and I think this should be ready now.

PR #1517, opened 2022-02-24
Fixed #1516 origin data from r

### LINKED ISSUE #1516: OriginData should be setable from R

msevestre:


src/OSPSuite.R/Services/DataImporterTask.cs
```diff
@@ -78,16 +78,8 @@ public bool IgnoreSheetNamesAtImport
...
-         return _dataImporter.ImportFromConfiguration(
-            GetConfiguration(configurationPath),
-            _metaDataCategories,
-            _columnInfos,
-            _dataImporterSettings,
-            dataPath
-         ).ToArray();
-      }
+         string dataPath) =>
+         ImportExcelFromConfiguration(GetConfiguration(configurationPath), dataPath);
```
msevestre, line 82: I am calling one overload from the other. It's typical to only have one implementation and the other are just calling the real implementation with other parameters (This is just a refactoring, nothing to do with the fix)

msevestre (commented):
@abdelr Please review when you have time. I will merge but you can still comment

abdelr (commented):
Looks good to me.

PR #1523, opened 2022-02-24
Fixes #1522 plus some code cleanup

### LINKED ISSUE #1522: Do not leak cache implementation to the API
Changes in latest PR with the column info are breaking encapsulation as callers need to know that the cache is created using display name and we have to do it in PKSim and MopBi. This should be internal

msevestre:


msevestre:
@abdelr the usage of this cache was leaking everywhere. I ensured that the API used by the callers do not know anything about this
Also I created a class so that we do not repeat the knowledge of how the cache is defined everywhere

src/OSPSuite.Presentation/Presenters/Importer/IImporterDataPresenter.cs
```diff
@@ -61,23 +61,24 @@ public interface IImporterDataPresenter : IPresenter<IImporterDataView>
...
       void SetTabMarks(ParseErrors errors, Cache<string, IDataSet> loadedDataSets);
       void SetTabMarks(ParseErrors errors);
-
    }
 
    public class TabMarkInfo
    {
-      public string ErrorMessage { get; private set; }
-      public bool IsLoaded { get; private set; }
-      public bool ContainsError { get => ErrorMessage != null; }
```
msevestre, line 76: Resharper suggestions...

msevestre (commented):
I don't like how this internal change is reflected everywhere. I feel that we are missing a layer of abstraction here but I am not sure what

msevestre:
Mergint to update PKSim and MoBI

PR #1526, opened 2022-03-02
1501 arithmetic error should follow measurement

### LINKED ISSUE #1501: Importer: Arithmetic error unit should follow measurement error.
### Example:
- Open the file - Select file [BookPercent.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/7824394/BookPercent.xlsx)
- Measurement dimension is "Fraction", unit is "%", Error is set to "Geometric"
- Changing error type to arithmetic requires user action to set the units:

image

### Expected behavior 
- When changing from "Geometric" to "Arithmetic" error type, error unit is automatically set to that of the measurement.

And I would go further:
- changing the dimension of the "Measurement" also changes the unit of the arithmetic error.
- Changing the unit only of measurement does not change the unit of arithmetic error (at this state, they are definitely compatible)
- If measurement unit is from column and a column is selected, changing "Geometric" to "Arithmetic" automatically pre-selects the same column for the units as the one selected for "Measurement" (currently, it must be explicitly set by the user).

abdelr:
Fix  #1501 (Arithmetic error should follow measurement)


tests/OSPSuite.Infrastructure.Tests/Import/MixColumnsDataFormatSpecs.cs
```diff
@@ -0,0 +1,119 @@
...
+            new ColumnInfo() { DisplayName = "Time", Name ="Time" },
+            new ColumnInfo() { DisplayName = "Concentration", Name = "Concentration"},
+            new ColumnInfo() { DisplayName = "Error", Name = "Error", IsMandatory = false, RelatedColumnOf = "Concentration"}
+         };
+         _columnInfos["Time"].SupportedDimensions.Add(DimensionFactoryForSpecs.TimeDimension);
+         _columnInfos["Concentration"].SupportedDimensions.Add(DimensionFactoryForSpecs.ConcentrationDimension);
+         _columnInfos["Error"].SupportedDimensions.Add(DimensionFactoryForSpecs.ConcentrationDimension);
+         _metaDataCategories = new List<MetaDataCategory>()
+         {
+            new MetaDataCategory() { Name = "Organ" },
```
msevestre, line 38: You do not need the () when initializing an object with props. It's just noise. Use reshaper to remove them

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -213,6 +216,15 @@ public void UpdateDescriptionForModel(MappingDataFormatParameter mappingSource)
          else //in this case the column is a measurement column
          {
             column.LloqColumn = _mappingParameterEditorPresenter.LloqFromColumn() ? _mappingParameterEditorPresenter.LloqColumn : null;
+            if (dimensionChanged)
```
msevestre, line 219: Move into a separate method. Also make sure if that have more than one line have { }

msevestre (commented):
I am not sure if I should review this. It's draft but I got a notification to review

abdelr:
> I am not sure if I should review this. It's draft but I got a notification to review

You may most certainly review but do not merge... it is a WIP yet.

src/OSPSuite.Infrastructure.Import/Core/ColumnInfo.cs
```diff
@@ -57,5 +59,10 @@ public ColumnInfoCache(IReadOnlyList<ColumnInfo> columnInfos) : this()
       {
          AddRange(columnInfos);
       }
+
+      public IEnumerable<ColumnInfo> RelatedColumnsFrom(string measurement)
```
msevestre, line 63: I think the parameter should be a ColumnInof here. you always call it using column.Name. 
And it also reads better: RelatedColumnFrom (anotherColumn)

src/OSPSuite.Infrastructure.Import/Extensions/ColumnInfoExtensions.cs
```diff
@@ -13,5 +13,10 @@ public static bool IsAuxiliary(this ColumnInfo info)
       {
          return !info.IsMandatory;
       }
+
+      public static bool IsMeasurement(this ColumnInfo info)
```
msevestre, line 17: Why are those guys not defined in ColumnInfo (I mean all the extensions)?

src/OSPSuite.Infrastructure.Import/Core/ColumnInfo.cs
abdelr: > I think the parameter should be a ColumnInof here.

Not really so. For example in ColumnMappingPresenter.cs#L168, column is not a ColumnInfo but a Column (although Column also has a string Name getter). I do like renaming the parameter to anotherColumn though.

src/OSPSuite.Infrastructure.Import/Extensions/ColumnInfoExtensions.cs
abdelr: > Why are those guys not defined in ColumnInfo (I mean all the extensions)?

Furthermore, they should be getters. Moving logic already ;)

PR #1528, opened 2022-03-04
1482 group by column

### LINKED ISSUE #1482: Observed data import: Error
Open the CSV file [prefiltered_testcsv.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/8014219/prefiltered_testcsv.zip) and load the configuration [Config01_withFilter.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/8014252/Config01_withFilter.zip)

Confirm the warnng with Yes (s. #1481) and click "Add current sheet". The error below comes:

```
Application:
PK-Sim® 10.0.257

An item with the key 'Dose' has already been added
-----------------------------------------------------
An item with the same key has already been added.

Stack trace:
at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at OSPSuite.Utility.Collections.Cache`2.Add(TKey key, TValue value)
 --- End of inner exception stack trace ---
   at OSPSuite.Utility.Collections.Cache`2.Add(TKey key, TValue value)
   at OSPSuite.Utility.Collections.Cache`2.Add(TValue value)
   at OSPSuite.Infrastructure.Import.Core.Mappers.DataSetToDataRepositoryMapper.ConvertImportDataSet(ImportedDataSet dataSet)
   at OSPSuite.Presentation.Presenters.Importer.ImporterPresenter.plotDataSet(Object sender, DataSetSelectedEventArgs e)
   at System.EventHandler`1.Invoke(Object sender, TEventArgs e)
   at OSPSuite.Presentation.Presenters.Importer.ImportConfirmationPresenter.DataSetSelected(Int32 index)
   at OSPSuite.UI.Views.Importer.ImportConfirmationView.onDataSetNameSelected(Object sender, EventArgs eventArgs)
   at OSPSuite.UI.Controls.BaseUserControl.<>c__DisplayClass26_0`1.<OnEvent>b__0()
```

The reason is probably that we have a "Dose" column which is not mapped to "Dose" but used as a group by instead and another column mapped to "Dose":
grafik

georgeDaskalakis:
closes #1482

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
@@ -51,7 +51,7 @@ namespace OSPSuite.Infrastructure.Import.Core
       void SetDataFormat(IDataFormat dataFormat);
       void SetNamingConvention(string namingConvention);
       ParseErrors AddSheets(Cache<string, DataSheet> dataSheets, ColumnInfoCache columnInfos, string filter);
-      void SetMappings(string fileName, IEnumerable<MetaDataMappingConverter> mappings);
+      void SetMappings(string fileName, IList<MetaDataMappingConverter> mappings);
       ImporterConfiguration GetImporterConfiguration();
       IEnumerable<MetaDataMappingConverter> GetMappings();
       Cache<string, IDataSet> DataSets { get; }
@@ -66,7 +66,7 @@ namespace OSPSuite.Infrastructure.Import.Core
    {
       private readonly IImporter _importer;
       private readonly ImporterConfiguration _configuration;
-      private IEnumerable<MetaDataMappingConverter> _mappings;
+      private IReadOnlyList<MetaDataMappingConverter> _mappings;
       public Cache<string, IDataSet> DataSets { get; } = new Cache<string, IDataSet>();
 
       public DataSource(IImporter importer)
@@ -134,10 +134,10 @@ namespace OSPSuite.Infrastructure.Import.Core
          return errors;
       }
 
-      public void SetMappings(string fileName, IEnumerable<MetaDataMappingConverter> mappings)
+      public void SetMappings(string fileName, IList<MetaDataMappingConverter> mappings)
       {
          _configuration.FileName = fileName;
-         _mappings = mappings;
+         _mappings = (IReadOnlyList<MetaDataMappingConverter>)mappings;
       }
 
       public ImporterConfiguration GetImporterConfiguration()
--- a/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
@@ -51,7 +51,7 @@ namespace OSPSuite.Infrastructure.Import.Core
       void SetDataFormat(IDataFormat dataFormat);
       void SetNamingConvention(string namingConvention);
       ParseErrors AddSheets(Cache<string, DataSheet> dataSheets, ColumnInfoCache columnInfos, string filter);
-      void SetMappings(string fileName, IList<MetaDataMappingConverter> mappings);
+      void SetMappings(string fileName, IReadOnlyList<MetaDataMappingConverter> mappings);
       ImporterConfiguration GetImporterConfiguration();
       IEnumerable<MetaDataMappingConverter> GetMappings();
       Cache<string, IDataSet> DataSets { get; }
@@ -134,10 +134,10 @@ namespace OSPSuite.Infrastructure.Import.Core
          return errors;
       }
 
-      public void SetMappings(string fileName, IList<MetaDataMappingConverter> mappings)
+      public void SetMappings(string fileName, IReadOnlyList<MetaDataMappingConverter> mappings)
       {
          _configuration.FileName = fileName;
-         _mappings = (IReadOnlyList<MetaDataMappingConverter>)mappings;
+         _mappings = mappings;
       }
 
       public ImporterConfiguration GetImporterConfiguration()
--- a/src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
@@ -227,7 +227,8 @@ namespace OSPSuite.Presentation.Presenters.Importer
          (
             dataSourceFile.Format.Parameters.OfType<GroupByDataFormatParameter>().Select(md => new MetaDataMappingConverter()
             {
-               Id = dataMappings.Any(x=> x.Id == md.ColumnName) ? md.ColumnName + Constants.ImporterConstants.GroupingBySuffix : md.ColumnName,
+               //in case of a duplicate name coming from an excel column used as a grouping by with the same name as a metaData, we add a suffix 
+               Id = dataMappings.ExistsById(md.ColumnName) ? md.ColumnName + Constants.ImporterConstants.GroupingBySuffix : md.ColumnName,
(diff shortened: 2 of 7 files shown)


georgeDaskalakis:
@msevestre regarding unit testing - the only reasonable possibility is to add an integration test. I considered it an overkill though to add a file and a configuration actually just for this edge case - but if you think we better have one do say. It should be quite fast to add.

msevestre:
I would add a test, especially for bug fixing and edge cases...when/if we refactor, we will not think about this use case again

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -51,7 +51,7 @@ public interface IDataSource
       void SetDataFormat(IDataFormat dataFormat);
       void SetNamingConvention(string namingConvention);
       ParseErrors AddSheets(Cache<string, DataSheet> dataSheets, ColumnInfoCache columnInfos, string filter);
-      void SetMappings(string fileName, IEnumerable<MetaDataMappingConverter> mappings);
+      void SetMappings(string fileName, IList<MetaDataMappingConverter> mappings);
```
msevestre, line 54: Why not IReadOnlyList here?

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -134,10 +134,10 @@ public ParseErrors AddSheets(Cache<string, DataSheet> dataSheets, ColumnInfoCach
          return errors;
       }
 
-      public void SetMappings(string fileName, IEnumerable<MetaDataMappingConverter> mappings)
+      public void SetMappings(string fileName, IList<MetaDataMappingConverter> mappings)
       {
          _configuration.FileName = fileName;
-         _mappings = mappings;
+         _mappings = (IReadOnlyList<MetaDataMappingConverter>)mappings;
```
msevestre, line 140: especially because you are casting back to IReadOnlyList... what? :)

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
```diff
@@ -216,18 +216,22 @@ private void loadSheets(IDataSourceFile dataSourceFile, Cache<string, DataSheet>
...
+               Id = md.MetaDataId,
+               Index = sheetName => md.IsColumn ? dataSourceFile.DataSheets[sheetName].RawData.GetColumnDescription(md.ColumnName).Index : -1
+            }).ToList();
+            
+         var mappings   = dataMappings.Union
          (
             dataSourceFile.Format.Parameters.OfType<GroupByDataFormatParameter>().Select(md => new MetaDataMappingConverter()
             {
-               Id = md.ColumnName,
+               Id = dataMappings.Any(x=> x.Id == md.ColumnName) ? md.ColumnName + Constants.ImporterConstants.GroupingBySuffix : md.ColumnName,
```
msevestre, line 230: We have extensions methods such as ExistsById etc... 
I think a comment explaining WHY you are doing the concat here makes sense (*e..g because of mapping to a column already mapped etc)

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
```diff
@@ -216,18 +216,22 @@ private void loadSheets(IDataSourceFile dataSourceFile, Cache<string, DataSheet>
...
-         {
-            Id = md.MetaDataId,
-            Index = sheetName => md.IsColumn ? dataSourceFile.DataSheets[sheetName].RawData.GetColumnDescription(md.ColumnName).Index : -1
-         }).Union
+         var dataMappings = dataSourceFile.Format.Parameters.OfType<MetaDataFormatParameter>().Where(p => p.ColumnName != null).Select(md =>
+            new MetaDataMappingConverter()
+            {
+               Id = md.MetaDataId,
+               Index = sheetName => md.IsColumn ? dataSourceFile.DataSheets[sheetName].RawData.GetColumnDescription(md.ColumnName).Index : -1
+            }).ToList();
```
msevestre, line 224: no need for a tolist here since you are going to create a union right away after

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
```diff
@@ -216,18 +216,22 @@ private void loadSheets(IDataSourceFile dataSourceFile, Cache<string, DataSheet>
...
-         var mappings = dataSourceFile.Format.Parameters.OfType<MetaDataFormatParameter>().Where(p => p.ColumnName != null).Select(md => new MetaDataMappingConverter()
-         {
-            Id = md.MetaDataId,
-            Index = sheetName => md.IsColumn ? dataSourceFile.DataSheets[sheetName].RawData.GetColumnDescription(md.ColumnName).Index : -1
-         }).Union
+         var dataMappings = dataSourceFile.Format.Parameters.OfType<MetaDataFormatParameter>().Where(p => p.ColumnName != null).Select(md =>
+            new MetaDataMappingConverter()
+            {
+               Id = md.MetaDataId,
+               Index = sheetName => md.IsColumn ? dataSourceFile.DataSheets[sheetName].RawData.GetColumnDescription(md.ColumnName).Index : -1
```
msevestre, line 223: -1 again.  @abdelr and I talked about this. I don't like those magic values popping up everwhere. 
also `dataSourceFile.DataSheets[sheetName].RawData.GetColumnDescription(md.ColumnName)` type of code is everywhere. I think we need to create helper method here, maybe in the dataSourceFile to simplify all our a.b.c.d.

msevestre (changes requested):
Code improvement and some weird stuff going on with IList. I am going to create a separate issue for bigger changes

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
georgeDaskalakis: corrected

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
georgeDaskalakis: true, corrected

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
georgeDaskalakis: Main reason for this is that without the ToList() Resharper gives me a warning that there is a possible multiple enumeration. So I thought it is safer this way. But again it can easily be that I am missing something...

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
georgeDaskalakis: MetaDataMappingConverter was not implementing IWithId (although it did have an Id), that's why I could not find the method in autocomplete. This is a good tip, I need to keep learning things like this that we have already implemented :) .
Also comment added

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
georgeDaskalakis: yeap, we need a new issue for this :)

src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs
msevestre: ok

PR #1530, opened 2022-03-04
Importer performs better when loading several datasets

### LINKED ISSUE #1492: Importer should perform better when loading several datasets
Check performance when not checking the filter active on [1491](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1491)

abdelr:
Fix #1492 (Importer performs better when loading several datasets)

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
@@ -234,31 +234,31 @@ namespace OSPSuite.Infrastructure.Import.Core.DataFormat
          return buildDataSets(data, groupingCriteria, columnInfos);
       }
 
-      private string rowId(IEnumerable<(string ColumnName, IReadOnlyList<string> ExistingValues)> parameters, IUnformattedData data, UnformattedRow row)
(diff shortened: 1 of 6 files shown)


abdelr:
Definitely, there was a silly performance issue when importing the data. If we now try to load the original data and apply the filters, it only takes a few seconds instead of 2,5+h. The main difference is that we do not traverse the data for each possible data set filtering the whole file once and again but instead we separate the rows first, and just parse the data that has been provided on each cluster. No new test is written since there are plenty of tests checking that data loads correctly and all those tests are still valid.

abdelr:
I am adding the original data here again, keep in mind that by applying the filter you end up with erroneous data (non monotonical time) but still the process is much faster and tests are green.
[testcsv+filters.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/8186053/testcsv%2Bfilters.zip)

abdelr:
Also, the code is much easier to read now I would say ;)

msevestre:
I need to chat with you to understand what's going on here. Can we have a quick call?

msevestre:
Build fails,,,

msevestre:
https://ci.appveyor.com/project/open-systems-pharmacology-ci/ospsuite-core/builds/42790166

abdelr:
@msevestre fixed now

PR #1531, opened 2022-03-04
Resetting sheet marks on reset format based on current sheet

### LINKED ISSUE #1503: Importer: Sheet status does not get updated
[CompiledDataSet.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/8097038/CompiledDataSet.xlsx)

- Open the attached file
- Add the first sheet (TestSheet_1)
- Select the second sheet (UnitsFromColumn)
- "Reset mapping based on current sheet" -> "This action will result in dropping..." -> "Yes"
- The loaded data sets are dropped, but the green checkmark for "TestSheet_1" remeains:
image

- Set units for the second sheet (does not matter which units, just to be able to add the second sheet)
- Error , but the green checkmark disappears.

abdelr:
Fix #1503 (Resetting sheet marks on reset format based on current sheet)

### DIFF
--- a/src/OSPSuite.UI/Views/Importer/ImporterDataView.cs
+++ b/src/OSPSuite.UI/Views/Importer/ImporterDataView.cs
@@ -350,10 +350,5 @@ namespace OSPSuite.UI.Views.Importer
          refreshErrorMessage();
          refreshErrorMarks();
       }
-
-      public void ClearTabMarks()
-      {
-         SetTabMarks(new Cache<string, TabMarkInfo>(onMissingKey: _ => new TabMarkInfo(errorMessage: null, isLoaded: false)));
-      }
    }
 }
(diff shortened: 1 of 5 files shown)


src/OSPSuite.UI/Views/Importer/ImporterDataView.cs
```diff
@@ -350,5 +350,10 @@ public void SetTabMarks(Cache<string, TabMarkInfo> tabMarks)
          refreshErrorMessage();
          refreshErrorMarks();
       }
+
+      public void ClearTabMarks()
+      {
+         SetTabMarks(new Cache<string, TabMarkInfo>(onMissingKey: _ => new TabMarkInfo(errorMessage: null, isLoaded: false)));
```
msevestre, line 356: the info on how the cache is created is leaking here. Probably the presenter should be the one to call SetTabMarks with the right init (or just clear the cache that it has locally presumably),. This can be merged but this could be optimized a bIT

msevestre (approved):
suggestion for improvement. If you don't see it like that, please merge

src/OSPSuite.UI/Views/Importer/ImporterDataView.cs
abdelr: I think you are right... It will be enough with setting the marks again from the presenter.

PR #1548, opened 2022-03-14
Some WO for #1547

msevestre:


src/OSPSuite.UI/UIConstants.cs
```diff
@@ -39,24 +39,23 @@ public static class Size
          public static readonly int EMPTY_SPACE_HEIGHT = ScaleForScreenDPI(20);
          public static readonly int RADIO_GROUP_HEIGHT = ScaleForScreenDPI(24);
          public static readonly int LARGE_BUTTON_WIDTH = ScaleForScreenDPI(150);
-         //DO NOT SCALE HEIGHT WITH DPI of BUTTONS
-         public static readonly int LARGE_BUTTON_HEIGHT = 29;
+         public static readonly int LARGE_BUTTON_HEIGHT = ScaleForScreenDPI(29);
```
msevestre, line 42: @Yuri05 I am now scaling everything with a DPI and I added another scale factor for very large DPI ratio

msevestre (commented):
Merging to see how it behavs in PK-Sim

PR #1552, opened 2022-03-17
1546 changing meta data malfunctions

### LINKED ISSUE #1546: Importer: Changing a mapped metaData results in error
Steps to reproduce:

1. Open a file (eg in the case of this screenshot [Data.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/8242860/Data.xlsx) )
2. Select for "Route" one mapping, fe excel column Group.
3. Change the selection  for "Route" to another mapping, fe excel column Route.
The new selection results in the following error:

image

georgeDaskalakis:
closes #1546 

@msevestre sorry, running the ReSharper formatter makes the review a bit harder. But at least the formating in the files is correct that way :)

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
@@ -590,11 +590,14 @@ namespace OSPSuite.Presentation.Presenters.Importer
          mappingDataFormatParameter.MappedColumn.Dimension = supportedDimensions.FirstOrDefault(x => x.HasUnit(unit.SelectedUnit));
       }
 
+      //ToDo: this big switch statement here underneath should be refactored, probably broken to more than one functions.
+      //ToDo: also the part of the switch(model.Source) that also then in the cases changes the model.Source is kind of murky. 
       private void setDescriptionForRow(ColumnMappingDTO model, bool isColumn)
       {
          switch (model.Source)
          {
-            case null:
+            //this is the case for the first setting of Mapping or MetaData. GroupBy and AddGroupBy are never null.
+            case null: 
                switch (model.CurrentColumnType)
                {
                   case ColumnMappingDTO.ColumnType.MetaData:
@@ -611,12 +614,17 @@ namespace OSPSuite.Presentation.Presenters.Importer
 
                _format.Parameters.Add(model.Source);
                break;
-            case AddGroupByFormatParameter _:
+            //this is the case for setting a new GroupBy parameter. The AddGroupByFormatParameter corresponds to the line in the ColumnMapping Grid
+            //that has the "+" button for adding the GroupBy, that's why in this case we are introducing a new Parameter, instead of changing the existing.
+            //The AddGroupByFormatParameter, should never be changed, as the "+" button line should always present in the Grid
+            case AddGroupByFormatParameter _: 
                model.Source = new GroupByDataFormatParameter(model.ExcelColumn);
                _format.Parameters.Add(model.Source);
                setDataFormat(_format.Parameters);
                break;
-            default:
+            //this is the case that corresponds to the changing of the mapping for a Parameter.
+            //The Parameter in this case can be Mapping, MetaData or GroupBy
+            default: 
             {
                model.Source.ColumnName = model.ExcelColumn;
                switch (model.CurrentColumnType)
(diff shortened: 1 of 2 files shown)


src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -523,14 +533,15 @@ public IEnumerable<ColumnMappingOption> GetAvailableOptionsFor(ColumnMappingDTO
...
-      private IEnumerable<string> 
+      private IEnumerable<string>
          availableColumns()
       {
          return _format.ExcelColumnNames
             .Where
             (
                cn =>
-                  _format.Parameters.OfType<MappingDataFormatParameter>().All(p => p.ColumnName != cn && p.MappedColumn?.Unit?.ColumnName != cn && p.MappedColumn?.LloqColumn != cn) &&
+                  _format.Parameters.OfType<MappingDataFormatParameter>().All(p =>
```
msevestre, line 543: This statement needs to be rewritten at some point. I have absolutely NO idea what this means. So many ANDs, NOT, without comment. Let's be kind to own future selves and not do that anymore

msevestre:
Unfortuntely I do not see what has changed that is related to the actual issue at hand. When so much formatting is going on, it would be helpful for me if you could point out the actual changes. So I am going to ask for revision for now until I see what I need to review.

msevestre (commented):
I cannot review this changes at the moment as running resharper changed way too much. That's why running resharper before each PUSH is something we should try to do

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
georgeDaskalakis: OK, I ll add a todo here for now and we can talk about refactoring such code.

georgeDaskalakis:
> I cannot review this changes at the moment as running resharper changed way too much. That's why running resharper before each PUSH is something we should try to do

@msevestre  you mean doing a PR just with the format changes seperately right? Yeap, I agree with that.

tests/OSPSuite.Presentation.Tests/Importer/Presenters/ColumnMappingPresenterSpecs.cs
```diff
@@ -545,6 +562,108 @@ public void the_unit_and_dimension_are_set_properly(string oldUnitDescription, s
...
+      }
+
+      [Observation]
+      public void the_meta_data_row_should_be_correctly_added()
+      {
+         var metaDataFormat = (_secondModel.Source as MetaDataFormatParameter);
+         metaDataFormat.ColumnName.ShouldBeEqualTo("Patient id");
+      }
+   }
+
```
georgeDaskalakis, line 666: @msevestre in this file, only the tests written in these lines here are new

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -572,45 +584,57 @@ private void setUnitAndDimension(ColumnMappingDTO model)
...
+                     break;
+                  default:
+                     throw new NotImplementedException($"Setting description for unhandled column type: {model.CurrentColumnType}");
+               }
+
+               break;
             }
-            setUnitAndDimension(model);
          }
       }
```
georgeDaskalakis, line 640: @msevestre here it is only this function SetDescriptionForRow that has been changed. The actual fix here is the moving setUnitAndDimension(model) call to the correct place. And I have also refactored a bit the switch/if...then..else statements  according to ReSharper suggestions.

georgeDaskalakis:
> Unfortuntely I do not see what has changed that is related to the actual issue at hand. When so much formatting is going on, it would be helpful for me if you could point out the actual changes. So I am going to ask for revision for now until I see what I need to review.

@msevestre I have made comments on the 2 places that the code has really changed. Not sure if it is easier for you that I marked the whole code, or if it generally easier to mark just the first 2-3 lines and explain, but please give me feedback so that I can do in a way that works better for you next time :)

georgeDaskalakis:
@msevestre I did not refactor the function that we discussed yesterday, I am in the middle of that bigger refactoring with DataSheetCollection, and I think I cannot approach both concurrently with the necessary caution. I have though left a todo there for the refactoring, along with detailed comments of what every case in the switch does. But this PR should be ready now. 

We could open an issue where we kind of put together things that we have to refactor, what do you think?

msevestre:
>We could open an issue where we kind of put together things that we have to refactor, what do you think?

yes please. Can you do it? Merging this

PR #1553, opened 2022-03-20
Better BaseModalView layout

msevestre:


msevestre:
@abdelr @georgeDaskalakis @Yuri05 
I have spent too much time trying to make those freaking buttons work with the previous implementation. It just does not work. So I am trying a new approach with a table layout. Looks very good on my tests. I will update PK-Sim ad see how it behaves

PR #1554, opened 2022-03-20
Fixes #1551 get mearurement column

### LINKED ISSUE #1551: `GetMeasurementColumn` from `DataRepositoryTask`
Required for this:

https://github.com/Open-Systems-Pharmacology/OSPSuite-R/issues/825

msevestre:


src/OSPSuite.R/Services/DataRepositoryTask.cs
```diff
@@ -27,6 +27,13 @@ public interface IDataRepositoryTask
       /// <param name="fileName">Full path where the <paramref name="dataRepository" /> will be saved</param>
       void SaveDataRepository(DataRepository dataRepository, string fileName);
 
+      /// <summary>
+      ///    Returns the first instance of a measurement column (e.g. ColumnOrigin is Observation) defined in the repository of
+      ///    null if not found.
+      ///    <exception cref="OSPSuiteException">is thrown if there are more than one measurement column in the data repository</exception>
+      /// </summary>
+      DataColumn GetMeasurementColumn(DataRepository dataRepository);
```
msevestre, line 35: @PavelBal This is implemented. Throwing exception if we have multiplied observed data column . This should never be the case

PR #1557, opened 2022-03-21
Fixes #1556 mol weight for molecule name

### LINKED ISSUE #1556: `Simulation$molWeightFor` should accept molecule entity as path
`Simulation$molWeightFor` only works for quantities, i.e., the instance of the molecule in any container. It is not possible to get the MW of the molecule entity that is located on the top-level:

```
sim <- loadTestSimulation("S1")
path <- "Caffeine"
sim$molWeightFor(path)
[1] NA
```

msevestre:


src/OSPSuite.Core/Domain/Model.cs
```diff
@@ -80,15 +69,31 @@ public string MoleculeNameFor(IQuantity quantity)
...
-         var quantity = Root?.EntityAt<IQuantity>(quantityPath.ToPathArray());
-         return MolWeightFor(quantity);
+         var pathArray = quantityPath.ToPathArray();
+
+         //we have a real quantity path
+         if (pathArray.Length > 1)
+            return MolWeightFor(Root?.EntityAt<IQuantity>(pathArray));
+
+         //this is potentially just the molecule name
+         return molWeightFor(quantityPath);
```
msevestre, line 84: @PavelBal This is the change.

PavelBal (approved):

PR #1561, opened 2022-03-22
Fixes #1558 extract header unit

### LINKED ISSUE #1558: Add method to extra units from header
based on Core implementation

msevestre:


src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -154,18 +154,9 @@ protected virtual void ExtractQualifiedHeadings(List<string> keys, List<string>
...
+     protected string GetAndValidateUnitFromBrackets(string unit, IReadOnlyList<IDimension> supportedDimensions)
       {
-         return Regex.Match(header, @"\[([^\]\[]*)\]$").Value;
-      }
-
-      protected string GetAndValidateUnitFromBrackets(string units, IReadOnlyList<IDimension> supportedDimensions)
-      {
-         return units
-            .Trim().Substring(1, units.Length - 2).Trim() //remove the brackets and whitespaces from end and beginning
-            .Split(',')                                   //we split in case there are more than one units separated with ,
```
msevestre, line 166: @Yuri05 @PavelBal @georgeDaskalakis @abdelr 
What is this here? Split unit with a "," ? Where did we specify something like this?

msevestre (commented):
This takes the code that was implemented to extract unit, puts it outside so that it can be reused. Adds tests to support some new requirements. Fixes the implementation to support the tests
And export the method to be used in R

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
PavelBal: I cannot remember us specifying something like this.

Yuri05: We had already a discussion in the past about the unit extraction, and I thought the `.Split(',')` was removed...
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/pull/1204#pullrequestreview-712614163

msevestre: well actually Yuri proposes a solution without regex which reads much much better. I am going to implement it. 
When you solve a problem with regex, you have two problems :)

PR #1580, opened 2022-04-06
1511 error unit reset fixed

### LINKED ISSUE #1511: Importer: Adding grouping resets error unit selection
- Open [Data.xlsx](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/8124568/Data.xlsx) for import
- Error unit is set to "mg/l" (I would expect it to be inferred from the header, i.e., "ng/ml"... but summarizing all the expected behavior in https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1501)
- Change error unit to "ng/ml"
- "Add group by" select column "Group"
- The error unit is reset to "mg/l"!

georgeDaskalakis:
closes #1511

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
@@ -5,7 +5,6 @@ using OSPSuite.Assets;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.UnitSystem;
 using OSPSuite.Core.Import;
-using OSPSuite.Core.Services;
 using OSPSuite.Infrastructure.Import.Core;
 using OSPSuite.Infrastructure.Import.Services;
 using OSPSuite.Presentation.Views.Importer;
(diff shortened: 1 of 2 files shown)


src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -125,10 +124,17 @@ public void InitializeErrorUnit()
...
 
-         if (measurementColumn != null)
+         if (measurementColumn != null && unitNotSet(errorColumn))
             errorColumn.Unit = measurementColumn.Unit;
       }
 
+      private bool unitNotSet(Column column)
+      {
+         //when unit is not set and also columnName is not set
+         return (column.Unit.SelectedUnit == "?" || column.Unit.SelectedUnit == null)
```
msevestre, line 134: null first

msevestre (approved):
ok. just swap conditions

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
georgeDaskalakis: done

PR #1589, opened 2022-04-18
1588 deleting empty observed data folder fixed

georgeDaskalakis:
closes 1588

### DIFF
--- a/src/OSPSuite.Core/Domain/Services/ObservedDataTask.cs
+++ b/src/OSPSuite.Core/Domain/Services/ObservedDataTask.cs
@@ -73,6 +73,9 @@ namespace OSPSuite.Core.Domain.Services
       {
          var observedDataToRemoveList = observedDataToBeRemoved.ToList();
 
+         if (observedDataToRemoveList.Count < 1) 
+            return true;
+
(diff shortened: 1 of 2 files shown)


src/OSPSuite.Core/Domain/Services/ObservedDataTask.cs
```diff
@@ -73,6 +73,9 @@ public bool Delete(IEnumerable<DataRepository> observedDataToBeRemoved, bool sil
       {
          var observedDataToRemoveList = observedDataToBeRemoved.ToList();
 
+         if (observedDataToRemoveList.Count < 1)
```
msevestre, line 76: use Any() for this. Is the list empty or not basically

msevestre: and while you are here, can you check for null? 

e.g. var observedDataToRemoveList = observedDataToBeRemoved?.ToList();

if(observedDataToRemoveList ==null || !observedDataToRemoveList.Any())
    return true

This should never happen but who knows
And add a test for this

msevestre (changes requested):

PR #1594, opened 2022-04-20
Fixes #805 molecule name parent container

### LINKED ISSUE #805: MoleculeNameFor: check if ParentContainer is a molecule amount before returning its name
If the quantity is not a molecule amount: the current implementation returns the name of the parent container without further checks.
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/fe4aaf1206d2f58c10b8c70e1c0c3e42978cdbd4/src/OSPSuite.Core/Domain/Model.cs#L78-L86
This will work for all molecule amounts and most (all?) observers (whose parent container is usually a molecule amount).
However for e.g. _plottable parameters_ (which in general are not molecule related) the parent container can be organ/compartment/whatever.

Before returning the parent container name, I would check:
* Is parent container a molecule amount.
* If not: is the name of the parent container in the list of all molecules available in the simulation

If both is false: return empty name; otherwise return the name of the parent container.

msevestre:


src/OSPSuite.Core/Domain/Model.cs
```diff
@@ -60,16 +63,23 @@ public IContainer Neighborhoods
...
 
-         return quantity.IsAnImplementationOf<IMoleculeAmount>() ? quantity.Name : quantity.ParentContainer?.Name;
+         switch (entity)
+         {
+            case IMoleculeAmount _:
+            case IContainer container when container.ContainerType == ContainerType.Molecule:
+               return entity.Name;
+         }
+
+         return MoleculeNameFor(entity.ParentContainer);
```
msevestre, line 82: @Yuri05 Instead of your proposed solution, it feels like we can make it a big more generic by traversing the tree recursively and returning the name of the first container (including self) that is either a MoleculeAmount or a Container of type Molecule. 
Thoughts?

src/OSPSuite.Core/Domain/Model.cs
```diff
@@ -60,16 +63,23 @@ public IContainer Neighborhoods
...
       {
-         if (quantity == null)
+         if (entity == null)
             return string.Empty;
 
-         return quantity.IsAnImplementationOf<IMoleculeAmount>() ? quantity.Name : quantity.ParentContainer?.Name;
+         switch (entity)
+         {
+            case IMoleculeAmount _:
+            case IContainer container when container.ContainerType == ContainerType.Molecule:
```
msevestre, line 78: pretty nifty switch with a filtering statement.

src/OSPSuite.Core/Domain/Model.cs
Yuri05: Probably OK

Yuri05 (approved):

PR #1603, opened 2022-04-21
1602 remove unused properties

msevestre:


src/OSPSuite.Core/Domain/Data/DataInfo.cs
```diff
@@ -61,12 +61,6 @@ public class DataInfo
       /// </summary>
       public DateTime Date { get; set; }
 
-      /// <summary>
-      ///    For observed data it is normally the name of the excel sheet or csv file. For simulated data,
-      ///    it could be the name of the simulation
-      /// </summary>
-      public string Source { get; set; }
```
msevestre, line 68: Remove this from the Data Info.. Not only is it saved in the data repo, but also in each column!

src/OSPSuite.Core/Domain/Data/DataInfo.cs
```diff
@@ -61,12 +61,6 @@ public class DataInfo
       /// </summary>
       public DateTime Date { get; set; }
```
msevestre, line 62: Why do we need the Date? I would remove this too. Why do we care when this column was imported? (in the project there is a history and there is also the configuration)

src/OSPSuite.Core/Domain/Data/QuantityInfo.cs
```diff
@@ -26,19 +21,17 @@ public class QuantityInfo
...
       {
       }
 
       /// <summary>
       ///   Create a new quantity info
       /// </summary>
-      /// <param name = "name">Name of quantity but could be set to sthg else. Only used in DataColumn.QuantityName</param>
       /// <param name = "quantityPath">Full path of quantity</param>
       /// <param name = "quantityType">Type of quantity (such as species, observer, parameter). </param>
-      public QuantityInfo(string name, IEnumerable<string> quantityPath, QuantityType quantityType)
```
msevestre, line 39: Remote the name of the quantity info which is ALWAYS set to the name of the DataColumn except in our tests. Since it can only be created programmatically, remove

src/OSPSuite.Infrastructure.Import/Core/Mappers/DataSetToDataRepositoryMapper.cs
```diff
@@ -190,27 +193,16 @@ private static DataColumn findColumnByName(IEnumerable<DataColumn> columns, stri
...
-      private static void addExtendedPropertyForSource(string fileName, string sheetName,  DataRepository dataRepository)
+
+      private static void addExtendedPropertyForSource(string fileName, string sheetName, DataRepository dataRepository)
       {
-         if (!string.IsNullOrEmpty(fileName))
-         {
-            var sourceProperty =
-               Activator.CreateInstance(typeof(ExtendedProperty<>).MakeGenericType(fileName.GetType()))
-                  as IExtendedProperty;
-            if (sourceProperty != null)
```
msevestre, line 200: WTF. I have no idea what this code is doing so I removed it for something that makes sense
I commented the code out because I do not want to save the source anymore

msevestre (commented):
I would also remove the Date. Thoughts?

msevestre:
All right, removing Date as well and merging

PR #1605, opened 2022-04-21
Fixes #1604 use parameter name. Also add time reference

### LINKED ISSUE #1604: Use  SA parameter name instead of path when retrieving output values

msevestre:


src/OSPSuite.Core/Domain/SensitivityAnalyses/OutputParameterSensitivity.cs
```diff
@@ -17,6 +17,12 @@ public class OutputParameterSensitivity
       /// </summary>
       public float[] OutputValues { get; }
 
+      /// <summary>
+      ///    Simulated time for the given <see cref="OutputPath" />
+      /// </summary>
+      public float[] TimeValues { get; }
```
msevestre, line 23: All right  I have added a TimeValues. At the moment, each output is going to get a reference to this array. Becaus each results is seen as a different individuals, we will have one for each variation. Probably something to optimize since they should normally all be the same. But good enough for now

src/OSPSuite.Core/Domain/SensitivityAnalyses/SensitivityAnalysisRunResult.cs
```diff
@@ -28,25 +28,25 @@ public IEnumerable<PKParameterSensitivity> AllPKParameterSensitivitiesFor(string
...
 
-      public PKParameterSensitivity PKParameterSensitivityFor(string pkParameterName, string outputPath, string parameterName)
+      public PKParameterSensitivity PKParameterSensitivityFor(string pkParameterName, string outputPath, string sensitivityParameterName)
       {
-         return AllPKParameterSensitivitiesFor(pkParameterName, outputPath).Find(x => string.Equals(parameterName, x.ParameterName));
+         return AllPKParameterSensitivitiesFor(pkParameterName, outputPath).Find(x => string.Equals(sensitivityParameterName, x.ParameterName));
       }
 
-      public OutputParameterSensitivity[] OutputParameterSensitivitiesFor(string outputPath, string parameterPath)
+      public OutputParameterSensitivity[] OutputParameterSensitivitiesFor(string outputPath, string sensitivityParameterName)
```
msevestre, line 42: using sensitivyParameterName here instead of parmaeterpaht

Yuri05: name would be something like "Volume"? or like "Organism|Bone|Volume" (so in fact the path)?

msevestre: no name is the SenstitiveyParameterName (something like Liver-Volume). I am not sure why we are doing it lije this
I would vote to create new method to work with path instead of name everywhere to be honest

src/OSPSuite.Core/Domain/Constants.cs
```diff
@@ -346,6 +346,7 @@ public static class SensitivityAnalysisResults
          public const string PARAMETER = "Parameter";
          public const string VALUE = "Value";
          public const string PK_PARAMETER = "PKParameter";
+         public const string PARAMETER_PATH = "ParameterPath";
```
msevestre, line 349: @Yuri05 I am also adding the ParameterPath to the export as the Parameter may not been enough

src/OSPSuite.Core/Domain/SensitivityAnalyses/SensitivityAnalysisRunResult.cs
```diff
@@ -28,25 +28,46 @@ public IEnumerable<PKParameterSensitivity> AllPKParameterSensitivitiesFor(string
             x => string.Equals(x.QuantityPath, outputPath) && string.Equals(x.PKParameterName, pkParameterName));
       }
 
-      public IEnumerable<OutputParameterSensitivity> AllOutputParameterSensitivitiesFor(string outputPath, string parameterPath)
+      public IEnumerable<OutputParameterSensitivity> AllOutputParameterSensitivitiesBySensitivityParameterName(string outputPath, string sensitivityParameterName)
```
msevestre, line 31: @PavelBal @Yuri05 I am renaming this for clarification
Two way. By SenstitivyPArameteRName or by ParameterPath

msevestre (commented):
Let me know what yuo think
We will have to update ospsuite-R but this makes sense I think

src/OSPSuite.Core/Domain/SensitivityAnalyses/SensitivityAnalysisRunResult.cs
PavelBal: I prefer path over the name, too. Path is what we define in the first place when starting SA.

PavelBal (approved):

Yuri05 (approved):

Issue #1606, opened 2022-04-21
Random digits added to imported datapoints


msevestre:
### Discussed in https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/discussions/1601

<div type='discussions-op-text'>

<sup>Originally posted by **moPellowe** April 21, 2022</sup>
I've noticed that random digits get added to imported data points with decimals in PK-Sim.

I imported an excel sheet (attached below) into PK-Sim for a generic compound (I just loaded the Aciclovir template), and when I look at the data within the building block, there are digits added that weren't in the original xlsx file.

![image](https://user-images.githubusercontent.com/57540473/164476337-296d43b8-1902-4367-95f8-b53e718ecd26.png)
![image](https://user-images.githubusercontent.com/57540473/164476408-a8f35ef7-5c14-40a7-9f49-912c703d2df2.png)

[test.xlsx](https://github.com/Open-Systems-Pharmacology/Forum/files/8531646/test.xlsx)

</div>

PR #1608, opened 2022-04-21
WIP #1606 random digits

msevestre:


msevestre:
This does not help. The Factor and Offset are double. When I define them as float, some basic test such as 60min in s does not return 3600 but 3599.9998

Very weird all in all

msevestre:
We are going to mitigate the issue by displaying them as float instead of double I believe .But all in all, we have some weird stuff going on (won't change the model output whatsoever but still_

Yuri05 (approved):

PR #1611, opened 2022-04-21
1609 disable frame for exported Figures

### LINKED ISSUE #1609: Disable frame for exported Figures (Copy to Clipbord)
Would that be possible? That the very OUTER frame is not shown, or we can select the Color = None in the Utilities?

### LINKED ISSUE #1610: Copy to clipboard: remove gray filling around plot

georgeDaskalakis:
closes #1609
closes #1610


src/OSPSuite.UI/Extensions/ChartControlExtensions.cs
```diff
@@ -99,6 +110,8 @@ public static void CopyToClipboard(this UxChartControl chartControl, string wate
          {
             copyFontAndSizeSettings(chartControl, cloneOfChartControl);
             AddWatermark(cloneOfChartControl, watermark);
+            prepareChartForCopying(cloneOfChartControl);
```
msevestre, line 113: I am wondering why you are not doing this line on line 104
In fact, I think this should be changed a bit. Why do we have something like this?
`chartControl.CopyChartToClipboard(chartControl);`

It should be `chartControl.CopyChartToClipboard();`

and why do we have
`chartControl.CopyChartToClipboard(cloneOfChartControl);`

it should be
`cloneOfChartControl.CopyChartToClipboard();`

So somehow now that we are modifying the chart, we always need to clone it, update font and settings etc from the clone, and call this method (prepareChartForCopying)

And we need to verify for barcharts to make sure this works as expected. (as there was some issue it seems with devbepxress at some point, maybe the bug is corrected)

msevestre (changes requested):
While we are at it, I would try to make the code better
AT any rate, I think the code is not working as expected when No watermark is provided

src/OSPSuite.UI/Extensions/ChartControlExtensions.cs
georgeDaskalakis: I have changed 
`chartControl.CopyChartToClipboard(cloneOfChartControl);`

to 
`cloneOfChartControl.CopyChartToClipboard();`

this was actually written a bit weird till now, I agree. 

I have also changed the code for this to also be done if we do not set the watermark - this second function here is only being used for PI Time Analysis and Sensitivity analysis, this is why I did not find out the change was not working without watermark for these cases.

georgeDaskalakis: @msevestre so now we need to create the clone anyways, since we need to change the border of the image. 

tbh the version of the function for the ChartView (line 37), does not check if the watermark is null or empty, and still works fine without a watermark. If you agree I can remove the check, since it seems redundant.

Regarding the barcharts that are also mentioned in the ToDo, I am not sure how to reproduce this or what exactly the bug is...let's talk about this (also maybe solve in a separate issue )

msevestre: Bar chart are in PKSIm
image

For this, there is no watermark.

msevestre: Line 37 has a Chart as parameter. and there, there is always a Watermark. The other plot is independent from the Chart (IChart instance) not the control

If you can try with PKSim and we see that it works again, then we can remove OTherwise we need to keep

georgeDaskalakis: @msevestre so the watermark part ( when no watermark is specified) is being handled in AddWatermark function, in lines124 and then 130-131. So I will remove this check from both functions, it is not necessary.

georgeDaskalakis:
@msevestre this is ready now. I have removed the changes for the bar chart, meaning that for those charts there will be no changes to what we previously had: grey back color and frame + with watermark the settings get meesed up.

For the other charts everything works fine.

msevestre:
@georgeDaskalakis Can you crate an issue to solve the Histrogram problem?

src/OSPSuite.UI/Extensions/ChartControlExtensions.cs
```diff
@@ -36,33 +37,55 @@ public static void SetFontAndSizeSettings(this ChartControl chartControl, ChartF
...
             if (chart.IncludeOriginData)
                AddOriginData(cloneOfChartControl, chart);
 
-            AddWatermark(cloneOfChartControl, watermark, chart);
+            prepareChartForCopying(cloneOfChartControl);
 
-            chartControl.CopyChartToClipboard(cloneOfChartControl);
+            if (!string.IsNullOrEmpty(watermark))
+            {
+               //Bug in devexpress that does not keep the color in bar charts
```
msevestre, line 51: Do we agree that this comment is at the wrong place here as we have a chart of type IChart? Please move at the right place

msevestre (changes requested):
I think there is some left over

src/OSPSuite.UI/Extensions/ChartControlExtensions.cs
```diff
@@ -36,33 +37,55 @@ public static void SetFontAndSizeSettings(this ChartControl chartControl, ChartF
...
+            ms.Seek(0, SeekOrigin.Begin);
 
-         setFontAndSizeSettings(targetChartControl, width, height, fontTitle, fontDescription, fontAxis, fontLegend);
+            using (var mf = new Bitmap(ms))
+               Clipboard.SetImage(mf);
+         }
+      }
+
+      //returns whether the color was changed or not
+      private static bool prepareChartForCopying(ChartControl cloneOfChartControl)
```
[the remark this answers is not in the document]
georgeDaskalakis: true, was left over from the intermediate changes

georgeDaskalakis:
> @georgeDaskalakis Can you crate an issue to solve the Histrogram problem?

@msevestre  done, https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1622

georgeDaskalakis:
@msevestre I would say this is ready now

PR #1617, opened 2022-04-25
Support one data point

msevestre:


src/OSPSuite.Core/Domain/Data/DataColumn.cs
```diff
@@ -137,6 +137,14 @@ public virtual double[] ValuesAsArray
          set => Values = value?.ToFloatArray();
       }
 
+      /// <summary>
+      ///    Required for R (one value being set is not found properly by rCLR)
+      /// </summary>
+      public virtual double Value
```
msevestre, line 143: @PavelBal Adding this guy here for R compatibility

PR #1625, opened 2022-04-29
Fixes #1624 parameter path in export should be in quote

### LINKED ISSUE #1624: Parameter export should put parameter path in quotes

msevestre:


msevestre:
@Yuri05 Trying to debug the issue in PKSim. Of course, this creates problems because of , in our parameter names. This was missing

PR #1627, opened 2022-05-03
1626 missing unit warning fixed

### LINKED ISSUE #1626: PK-Sim 11: Missing unit warnings during data import
When importing data into PKSim 11: When deleting the automatic time entry mapping with the "x" on the right side and then selecting a column, one can forget to choose a unit without getting the red "x" error symbol and without the import being blocked. Thet results in an error during import but not a helpfull one. The red "x" from PK-Sim 10 in the mapping as well as the inactivation of the "Add current sheet" / " Add All sheets" button was way more helpful.
PK-Sim 10:
image

PK-Sim 11:
image

georgeDaskalakis:
closes #1626

### DIFF
--- a/src/OSPSuite.Core/Import/Column.cs
+++ b/src/OSPSuite.Core/Import/Column.cs
@@ -62,13 +62,15 @@ namespace OSPSuite.Core.Import
          return $"{Name} [{Unit}]";
       }
 
-      //unit is missing if the SelectedUnit is undefined and 1. the error is not geometric or 2. the mapped column is empty
       public bool MissingUnitMapping()
       {
-         return (Unit.SelectedUnit == UnitDescription.InvalidUnit &&
+         return 
+            //manual selection: selected unit not set and the column is not a Geometric Standard Deviation
+            (Unit.SelectedUnit == UnitDescription.InvalidUnit &&
                   (ErrorStdDev == null || ErrorStdDev.Equals(Constants.STD_DEV_ARITHMETIC)))
-                  ||
-                  ((Unit.SelectedUnit == null || Unit.SelectedUnit == UnitDescription.InvalidUnit) &&
+            ||
+            //select from a column: selected unit is null (not set) and no column has been selected
+            (Unit.SelectedUnit == null  &&
                   Unit.ColumnName == string.Empty);
       }
    }
--- a/src/OSPSuite.Core/Import/Column.cs
+++ b/src/OSPSuite.Core/Import/Column.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.UnitSystem;
 
 namespace OSPSuite.Core.Import
@@ -59,6 +60,16 @@ namespace OSPSuite.Core.Import
       public override string ToString()
       {
          return $"{Name} [{Unit}]";
+      }
+
+      //unit is missing if the SelectedUnit is undefined and 1. the error is not geometric or 2. the mapped column is empty
+      public bool MissingUnitMapping()
+      {
+         return (Unit.SelectedUnit == UnitDescription.InvalidUnit &&
+                  (ErrorStdDev == null || ErrorStdDev.Equals(Constants.STD_DEV_ARITHMETIC)))
+                  ||
+                  ((Unit.SelectedUnit == null || Unit.SelectedUnit == UnitDescription.InvalidUnit) &&
+                  Unit.ColumnName == string.Empty);
       }
    }
 }
--- a/src/OSPSuite.Infrastructure.Import/Services/Importer.cs
+++ b/src/OSPSuite.Infrastructure.Import/Services/Importer.cs
@@ -58,7 +58,10 @@ namespace OSPSuite.Infrastructure.Import.Services
       private readonly IDataSetToDataRepositoryMapper _dataRepositoryMapper;
       private readonly IDimension _molWeightDimension;
 
-      public Importer(IoC container, IDataSourceFileParser parser, IDataSetToDataRepositoryMapper dataRepositoryMapper,
+      public Importer(
+         IoC container, 
+         IDataSourceFileParser parser, 
+         IDataSetToDataRepositoryMapper dataRepositoryMapper,
          IDimensionFactory dimensionFactory)
       {
          _container = container;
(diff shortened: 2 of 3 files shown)


src/OSPSuite.Infrastructure.Import/Services/Importer.cs
```diff
@@ -173,16 +178,25 @@ public MappingProblem CheckWhetherAllDataColumnsAreMapped(ColumnInfoCache dataCo
...
+            //all the mandatory mappings that have not been mapped to a column
             MissingMapping = dataColumns
                .Where(col => col.IsMandatory && subset.All(cm =>
                   cm.MappedColumn.Name != col.Name)).Select(col => col.Name)
                .ToList(),
-            MissingUnit = subset.Where(cm => cm.MappedColumn.Unit.SelectedUnit == UnitDescription.InvalidUnit && (cm.MappedColumn.ErrorStdDev == null || cm.MappedColumn.ErrorStdDev.Equals(Constants.STD_DEV_ARITHMETIC))).Select(cm => cm.MappedColumn.Name)
+            //all the mappings where the unit is undefined and 1. the error is not geometric or 2. the mapped column is empty
+            MissingUnit = subset
+               .Where(
+                  cm => cm.MappedColumn.Unit.SelectedUnit == UnitDescription.InvalidUnit &&
```
msevestre, line 189: This is so hard to read. This needs to be changed.
It looks to me like this could be a property of the MappedColumn instead of constantly checking properties on the object

src/OSPSuite.Infrastructure.Import/Services/Importer.cs
```diff
@@ -56,15 +58,17 @@ public class Importer : IImporter
       private readonly IDataSetToDataRepositoryMapper _dataRepositoryMapper;
       private readonly IDimension _molWeightDimension;
 
-      public Importer(IoC container, IDataSourceFileParser parser, IDataSetToDataRepositoryMapper dataRepositoryMapper, IDimensionFactory dimensionFactory)
+      public Importer(IoC container, IDataSourceFileParser parser, IDataSetToDataRepositoryMapper dataRepositoryMapper,
```
msevestre, line 61: Let's align all the parameters in the constructor sonthat we habe one by line

src/OSPSuite.Infrastructure.Import/Services/Importer.cs
```diff
@@ -173,16 +178,25 @@ public MappingProblem CheckWhetherAllDataColumnsAreMapped(ColumnInfoCache dataCo
...
          return new MappingProblem()
          {
+            //all the mandatory mappings that have not been mapped to a column
             MissingMapping = dataColumns
                .Where(col => col.IsMandatory && subset.All(cm =>
                   cm.MappedColumn.Name != col.Name)).Select(col => col.Name)
                .ToList(),
-            MissingUnit = subset.Where(cm => cm.MappedColumn.Unit.SelectedUnit == UnitDescription.InvalidUnit && (cm.MappedColumn.ErrorStdDev == null || cm.MappedColumn.ErrorStdDev.Equals(Constants.STD_DEV_ARITHMETIC))).Select(cm => cm.MappedColumn.Name)
+            //all the mappings where the unit is undefined and 1. the error is not geometric or 2. the mapped column is empty
+            MissingUnit = subset
```
msevestre, line 187: Aside from the  code hard to understand, I don't understand why this fixes the issue.
You are checking now for the case where the unit was mapped to another column right? But in the issue, this was dealing with unit being not selected. How does it relate? The first condition should still have been Invalid no?

georgeDaskalakis: No the Selected unit is set to invalid by creation. The problem is that also "get unit by column" was selected, but no column was mapped (ColumnName = ""). This is why this solves the problem

src/OSPSuite.Infrastructure.Import/Services/Importer.cs
georgeDaskalakis: done

src/OSPSuite.Infrastructure.Import/Services/Importer.cs
georgeDaskalakis: @msevestre moved it to the Column class

src/OSPSuite.Core/Import/Column.cs
```diff
@@ -60,5 +61,15 @@ public override string ToString()
       {
          return $"{Name} [{Unit}]";
       }
+
+      //unit is missing if the SelectedUnit is undefined and 1. the error is not geometric or 2. the mapped column is empty
+      public bool MissingUnitMapping()
+      {
```
msevestre, line 67: This has moved, probably where it belongs, but this is still super hard to read. 
Or do you find this expression acceptable?
Asking @Yuri05 for feedback here. In my opinion, this was accepted by mistake previously

georgeDaskalakis: @msevestre I have updated the comments and removed an unnecessary check in the second case. Unofrtunately in the first case we need both checks: we just need to maske sure the Error is not standard geometric deviation, but we also need to check for null, otherwise we get an exception in startup. I would say in this updated version it should not be hard to read....

msevestre (approved):

src/OSPSuite.Core/Import/Column.cs
Yuri05: I would extract `(ErrorStdDev == null || ErrorStdDev.Equals(Constants.STD_DEV_ARITHMETIC))` into a new method `IsArithmeticStd `or `IsArithmeticError`
(and may be also put some comments there e.g. explaining why `ErrorStdDev == null` means "arithmetic".)

Also I would rename the method to `IsUnitMappingMissing()`

msevestre: I think the problem is also the lack of consistency within the method. Sometimes we use ==, sometimes we use Equals, the comment specifies `geometric` but we are checking for arithmetic. 

We need to do ourselves favors and stop those multiple combinations of `&&`  `||`  `((()` etc.. There is absolutely no reason to have this complexity and also no reason to try to return all in one expression. The code should read as a story, and for me it does not read at all

msevestre:
I have merged because we need to ship. I'll not merge this anymore in the future or I will refer to a higher power to merge :)

Yuri05:
>I will refer to a higher power...

higher power like **JL**? 😉

msevestre:
I meant you;-)

msevestre:
I meant you;-)

And for the record, this was my mistake. I should have reviewed it more thoroughly when it was implemented a while back. I'll do my best to be more careful in the future

PR #1640, opened 2022-06-07
1637 grouping of simulated results

### LINKED ISSUE #1637: Default grouping of simulated results and observed data in Chart Editor
Initial discussion in https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/discussions/1628

The default layout of the Chart Editor should group by the "Origin" column:

In PI time profiles:
image

- "Origin" maybe can be renamed to e.g. "Data origin" or "Data source"?
- Can the prefix "Origin" be removed here?:
image

- "CalculationAuxiliary" should be called "Simulation"

In simulation:
image

- "Calculation" should be called "Simulation"
- "BaseGrid" should not be shown at all!

@msevestre Happy to hear your comments.

georgeDaskalakis:
closes #1637

### DIFF
--- a/src/OSPSuite.Core/Chart/ChartSettings.cs
+++ b/src/OSPSuite.Core/Chart/ChartSettings.cs
@@ -16,7 +16,10 @@ namespace OSPSuite.Core.Chart
 
    public enum GridGroupRowFormats
    {
-      HideColumnName
+      HideColumnName,
+
+      ///return the default format of the grid grouping row 
+      Default
    }
 
    public class ChartSettings : Notifier, IUpdatable
--- a/src/OSPSuite.Core/Chart/ChartSettings.cs
+++ b/src/OSPSuite.Core/Chart/ChartSettings.cs
@@ -19,20 +19,6 @@ namespace OSPSuite.Core.Chart
       HideColumnName
    }
 
-   public static class GridGroupRowFormatsExtensions
-   {
-      public static string GetFormatString(this GridGroupRowFormats formatName)
-      {
-         switch (formatName)
-         {
-            case GridGroupRowFormats.HideColumnName:
-               return "[#image]{1} {2}";
-            default:
-               return "{0}: [#image]{1} {2}"; //return the default format of the grid grouping row 
-         }
-      }
-   }
-
    public class ChartSettings : Notifier, IUpdatable
    {
       private bool _sideMarginsEnabled;
--- a/src/OSPSuite.Core/Chart/ChartSettings.cs
+++ b/src/OSPSuite.Core/Chart/ChartSettings.cs
@@ -14,6 +14,11 @@ namespace OSPSuite.Core.Chart
       BottomInside
    }
 
+   public static class GridGroupRowFormats
+   {
+      public const string HideColumnName = "[#image]{1} {2}";
+   }
+
    public class ChartSettings : Notifier, IUpdatable
    {
       private bool _sideMarginsEnabled;
--- a/src/OSPSuite.Core/Chart/ChartSettings.cs
+++ b/src/OSPSuite.Core/Chart/ChartSettings.cs
@@ -14,12 +14,26 @@ namespace OSPSuite.Core.Chart
       BottomInside
    }
 
-   public static class GridGroupRowFormats
+   public enum GridGroupRowFormats
    {
-      public const string HideColumnName = "[#image]{1} {2}";
+      HideColumnName
    }
 
-   public class ChartSettings : Notifier, IUpdatable
+   public static class GridGroupRowFormatsExtensions
+   {
+      public static string GetFormatString(this GridGroupRowFormats formatName)
+      {
+         switch (formatName)
+         {
+            case GridGroupRowFormats.HideColumnName:
+               return "[#image]{1} {2}";
+            default:
+               return "{0}: [#image]{1} {2}"; //return the default format of the grid grouping row 
+         }
+      }
+   }
+
+      public class ChartSettings : Notifier, IUpdatable
    {
       private bool _sideMarginsEnabled;
       private LegendPositions _legendPosition;
--- a/src/OSPSuite.Core/Chart/ChartSettings.cs
+++ b/src/OSPSuite.Core/Chart/ChartSettings.cs
@@ -33,7 +33,7 @@ namespace OSPSuite.Core.Chart
       }
    }
 
-      public class ChartSettings : Notifier, IUpdatable
+   public class ChartSettings : Notifier, IUpdatable
    {
       private bool _sideMarginsEnabled;
       private LegendPositions _legendPosition;
--- a/src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
@@ -57,6 +57,11 @@ namespace OSPSuite.Presentation.Presenters.Charts
       void AddDataRepositories(IEnumerable<DataRepository> dataRepositories);
 
       /// <summary>
+      /// sets the group row format in the gridView of the DataBrowserView to the specified string.
+      /// </summary>
+      void SetGroupRowFormat(string format);
+
+      /// <summary>
       ///    Returns all DataColumns used in the DataBrowser
       /// </summary>
       IReadOnlyList<DataColumn> AllDataColumns { get; }
@@ -467,6 +472,11 @@ namespace OSPSuite.Presentation.Presenters.Charts
          _dataBrowserPresenter.AddDataColumns(allColumnsToAdd);
 
          updateUsedColumns();
+      }
+
+      public void SetGroupRowFormat(string format)
+      {
+         _dataBrowserPresenter.SetGroupRowFormat(format);
       }
 
       private bool hasColumn(DataColumn dataColumn) => _dataBrowserPresenter.ContainsDataColumn(dataColumn);
--- a/src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
@@ -186,6 +186,10 @@ namespace OSPSuite.Presentation.Presenters.Charts
       /// </summary>
       void Refresh();
 
+      /// <summary>
+      ///    sets the Group Values (Dictionary Key) in the DataBrowser Grid for which the group title should be
+      ///    replaced by the given Dictionary Value. 
+      /// </summary>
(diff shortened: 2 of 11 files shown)


msevestre:
All of this seems to be very convoluted to me for just "Setting up" the header. 
The usage of a Dictionary<string, string> feels wrong to me, specifically because we have a type that you can use as a key (ColumnOrigins) which you always use and cast... as a string

Last, there is the event in the view that is going to be fired all the time to replace the text. Looks also a bit suspicious. I'll go in details in the PR but I think we need to review this

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -56,6 +56,11 @@ public interface IChartEditorPresenter : IPresenter<IChartEditorView>,
       /// </summary>
       void AddDataRepositories(IEnumerable<DataRepository> dataRepositories);
 
+      /// <summary>
+      /// sets the group row format in the gridView of the DataBrowserView to the specified string.
+      /// </summary>
+      void SetGroupRowFormat(string format);
```
msevestre, line 62: I do not know what you need this at this level of present.er Don't we have access to the underlying DAtaBrowser presenter with a property get?

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -649,6 +665,11 @@ public void Refresh()
          updateUsedColumns();
       }
 
+      public void SetGroupRowNamingReplacements(Dictionary<string, string> groupingNamesReplacementDictionary)
```
msevestre, line 668: THe way you defined this is not consistent. Here it's a function, in the databrowser, it's a property.
Why?

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
```diff
@@ -84,6 +83,17 @@ public interface IDataBrowserPresenter : IPresenter<IDataBrowserView>, IPresente
       ///    Returns all the DataColumns for the curves that are visible in the chart
       /// </summary>
       IReadOnlyList<DataColumn> GetAllUsedDataColumns();
+
+      /// <summary>
+      /// sets the group row format of the gridView to the specified string.
+      /// </summary>
+      void SetGroupRowFormat(string format);
```
msevestre, line 90: I fill this format should be an enum or sthg. You just cannot pass anything there can you?

src/OSPSuite.Presentation/Views/Charts/IDataBrowserView.cs
```diff
@@ -16,5 +17,16 @@ public interface IDataBrowserView : IView<IDataBrowserPresenter>, IViewWithColum
...
       /// </summary>
       IReadOnlyList<DataColumnDTO> SelectedDescendantColumns { get; }
+
+      /// <summary>
+      /// sets the group row format of the gridView to the specified string.
+      /// </summary>
+      void SetGroupRowFormat(string format);
+
+      /// <summary>
+      ///    Dictionary that specifies the Group Values (Dictionary Key) in the DataBrowser Grid for which the group title should be
```
msevestre, line 27: So,ething is off here with the comments

msevestre: If feels that this is going to be always the one. Do we need to have this?
In fact, it feels like we want to rename the columns all the time? Why don't we do this at binding time ?

src/OSPSuite.UI/Views/Charts/DataBrowserView.cs
```diff
@@ -40,6 +40,20 @@ public DataBrowserView()
          InitializeWith(gridView);
          _gridViewBinder = new GridViewBinder<DataColumnDTO>(gridView);
          _usedRepository = new UxRepositoryItemCheckEdit(gridView);
+
+         
+         gridView.CustomDrawGroupRow += (s, e) =>
```
msevestre, line 45: Create a function instead of implementing on the constructor

src/OSPSuite.UI/Views/Charts/DataBrowserView.cs
```diff
@@ -40,6 +40,20 @@ public DataBrowserView()
...
+         {
+            var info = e.Info as GridGroupRowInfo;
+
+            foreach (var replacementPair in GroupingNamesReplacementDictionary)
```
msevestre, line 49: This is very, very inefficient. I know you only have one entry in the group at the moment, but it's a key value pair where key is the way to look for an entry. Right now, you are treating it as a simple array, looping over everything, not breaking when found. 
I am not sure that we need this method but if we do, it needs to be rewritten.e.g.
something along the line of
```
if(GroupingNamesReplacementDictionary.HasKey((info.GroupValueText))
   info.GroupText = replacementPair.Value;
```

no foreach loop.

msevestre (changes requested):

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
georgeDaskalakis: Not really. We have access to the ChartEditorPresenter (meaning we do not have to go through the ChartPresenter too), but the DataBrowserPresenter is private to the ChartEditorPresenter. If you think it is better, of course I can expose the DataBrowserPresenter....

src/OSPSuite.UI/Views/Charts/DataBrowserView.cs
georgeDaskalakis: I have changed this. I have also originally thought of implementing this as  List<string, string> or with Tuples (or a new class for that matter.). We could briefly discuss this today in the meeting.

src/OSPSuite.Presentation/Views/Charts/IDataBrowserView.cs
georgeDaskalakis: Yes this is actually always called (I was under the impression the event was triggered once, when the grouping was happening). I have used this event according to the devexpress documentation. We are not really renaming the column name, but the value of the "Data Origin" column when the grouping is happening: "CalculationAuxiliary" -> "Simulation". 

I have thought a lot about a possible workaround to this, but as far as I understand the only other possible solution is to handle the gridView.EndGrouping event that does actually get triggered at the end of the grouping. The problem here (as far as I understand always ) is that traversing the grouping rows is not that straightforward. Again, let's discuss this in the meeting

georgeDaskalakis:
@msevestre I have changed this PR according to what we discussed yesterday. I would say now it is ready for review again. I have resolved all the conversations, since not much of the code has remained the same.

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationAnalysisChartPresenter.cs
```diff
@@ -15,25 +13,30 @@
...
 using OSPSuite.Presentation.Views;
 using OSPSuite.Presentation.Views.ParameterIdentifications;
+using OSPSuite.Utility.Collections;
+using OSPSuite.Utility.Extensions;
 
 namespace OSPSuite.Presentation.Presenters.ParameterIdentifications
 {
-   public abstract class ParameterIdentificationAnalysisChartPresenter<TChart, TView, TPresenter> : SimulationAnalysisChartPresenter<TChart, TView, TPresenter>, IParameterIdentificationAnalysisPresenter
+   public abstract class
+      ParameterIdentificationAnalysisChartPresenter<TChart, TView, TPresenter> : SimulationAnalysisChartPresenter<TChart, TView, TPresenter>,
```
msevestre, line 22: Let's put the name on the same line as the class . This looks just too weird

src/OSPSuite.Core/Chart/ChartSettings.cs
```diff
@@ -14,6 +14,25 @@ public enum LegendPositions
       BottomInside
    }
 
+   public enum GridGroupRowFormats
+   {
+      HideColumnName
+   }
+
+   public static class GridGroupRowFormatsExtensions
```
msevestre, line 22: Neat idea. Let's move this away from ChartSettings in the generic UI stuff 
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/tree/develop/src/OSPSuite.UI/Extensions

I would also create a constant for the default format in the enum (e.g. HideColumnName, Standard ) and put a comment with /// that will be shown on intellisense.
Then it would be cool to look for all the magic string in code and replace with this.

Last you can do it a bit differently also by creating your own enum on steroid. I
See https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/develop/src/PKSim.Core/Model/ApplicationType.cs#L7

This is obviously less complicated but you can add as many properties to your enum as you want. Just food for thoughts

msevestre (commented):
Minor changes. Looking good

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationAnalysisChartPresenter.cs
georgeDaskalakis: done

src/OSPSuite.UI/Extensions/GridGroupRowFormatsExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+   public static class GridGroupRowFormatsExtensions
+   {
+      public static string GetFormatString(this GridGroupRowFormats formatName)
+      {
+         switch (formatName)
+         {
+            case GridGroupRowFormats.HideColumnName:
+               return "[#image]{1} {2}";
+            default:
+               return Captions.Chart.GroupRowFormat.GridGroupingRowDefaultFormat;
```
msevestre, line 15: sorry I was not clear. Create an ENUM Values for this like you have with the other one
also the enum 

   public enum GridGroupRowFormats
   {
      HideColumnName
   }

needs to be moved here as well. 
No need to create a constant for `Captions.Chart.GroupRowFormat.GridGroupingRowDefaultFormat` This has nothing to do with chart. Just defined it here in line, it's fine as you do for the HideColumnNam,e. Just create a a default in ten enum that's all

```
   public enum GridGroupRowFormats
   {
      HideColumnName
      Default


   }

msevestre (changes requested):

src/OSPSuite.UI/Extensions/GridGroupRowFormatsExtensions.cs
georgeDaskalakis: OK, now I see how you mean. Done.

PR #1644, opened 2022-06-22
1641 linking mapped outputs in charts

### LINKED ISSUE #1641: PI and Simulation Charts: De-/Selecting a simulation output de-/selects the mapped observed data
After the grouping of simulation outputs and observations according to 

https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1637

it is proposed to implement a new check box that will link the (de)-selection of simulation outputs with their linked observed data. When the checkbox  “Link data to simulations” is selected, selecting or de-selecting a simulation output automatically selects (or de-selects) the observed data linked to that simulation. When the checkbox gets set to true, all the mappings will be updated so that only the observed data that is mapped to selected simulation outputs will also be selected.

image

georgeDaskalakis:
closes #1641


src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -103,6 +104,11 @@ public interface IChartEditorPresenter : IPresenter<IChartEditorView>,
       /// </summary>
       void SetDisplayQuantityPathDefinition(Func<DataColumn, PathElements> displayQuantityPathDefinition);
 
+      /// <summary>
+      /// Sets the output mappings of Simulation Outputs and their mapped Observed Data
+      /// </summary>
+      void SetOutputMappingNames(Cache<string, List<string>> outputObservedDataPairs);
```
msevestre, line 110: I think it will be much easier to understand if you have a dedicated object for this. You just have a bunch of strings mapped to string and are dealing with this all the time in the code. Encapsulate this properly and it will become much easier to use

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
```diff
@@ -93,12 +104,14 @@ public interface IDataBrowserPresenter : IPresenter<IDataBrowserView>, IPresente
 
    public class DataBrowserPresenter : PresenterWithColumnSettings<IDataBrowserView, IDataBrowserPresenter>, IDataBrowserPresenter
    {
+      public Cache<string, List<string>> OutputMappingNames { get; set; } = new Cache<string, List<string>>();
       private readonly Cache<DataColumn, DataColumnDTO> _dataColumnDTOCache = new Cache<DataColumn, DataColumnDTO>(x => x.DataColumn, x => null);
       private readonly List<DataColumn> _allDataColumns = new List<DataColumn>();
       private Func<DataColumn, PathElements> _displayQuantityPathDefinition;
+      private bool _linkMappedOutputs;
```
msevestre, line 111: not obbvioius that this is a boolean. ALso why do you need this state? Can't you derive it from the UI?
How is this done for Used?

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
```diff
@@ -155,6 +168,20 @@ public void UsedChangedFor(DataColumnDTO dataColumnDTO, bool used)
...
+
+      private void updateLinkedObservedData(DataColumn dataColumn, bool used)
+      {
+         if (!_linkMappedOutputs) return;
+
+         var updatedDataColumnName = dataColumn.PathAsString;
+         if (!OutputMappingNames.Contains(updatedDataColumnName)) return;
+
+         var linkedObservedDataNames = OutputMappingNames[updatedDataColumnName];
+         var linkedObservedData = _dataColumnDTOCache.KeyValues.Where(x => linkedObservedDataNames.Contains(x.Key.Repository.Name)).Select(x => x.Value).ToList();
```
msevestre, line 182: This is what I mean here. Create a method for this. getObservedDataLinkedToOuptut or something like this.

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationAnalysisChartPresenter.cs
```diff
@@ -54,11 +53,27 @@ public override void UpdateAnalysisBasedOn(IAnalysable analysable)
...
 
          Refresh();
       }
 
+      private void setOutputMappingNames()
+      {
+         var outputMappingNames = new Cache<string, List<string>>();
+         foreach (var outputMapping in _parameterIdentification.OutputMappings.All)
+         {
+            var outputName = outputMapping.FullOutputPath;
```
msevestre, line 67: All of this can be encapulated in a separate object. Dealing with Cache/Ditcionary as main API is, IMO, not the right level of API. Too low. Ok for internal stuff but not to be exosed

src/OSPSuite.UI/Views/Charts/ChartEditorView.cs
```diff
@@ -57,18 +60,51 @@ public ChartEditorView(IMenuBarItemToBarItemMapper barItemMapper, IImageListRetr
...
+         _barEditItemLinkSimulationObserved = new BarEditItem(_barManager)
+         {
+            Edit = repositoryItemCheckEditLinkSimulationAndObserved,
+            Alignment = BarItemLinkAlignment.Right,
+            AutoFillWidth = false,
+            Width = 20,
+            Caption = Captions.LinkDataToSimulations,
+            CaptionAlignment = HorzAlignment.Near,
+            PaintStyle = BarItemPaintStyle.Caption,
+            Hint = "test hint"
```
msevestre, line 72: Remove this and use a constant

src/OSPSuite.UI/Views/Charts/ChartEditorView.cs
```diff
@@ -57,18 +60,51 @@ public ChartEditorView(IMenuBarItemToBarItemMapper barItemMapper, IImageListRetr
 
          _barEditItemForUsedIn.SuperTip = new SuperToolTip().WithText(ToolTips.UseSelectedCurvesToolTip);
 
+         _barEditItemLinkSimulationObserved = new BarEditItem(_barManager)
+         {
+            Edit = repositoryItemCheckEditLinkSimulationAndObserved,
+            Alignment = BarItemLinkAlignment.Right,
+            AutoFillWidth = false,
+            Width = 20,
+            Caption = Captions.LinkDataToSimulations,
```
msevestre, line 69: This might be too long for a caption with a width of 20,
Also beware of fixed width... I know this is how the other one is done, but we know that this bites you later on with different screen resolution. This might work...do not know

src/OSPSuite.UI/Views/Charts/ChartEditorView.cs
georgeDaskalakis: this is actually unecessary, was just a test. removed.

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -170,6 +176,12 @@ public interface IChartEditorPresenter : IPresenter<IChartEditorView>,
       /// <param name="isUsed">true for checked, false for unchecked and null for indeterminate</param>
       void UpdateUsedForSelection(bool? isUsed);
 
+      /// <summary>
+      ///    Updates the LinkedOutputData menu item checkbox check state.
+      /// </summary>
+      /// <param name="linkSimToData">true for checked, false for unchecked</param>
+      void UpdateLinkSimulationToDataSelection(bool linkSimToData);
```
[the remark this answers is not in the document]
georgeDaskalakis: done

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
msevestre: @georgeDaskalakis It this clear?

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
georgeDaskalakis: renamed to _isLinkedMappedOutputs, as discussed yesterday it makes sense to keep it here

src/OSPSuite.UI/Views/Charts/ChartEditorView.cs
georgeDaskalakis: I have tried it with different resolutions and it seems fine. I have also tried not specifying a fixed width, but it seems to break the alignment to the right, adding a small space, so I would leave it as is.

georgeDaskalakis:
@msevestre I have ended up extending OutputMappings and then passing the whole object down to the DataBrowserPresenter, so no Cache anywhere anymore. This is ready for review again.

msevestre:
Sorry I was reviewing the wrong PR. For some reasons, I though this was the one you wanted me to review

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
```diff
@@ -172,6 +208,22 @@ public IReadOnlyList<DataColumn> GetAllUsedDataColumns()
          return _dataColumnDTOCache.KeyValues.Where(x => x.Value.Used).Select(x => x.Key).ToList();
       }
 
+      public OutputMappings AllOutputMappings { get; set; }
+
+      public void OutputObservedDataLinkingChanged(bool isLinkedSimToData)
+      {
+         _isLinkedMappedOutputs = isLinkedSimToData;
```
msevestre, line 215: call the variable the same as the parameter => ``isLinkedMappedOutputs`

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
```diff
@@ -155,6 +168,29 @@ public void UsedChangedFor(DataColumnDTO dataColumnDTO, bool used)
...
+         if (!_isLinkedMappedOutputs) return;
+         
+         var linkedObservedData = getLinkedObservedDataFromOutputPath(dataColumn.PathAsString);
+         SetUsedState(linkedObservedData, used);
+      }
+
+      private List<DataColumnDTO> getLinkedObservedDataFromOutputPath(string updatedDataColumnName)
+      {
+         var linkedObservedDataRepositories = AllOutputMappings.AllDataRepositoryMappedTo(updatedDataColumnName);
+         var linkedObservedData = getDataColumnDTOsFromDatarepositories(linkedObservedDataRepositories);
```
msevestre, line 185: return here

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
```diff
@@ -172,6 +208,22 @@ public IReadOnlyList<DataColumn> GetAllUsedDataColumns()
...
+      {
+         _isLinkedMappedOutputs = isLinkedSimToData;
+
+         if (!_isLinkedMappedOutputs) return;
+
+         foreach (var dataColumnPair in _dataColumnDTOCache.KeyValues)
+         {
+            var outputColumnUsed = dataColumnPair.Value.Used;
+            var linkedObservedData = getLinkedObservedDataFromOutputPath(dataColumnPair.Key.PathAsString);
+            SetUsedState(linkedObservedData.ToList(), outputColumnUsed);
```
msevestre, line 223: this is already a list

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -170,6 +172,12 @@ public interface IChartEditorPresenter : IPresenter<IChartEditorView>,
       /// <param name="isUsed">true for checked, false for unchecked and null for indeterminate</param>
       void UpdateUsedForSelection(bool? isUsed);
 
+      /// <summary>
+      ///    Updates the LinkedOutputData menu item checkbox check state.
+      /// </summary>
+      /// <param name="isLinkedSimToData">true for checked, false for unchecked</param>
+      void UpdateLinkSimulationToDataSelection(bool isLinkedSimToData);
```
msevestre: rename parameter (see below)

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
georgeDaskalakis: done

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
```diff
@@ -155,6 +168,29 @@ public void UsedChangedFor(DataColumnDTO dataColumnDTO, bool used)
...
+      {
+         if (!_isLinkedMappedOutputs) return;
+         
+         var linkedObservedData = getLinkedObservedDataFromOutputPath(dataColumn.PathAsString);
+         SetUsedState(linkedObservedData, used);
+      }
+
+      private List<DataColumnDTO> getLinkedObservedDataFromOutputPath(string updatedDataColumnName)
+      {
+         var linkedObservedDataRepositories = AllOutputMappings.AllDataRepositoryMappedTo(updatedDataColumnName);
```
msevestre, line 184: Method is called xxxFromOutputPath but the parameter is called `updatedDataColumnName`? Confusing

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
```diff
@@ -155,6 +168,29 @@ public void UsedChangedFor(DataColumnDTO dataColumnDTO, bool used)
...
+         var linkedObservedData = getDataColumnDTOsFromDatarepositories(linkedObservedDataRepositories);
+         return linkedObservedData;
+      }
+
+      private List<DataColumnDTO> getDataColumnDTOsFromDatarepositories(IEnumerable<DataRepository> linkedObservedDataRepositories)
```
msevestre, line 189: getDataColumnDTOsFromDatarepositories => getDataColumnDTOsFromDataRepositories

(use resharper for this)
image

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
```diff
@@ -155,6 +168,29 @@ public void UsedChangedFor(DataColumnDTO dataColumnDTO, bool used)
...
+      {
+         var linkedObservedData = _dataColumnDTOCache.KeyValues.Where(x => linkedObservedDataRepositories.Any(y => y.Id == x.Key.Repository.Id))
```
msevestre, line 191: Make the code easier to read. Where, Any, Select
This can be simplified by not using the primitive  such as Id but working on lists directly
Make the parameter IReadOnly
IReadOnlyList<DataRepository> linkedObservedDataRepositories

and then 
return _dataColumnDTOCache.Where(x => linkedObservedDataRepositories.Contains(x.DataColumn.Repository)).ToList();

you can also create a Repistory method in the DataColumn DTO Repository =>DataColum?.Repository
so it becomes
return _dataColumnDTOCache.Where(x => linkedObservedDataRepositories.Contains(x.Repository)).ToList();

In general, there is no reason to work with dictionary, cache, key and values at the presenter level. it is possible, but often not required

msevestre (changes requested):

src/OSPSuite.Core/Domain/ParameterIdentifications/OutputMappings.cs
```diff
@@ -46,6 +46,11 @@ public virtual IEnumerable<OutputMapping> OutputMappingsUsingDataRepository(Data
          return All.Where(x => x.UsesObservedData(dataRepository));
       }
 
+      public virtual IEnumerable<DataRepository> AllDataRepositoryMappedTo(string outputFullPath)
+      {
+         return All.Where(x => x.FullOutputPath == outputFullPath).Select(x => x.WeightedObservedData.ObservedData);
```
msevestre, line 51: Parameter name not consistnet here `FullOutputPath ` vs `outputFullPath`
note: this will be the name of the OUTPUT with the simulation. Is this what you need? I would add a comment to the method for intellisense: 
fullOutputPath: Full path of the output including simulation name or sthg like this

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
```diff
@@ -172,6 +208,22 @@ public IReadOnlyList<DataColumn> GetAllUsedDataColumns()
...
+      public void OutputObservedDataLinkingChanged(bool isLinkedMappedOutputs)
+      {
+         _isLinkedMappedOutputs = isLinkedMappedOutputs;
+
+         if (!_isLinkedMappedOutputs) return;
+
+         foreach (var dataColumnPair in _dataColumnDTOCache.KeyValues)
+         {
+            var outputColumnUsed = dataColumnPair.Value.Used;
+            var linkedObservedData = getLinkedObservedDataFromOutputPath(dataColumnPair.Key.PathAsString);
```
msevestre, line 222: This will also iterate over observed data column no? 
You can iterate over the _dataColumnDTOCache and access the Path or the datacolumn directly. No need to use the internal of the cache here

msevestre: in fact getLinkedObservedDataFromOutputPath could simply given a DataColumn as Parameter the way it's being used

msevestre (changes requested):
Looks good to me. A few suggestions to go away from using the KEY in the code. I am absolutely in favor of not knowing how the underlying structure is being cached. In this case, the code is just more complicated and not more efficient because you are NOT accessing properties by keys. Rather, you are iterating. So let's simplify the code

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
georgeDaskalakis: done

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
georgeDaskalakis: .ToList() removed

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
georgeDaskalakis: done

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
georgeDaskalakis: renamed to outputPath

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
georgeDaskalakis: typo fixed

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
georgeDaskalakis: used the IReadOnlyList  and direct return. I would say creating a new Repository method in DataColumnDTO is an overkill: to me it does not simplify the code that much

src/OSPSuite.Core/Domain/ParameterIdentifications/OutputMappings.cs
georgeDaskalakis: done

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
georgeDaskalakis: Hmm not sure how this would work... we need the "Used" value from DataColumnDTO and the "PathAsString" from the corresponding DataColumn...

georgeDaskalakis: done

georgeDaskalakis:
@msevestre discussed improvements implemented - this should be ready now

PR #1650, opened 2022-06-27
GetValueByPath from ContainerTask

### LINKED ISSUE #1645: `OSPSuite.R.Services.ContainerTask`: implement `getValueByPath`
It is possible to set values of quantities using `SetValueByPath` method. However, to get the value (e.g. when creating a simulation batch and setting start values), we still need an object of the quantity in R. A method that receives the simulation and the path to the quantity and returns it's current value would help to avoid (expensive) creation of objects.

@msevestre would be great for the parameter identification package.

abdelr:
Fixes #1645 (GetValueByPath from ContainerTask)

### DIFF
--- a/src/OSPSuite.R/Services/ContainerTask.cs
+++ b/src/OSPSuite.R/Services/ContainerTask.cs
@@ -179,40 +179,20 @@ namespace OSPSuite.R.Services
 
       public void SetValueByPath(IModelCoreSimulation simulation, string path, double value, bool throwIfNotFound)
       {
-         if (path.Contains(WILD_CARD))
-            throw new OSPSuiteException(Error.CannotSetValueByPathUsingWildCard(path));
-
-         var pathArray = path.ToPathArray();
-         var quantity = simulation.Model.Root.EntityAt<IQuantity>(pathArray);
-         if (quantity != null)
-         {
-            quantity.Value = value;
+         var quantity = singleQuantityByPath(simulation, path, throwIfNotFound);
+         if (quantity == null)
             return;
-         }
 
-         if (throwIfNotFound)
-            throw new OSPSuiteException(Error.CouldNotFindQuantityWithPath(path));
-
-         _logger.AddWarning(Error.CouldNotFindQuantityWithPath(path));
+         quantity.Value = value;
       }
 
       public double GetValueByPath(IModelCoreSimulation simulation, string path, bool throwIfNotFound)
       {
-         if (path.Contains(WILD_CARD))
-            throw new OSPSuiteException(Error.CannotGetValueByPathUsingWildCard(path));
+         var quantity = singleQuantityByPath(simulation, path, throwIfNotFound);
+         if (quantity == null)
+            return double.NaN;
 
-         var pathArray = path.ToPathArray();
-         var quantity = simulation.Model.Root.EntityAt<IQuantity>(pathArray);
-         if (quantity != null)
-         {
-            return quantity.Value;
-         }
-
-         if (throwIfNotFound)
-            throw new OSPSuiteException(Error.CouldNotFindQuantityWithPath(path));
-
-         _logger.AddWarning(Error.CouldNotFindQuantityWithPath(path));
-         return double.NaN;
+         return quantity.Value;
       }
 
       private IQuantity singleQuantityByPath(IModelCoreSimulation simulation, string path, bool throwIfNotFound)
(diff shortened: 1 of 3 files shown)


abdelr:
@PavelBal Waiting for review but already implemented

src/OSPSuite.R/Services/ContainerTask.cs
```diff
@@ -186,6 +196,25 @@ public void SetValueByPath(IModelCoreSimulation simulation, string path, double
          _logger.AddWarning(Error.CouldNotFindQuantityWithPath(path));
       }
 
+      public double GetValueByPath(IModelCoreSimulation simulation, string path, bool throwIfNotFound)
+      {
+         if (path.Contains(WILD_CARD))
+            throw new OSPSuiteException(Error.CannotGetValueByPathUsingWildCard(path));
+
+         var pathArray = path.ToPathArray();
+         var quantity = simulation.Model.Root.EntityAt<IQuantity>(pathArray);
```
Yuri05, line 205: you can pass `path` directly, `EntityAt ` can handle it and will split it in case it contains more than one path entry
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/1ce5c0b852f897b6c5b9aeae19ca3ad9abf9fb7e/src/OSPSuite.Core/Domain/ContainerExtensions.cs#L127-L131

msevestre (changes requested):

src/OSPSuite.R/Services/ContainerTask.cs
msevestre: and I would look at the SetValueByPath and so the same. Use the method instead of having the same code, three time in the same file

msevestre (changes requested):

Yuri05 (approved):

src/OSPSuite.R/Services/ContainerTask.cs
```diff
@@ -169,30 +179,28 @@ public IParameter[] AllParametersForSensitivityAnalysisMatching(ISimulation simu
...
-         if (throwIfNotFound)
-            throw new OSPSuiteException(Error.CouldNotFindQuantityWithPath(path));
+         quantity.Value = value;
+      }
 
-         _logger.AddWarning(Error.CouldNotFindQuantityWithPath(path));
+      public double GetValueByPath(IModelCoreSimulation simulation, string path, bool throwIfNotFound)
+      {
+         var quantity = singleQuantityByPath(simulation, path, throwIfNotFound);
+         if (quantity == null)
```
msevestre, line 192: there is the handy new (ish) syntax
return `quantity?.Value ?? double.NaN`

Don't have to be used but I wanted to point this out just in case

msevestre (changes requested):

msevestre:
build is failing. Can you check this out?

src/OSPSuite.R/Services/ContainerTask.cs
```diff
@@ -168,11 +168,11 @@ public IParameter[] AllParametersForSensitivityAnalysisMatching(ISimulation simu
...
+      public string BaseUnitNameByPath(IModelCoreSimulation simulation, string path, bool throwIfNotFound) => singleQuantityByPath(simulation, path, throwIfNotFound, Error.CannotGetValueByPathUsingWildCard(path)).BaseUnitName();
 
-      public string DimensionNameByPath(IModelCoreSimulation simulation, string path, bool throwIfNotFound) => singleQuantityByPath(simulation, path, throwIfNotFound).DimensionName();
+      public string DimensionNameByPath(IModelCoreSimulation simulation, string path, bool throwIfNotFound) => singleQuantityByPath(simulation, path, throwIfNotFound, Error.CannotGetValueByPathUsingWildCard(path)).DimensionName();
 
-      public bool IsExplicitFormulaByPath(IModelCoreSimulation simulation, string path, bool throwIfNotFound) => singleQuantityByPath(simulation, path, throwIfNotFound)?.Formula.IsExplicit() ?? false;
+      public bool IsExplicitFormulaByPath(IModelCoreSimulation simulation, string path, bool throwIfNotFound) => singleQuantityByPath(simulation, path, throwIfNotFound, Error.CannotGetValueByPathUsingWildCard(path))?.Formula.IsExplicit() ?? false;
```
msevestre, line 175: I do not like this passing this stuff everywhere.
Do we need to differentiate between Get and set? What about we just change the message to get/set instead ?

abdelr: @msevestre Fair enough... I have unified the message

PR #1663, opened 2022-07-15
1647 adding output mappings to simulations

### LINKED ISSUE #1647: Mapping of simulation outputs to observed data in simulations
Individual and population simulations should get a new tab “Data” next to the tabs “Parameters” and “Reaction Diagram” (PK-Sim only). The tab cannot be closed.

image

The tab lists all observed data that has been added to any of the Time Profile plots of the respective simulation (in PK-Sim, multiple Time Profile plots can be created per simulation, in MoBi only one Time Profile). The data are listed in similar way as in the “Data” tab of the parameter identification, with some differences:

The column “Observed Data” is the first column, the column “Outputs” is the second
“Observed Data” is NOT a drop-down menu
Preview of observed data is not shown

image

For each observed data, a simulation output can be selected. “None” can also be selected. When adding observed data, the software should try to map the respective simulation output based on the meta data of the observed data (the same way it is done for parameter identification)

Removing the data from the “Data” tab removes it from all plots of the respective simulation where it is used.

georgeDaskalakis:
As discussed, the OutputMappings still need to be moved out of the ParameterIdentification folder, but I will do this in a separate commit, in order not to pollute the PR with too many file changes.

closes #1647


src/OSPSuite.Presentation/DTO/SimulationOutputMappingDTO.cs
```diff
@@ -0,0 +1,41 @@
...
+using OSPSuite.Assets;
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Data;
+using OSPSuite.Core.Domain.ParameterIdentifications;
+using OSPSuite.Presentation.DTO.ParameterIdentifications;
+using OSPSuite.Utility.Validation;
+
+namespace OSPSuite.Presentation.DTO
+{
+   public class SimulationOutputMappingDTO : OutputMappingDTO
```
msevestre, line 11: This looks fine to me. I do not think we need to do all of this just to add or remove one rule however
Can you create an issue to investigate removing SimulationOutputMappingDTO (explaining why it was created) and assign it to me?

msevestre (approved):
Looks good. One issue to create and assign to me please.
Merging

src/OSPSuite.Presentation/DTO/SimulationOutputMappingDTO.cs
georgeDaskalakis: @msevestre issue created:
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1670

PR #1680, opened 2022-08-06
Fixes #1679 move helper

### LINKED ISSUE #1679: Move helper methods from PKSim for GridLayout and TablePanel to core
and check OptimalHeight method from a few base classes where the offset needs to be added

msevestre:


src/OSPSuite.UI/Controls/UxGridView.cs
```diff
@@ -306,14 +308,15 @@ public int OptimalHeight
...
+               int height = viewInfo.CalcRealViewHeight(new Rectangle(0, 0, MaxValue, MaxValue));
                viewInfo.CalcRealViewHeight(oldBounds);
 
                var scrollInfo = (ScrollInfo) fi.GetValue(this);
                if (scrollInfo.HScrollVisible)
                   height += scrollInfo.HScrollSize;
 
-               return height;
+               //We had this extra height for each row as some is always missing on larger scaling.
+               return height + RowCount * COMPUTED_EXTRA_HEIGHT;
```
msevestre, line 319: This was seems to fix it for me. On larger scaling, I need to add some extra pixel per height. This works great in PKSim so I am putting it here. Let see

PR #1686, opened 2022-08-24
New simulation charts

georgeDaskalakis:


src/OSPSuite.Core/Domain/ISimulation.cs
```diff
@@ -9,5 +10,7 @@ public interface ISimulation : ILazyLoadable, IAnalysable, IUsesObservedData, IM
       IEnumerable<CurveChart> Charts { get; }
 
       OutputMappings OutputMappings { get; set; }
+
+      DataRepository ResultsDataRepository { get; set; }
```
msevestre, line 14: Hum...this will require some explanation here. Why do we suddenly need some results in the ISimulation interface
This will create problems with implementation that have results already

georgeDaskalakis: I agree a comment needs to be added - about the implementation, this actually already exists in MoBi Simulation, I have just added it to PkSimSimulation

msevestre: As I said, we already have it in PKSim simulation

My question is why do we need this. You can imagine that if the results were not defined in the ISimulation interface, there was probably a reason. Maybe because a Popsimulation also implements the ISim and it does not make sense? I don't have access to my laptop but I think this needs to be explained (e.g. To me;-)) because I doubt that this is needed

georgeDaskalakis: I will check that there are no conflicts in PK-Sim and see if it is possible to avoid having it at all. ( btw I am also in vacation, so I will probably not be able to do any changes until the 31st of August )

src/OSPSuite.Core/Chart/Simulations/SimulationPredictedVsObservedChart.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace OSPSuite.Core.Chart.Simulations
+{
+   public class SimulationPredictedVsObservedChart : PredictedVsObservedChart
+   {
+      public SimulationPredictedVsObservedChart()
+      {
+         ChartSettings.LegendPosition = LegendPositions.BottomInside;
```
msevestre, line 7: @georgeDaskalakis Shouldn't that be the default for PI as well?

src/OSPSuite.Core/Domain/ISimulation.cs
msevestre, line 14: Ok let's keep this. Let's make sure we explain that it can be null for some simulations
In fact, I feel that for the population, we should throw an exception saying that this should never be called 
@Yuri05 Thoughts?

src/OSPSuite.Core/Domain/Services/ParameterIdentifications/IResidualCalculator.cs
```diff
@@ -7,5 +8,6 @@ public interface IResidualCalculator
    {
       void Initialize(RemoveLLOQMode removeLLOQMode);
       ResidualsResult Calculate(IReadOnlyList<SimulationRunResults> simulationsResults, IReadOnlyList<OutputMapping> allOutputMappings);
+      ResidualsResult CalculateForSimulation(DataRepository simulationResultRepository, IReadOnlyList<OutputMapping> allOutputMappings);
```
msevestre, line 11: Probably this class should be moved to another namesepase?
Also call the method Calculate (overload?)

src/OSPSuite.Core/Domain/Services/ParameterIdentifications/ResidualCalculator.cs
```diff
@@ -34,14 +33,23 @@ public ResidualsResult Calculate(IReadOnlyList<SimulationRunResults> simulations
...
+      }
+
+      private ResidualsResult calculateResidualsResult(IReadOnlyList<OutputMapping> allOutputMappings,
+         Cache<string, DataColumn> simulationColumnsCache)
+      {
          var residualResult = new ResidualsResult();
 
          foreach (var outputMapping in allOutputMappings)
          {
+            if (outputMapping.Output == null)
```
msevestre, line 50: How can this be null?
Also, please use allOutputMappings.where(x=>x.Output!=null) in the foreach

src/OSPSuite.Core/Domain/Services/ParameterIdentifications/ResidualCalculator.cs
```diff
@@ -122,5 +130,13 @@ protected double LogSafe(float x)
 
          return Math.Log10(x);
       }
+
+      public ResidualsResult CalculateForSimulation(DataRepository simulationResultRepository, IReadOnlyList<OutputMapping> allOutputMappings)
```
msevestre, line 134: Move this method up right below the other one. Public method first

src/OSPSuite.Presentation/Presenters/Charts/CommonAnalysisChartPresenter.cs
```diff
@@ -0,0 +1,130 @@
...
+using OSPSuite.Core.Extensions;
+using OSPSuite.Presentation.Extensions;
+using OSPSuite.Presentation.Services.Charts;
+using OSPSuite.Presentation.Views;
+using OSPSuite.Utility.Collections;
+using OSPSuite.Utility.Extensions;
+
+namespace OSPSuite.Presentation.Presenters.Charts
+{
+   public interface ICommonAnalysisChartPresenter : ISimulationAnalysisPresenter
```
msevestre, line 18: Why do you need an empty interface?

src/OSPSuite.Presentation/Presenters/Charts/CommonAnalysisChartPresenter.cs
```diff
@@ -0,0 +1,130 @@
...
+   {
+   }
+
+   public abstract class CommonAnalysisChartPresenter<TChart, TView, TPresenter> : SimulationAnalysisChartPresenter<TChart, TView, TPresenter>,
```
msevestre, line 22: why does common means? What is the difference with SimulationAnalysisChartPresenter
Name is so what we know what is implemented

src/OSPSuite.Presentation/Presenters/PresenterConstants.cs
```diff
@@ -13,6 +13,7 @@ public static class PresenterKeys
          public static readonly string ParameterIdentificationResidualVsTimeChartPresenter = createConstant("ParameterIdentificationResidualVsTimeChartPresenter");
          public static readonly string ParameterIdentificationResidualHistogramPresenter = createConstant("ParameterIdentificationResidualHistogramPresenter");
          public static readonly string ParameterIdentificationPredictedVsActualChartPresenter = createConstant("ParameterIdentificationPredictedVsActualChartPresenter");
+         public static readonly string SimulationPredictedVsActualChartPresenter = createConstant("SimulationPredictedVsActualChartPresenter");
```
msevestre, line 16: let's call this SimulationPredictedVsObservedChartPresenter?

what about htis one? `ParameterIdentificationPredictedVsActualChartPresenter ` is this being used anywhere?
We should probably rename this as well

src/OSPSuite.Presentation/Presenters/SimulationResidualVsTimeChartPresenter.cs
```diff
@@ -0,0 +1,184 @@
...
+         dataRepository.FirstDataColumn().Values = new[] { 0f, 0f };
+         return dataRepository;
+      }
+
+      private float minObservedDataTime()
+      {
+         return getAllAvailableObservedData().Select(x => x.BaseGrid.Values.First()).Min();
+      }
+
+      private float MaxObservedDataTime()
```
msevestre, line 122: maxObservedDataTime?

src/OSPSuite.Presentation/Presenters/SimulationRunAnalysisPresenter.cs
```diff
@@ -0,0 +1,94 @@
...
+         else
+            UpdateCacheColor();
+
+         if (!_simulation.ResultsDataRepository.IsNull())
+         {
+            UpdateAnalysis();
+            ChartEditorPresenter.SetOutputMappings(_simulation.OutputMappings);
+         }
+
+         updateAnalysis();
```
msevestre, line 54: This is confusiing. There us UpdateAnalysis and updateAnalysis

src/OSPSuite.R/Domain/Simulation.cs
```diff
@@ -21,6 +21,12 @@ public class Simulation : ISimulation
       public bool UsesObservedData(DataRepository observedData) => false;
       public IEnumerable<CurveChart> Charts { get; } = new List<CurveChart>();
       public OutputMappings OutputMappings { get; set; }
+      public DataRepository ResultsDataRepository { get; set; }
+
+      public DataRepository GetResultRepository()
```
msevestre, line 26: what is this?

src/OSPSuite.UI/Views/SimulationRunAnalysisView.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace OSPSuite.UI.Views
+{
+   public partial class SimulationRunAnalysisView : BaseUserControl, ISimulationRunAnalysisView
+   {
+      private ISimulationRunAnalysisPresenter _presenter;
+
+      public SimulationRunAnalysisView()
+      {
+         InitializeComponent();
+         layoutControlItem2.TextVisible = true;
```
msevestre, line 16: Rename those controls. Thsi is not ok to have the default n gaming in code like this

src/OSPSuite.UI/Views/SimulationRunAnalysisView.cs
```diff
@@ -0,0 +1,38 @@
...
+         layoutControlItem2.Visibility = LayoutVisibility.Never;
+         layoutControlItem2.Text = "Total error";
```
msevestre, line 18: use constants

src/OSPSuite.UI/Views/SimulationRunAnalysisView.cs
```diff
@@ -0,0 +1,38 @@
...
+
+      public void SetAnalysisView(IView view)
+      {
+         panelChart.FillWith(view);
+      }
+
+      public void SetTotalError(double totalError)
+      {
+         layoutControlItem2.Visibility = LayoutVisibility.Always;
+         totalErrorTextEdit.Text = totalError.ToString();
```
msevestre, line 30: Please never format double like this. We have methods everywhere for this.,

msevestre (changes requested):

src/OSPSuite.Core/Chart/Simulations/SimulationPredictedVsObservedChart.cs
georgeDaskalakis: true, moved this to  PredictedVsObservedChart

src/OSPSuite.Core/Domain/Services/ParameterIdentifications/IResidualCalculator.cs
georgeDaskalakis: Moving to another namespace done. About the renaming, do you think I should add a comment regarding where each overload is used (PI or Simulation), or better it is not necessary since it could actually theoretically be used also in other cases?

src/OSPSuite.UI/Views/SimulationRunAnalysisView.cs
georgeDaskalakis: done

msevestre: @georgeDaskalakis I would appreciate it if you could review your PR before submitting them to me for correctness so that I do not waste to much time on triviality. Can we try from now on that you give it a first pass, add comments etc.. so that I can navigate in a PR nicely and focus on what matters? Thanks :)

src/OSPSuite.R/Domain/Simulation.cs
georgeDaskalakis: removed

src/OSPSuite.Core/Domain/Services/ParameterIdentifications/ResidualCalculator.cs
georgeDaskalakis: done

src/OSPSuite.Core/Domain/Services/ParameterIdentifications/ResidualCalculator.cs
georgeDaskalakis: Simulation can have null Outputs, meaning that the corresponding Observed Data is not mapped to an Output. This is not a valid OutputMapping and should not be taken into account for the calculation

Corrected the foreach and also added the above explanation as a comment

src/OSPSuite.Presentation/Presenters/PresenterConstants.cs
georgeDaskalakis: Yes both are used, probably something outdated. Renamed.

src/OSPSuite.Presentation/Presenters/SimulationResidualVsTimeChartPresenter.cs
georgeDaskalakis: done

src/OSPSuite.Presentation/Presenters/Charts/CommonAnalysisChartPresenter.cs
georgeDaskalakis: removed

src/OSPSuite.Presentation/Presenters/Charts/CommonAnalysisChartPresenter.cs
georgeDaskalakis: Common here refers to both PI and Simulation. But I agree this is not optimal: 

The thing is there exists a SimulationAnalysisChartPresenter that is also implemented by ParameterIdentificationAnalysisChartPresenter. I would actually suggest to merge these two presenters into one, then we could call it AnalysisChartPresenter?

src/OSPSuite.Core/Domain/ISimulation.cs
georgeDaskalakis: For Core I have added a comment in the code, then in the solutions we can actually throw the esception.

src/OSPSuite.UI/Views/SimulationRunAnalysisView.cs
georgeDaskalakis: @msevestre sorry for this you are right, I will be more careful with things like that in the future

src/OSPSuite.Presentation/Presenters/SimulationRunAnalysisPresenter.cs
georgeDaskalakis: Renamed the private function to "updateRunResults" since this is what it sspecifically does.

src/OSPSuite.Core/Domain/Services/ParameterIdentifications/ResidualCalculator.cs
msevestre: I don't understand. Ho w can a simulation mapping be valid if there is no output? IMO this is wrong. Please explain

georgeDaskalakis: It is not valid, that's why we are skipping it. In the Output Mapping presenter for the simulation we have all the observed data listed and an output can be mapped or not. The outputs that are not mapped ( == null  ) are not being used - here is where they are actually filtered out. Alternatively it could also be done at some earlier point in code.

georgeDaskalakis: (though not sure which point that could be)

msevestre: >In the Output Mapping presenter for the simulation we have all the observed data listed and an output can be mapped or not

This is a presentation thing. But why are objects created for invalid mapping? Do you want to have a call?

georgeDaskalakis: @msevestre I see your point. Give me a few minutes to go over the code

src/OSPSuite.Core/Domain/Services/ResidualCalculator.cs
```diff
@@ -34,13 +34,28 @@ public ResidualsResult Calculate(IReadOnlyList<SimulationRunResults> simulations
...
 
-         foreach (var outputMapping in allOutputMappings)
+      public ResidualsResult Calculate(DataRepository simulationResultRepository, IReadOnlyList<OutputMapping> allOutputMappings)
+      {
+         var simulationColumnsCache = new Cache<string, DataColumn>(x => x.PathAsString, x => null);
+         simulationColumnsCache.AddRange(simulationResultRepository.AllButBaseGrid());
+
+         return calculateResidualsResult(allOutputMappings, simulationColumnsCache);
+      }
+      private ResidualsResult calculateResidualsResult(IReadOnlyList<OutputMapping> allOutputMappings,
```
msevestre, line 51: @georgeDaskalakis Please run resharper in this file. It looks like its completely skewed

src/OSPSuite.Presentation/Presenters/Charts/CommonAnalysisChartPresenter.cs
msevestre: I am a bit confused now. I still don't understand what this Common presenter is about. What presenter implement this? What is the difference with SimulationAnalysisChartPresenter?

msevestre (commented):
I am still not convinced with the naming. Are you available for a call?

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
```diff
@@ -159,9 +162,16 @@ public void InitializeSimulation(ISimulation simulation)
          _simulation = simulation;
       }
```
georgeDaskalakis, line 164: @msevestre I implemented like this with onValueChanged ( from the view).

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
```diff
@@ -159,9 +162,16 @@ public void InitializeSimulation(ISimulation simulation)
...
 
+      public void UpdateSimulationOutputMappings(SimulationOutputMappingDTO simulationOutputMappingDTO)
+      {
+         if(!_simulation.OutputMappings.OutputMappingsUsingDataRepository(simulationOutputMappingDTO.ObservedData).Any())
+            _simulation.OutputMappings.Add(simulationOutputMappingDTO.Mapping);
+      }
+
```
georgeDaskalakis, line 170: @msevestre I implemented like this with onValueChanged ( from the view).

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
```diff
@@ -102,7 +103,9 @@ private void updateOutputMappingList()
 
             var newOutputMappingDTO = mapFrom(newOutputMapping);
 
-            _simulation.OutputMappings.Add(newOutputMapping);
+            if (newOutputMapping.Output != null) 
+               _simulation.OutputMappings.Add(newOutputMapping);
```
msevestre, line 107: Let's review this code here and the one below. I don not understand what's going on. You are creating a cache of all output in a loop where the cache is independent form the variable being changed. We are almost there but this code is not efficient nad we should check this out one more time

src/OSPSuite.Presentation/Presenters/SimulationResidualVsTimeChartPresenter.cs
```diff
@@ -0,0 +1,179 @@
...
+         var markerRepository = createMarkerRepository();
+         AddDataRepositoriesToEditor(new[] { markerRepository });
+         AddCurvesFor(markerRepository, (column, curve) =>
+         {
+            curve.UpdateMarkerCurve(ZERO);
+            _markerCurveId = curve.Id;
+         });
+      }
+
+      private DataRepository createMarkerRepository()
```
msevestre, line 103: That really looks a lot like the other code. This does not have to be refactored in this PR but this needs to be refactored absolutely. This is an exact copy paste of the same logic. Create a service for all of those methods

msevestre (commented):
I Am good with all of this except the one implementation if the outputMapping stuff. Let's review together to understand the flow. I think it's almost ready.
Happy to deal with this in another PR if you prefer as long as you create an issue for this before merging this one. 
So 1-You create one issue and you merge
2- we fix this in this PR and then you merge :)

georgeDaskalakis:
> I Am good with all of this except the one implementation if the outputMapping stuff. Let's review together to understand the flow. I think it's almost ready. Happy to deal with this in another PR if you prefer as long as you create an issue for this before merging this one. So 1-You create one issue and you merge 2- we fix this in this PR and then you merge :)

I do not see any reason in creating a separate issue for this - let's fix this as a part of this PR.

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
georgeDaskalakis: @msevestre we can talk about this today in the meeting (or even later if you prefer)

msevestre:
@georgeDaskalakis Any news on this?

georgeDaskalakis:
> @georgeDaskalakis Any news on this?

@msevestre the changes we discussed (the service for zeroMarker in residualsVsTimeChart and handling the adding and removing OutputMappings to the simulation where the adding and removing of observed data happens) have been implemented here ( here meaning: the adding and removing of Observed Data happens in PK-Sim and MoBi -also implemented in different ways. Som code has been movewd there and will thus be reviewed in following PRs).

src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
```diff
@@ -88,39 +88,47 @@ private DataColumn createIdentityColumn(IDimension mergedDimension, params float
...
+         var yAxis = chart.AxisBy(AxisTypes.Y);
 
-         if (defaultDimension != null)
-            xAxis.Dimension = defaultDimension;
-
-         xAxis.Scaling = chart.AxisBy(AxisTypes.Y).Scaling;
-         xAxis.UnitName = chart.AxisBy(AxisTypes.Y).UnitName;
+         var defaultDimension = mostFrequentDimension(observationColumns.ToList());
 
+         setAxisDimension(defaultDimension, chart, xAxis);
```
msevestre, line 104: @georgeDaskalakis what are those changes about here? Is this to fix another issue?

src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
```diff
@@ -88,39 +88,47 @@ private DataColumn createIdentityColumn(IDimension mergedDimension, params float
...
+         xAxis.Caption += Captions.ParameterIdentification.ObservedChartAxis + defaultDimension;
+         yAxis.Caption += Captions.ParameterIdentification.SimulatedChartAxis + defaultDimension;
```
msevestre, line 106: at any rate, use string itnerpolation when concatenating string

src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
```diff
@@ -24,20 +23,21 @@ public interface IPredictedVsObservedChartService
...
+      void AddCurvesFor(IEnumerable<DataColumn> observationColumns, DataColumn calculationColumn, PredictedVsObservedChart chart,
+         Action<DataColumn, Curve> action = null);
 
       /// <summary>
       ///    Sets the <paramref name="chart" /> x axis dimension to the most occurring dimension from
-      ///    <paramref name="observationColumns" />
+      ///    <paramref name="observationColumns" /> and adjusts the axes names
       /// </summary>
-      void SetXAxisDimension(IEnumerable<DataColumn> observationColumns, ParameterIdentificationPredictedVsObservedChart chart);
+      void ConfigureAxesDimensionAndTitle(IEnumerable<DataColumn> observationColumns, PredictedVsObservedChart chart);
```
msevestre, line 40: Make this methods use IReadOnly instead of IEnumerable. You are always doing the cast in ToList before anyways it seems

src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
```diff
@@ -88,39 +88,47 @@ private DataColumn createIdentityColumn(IDimension mergedDimension, params float
...
-         var defaultDimension = mostFrequentDimension(dataColumns);
          var xAxis = chart.AxisBy(AxisTypes.X);
+         var yAxis = chart.AxisBy(AxisTypes.Y);
 
-         if (defaultDimension != null)
-            xAxis.Dimension = defaultDimension;
-
-         xAxis.Scaling = chart.AxisBy(AxisTypes.Y).Scaling;
-         xAxis.UnitName = chart.AxisBy(AxisTypes.Y).UnitName;
+         var defaultDimension = mostFrequentDimension(observationColumns.ToList());
```
msevestre, line 102: if you define IReadOnly at the top, no need to actually to a ToList here

src/OSPSuite.Core/Services/ResidualsVsTimeChartService.cs
```diff
@@ -0,0 +1,93 @@
...
+         var markerRepository = createMarkerRepository(minObservedDataTime, maxObservedDataTime, chart);
+         AddCurvesFor(markerRepository,(column, curve) =>
+         {
+            curve.UpdateMarkerCurve(ZERO);
+            _markerCurveId = curve.Id;
+         }, chart);
+
+         return markerRepository;
+      }
+      private DataRepository createMarkerRepository(float minObservedDataTime, float maxObservedDataTime, AnalysisChartWithLocalRepositories chart)
```
msevestre, line 56: Run resharper in this file. It seems to be a bit off

src/OSPSuite.Core/Services/ResidualsVsTimeChartService.cs
```diff
@@ -0,0 +1,93 @@
...
+      DataRepository AddZeroMarkerCurveToChart(AnalysisChartWithLocalRepositories chart, float minObservedDataTime, float maxObservedDataTime);
+
+      /// <summary>
+      ///    Creates the DataRepository for the given residuals <paramref name="outputResidual" /> with given
+      /// <paramref name="id" /> and <paramref name="repositoryName" /> 
+      /// </summary>
+      DataRepository CreateScatterDataRepository(string id, string repositoryName, OutputResiduals outputResidual);
+   }
+
+   public class ResidualsVsTimeChartService : IResidualsVsTimeChartService
```
msevestre, line 32: Let's add a test for this service. I feel that this can be completely tested

msevestre (commented):
Looking very good. I think we are very clean now and removed all duplication also. Just a few comments and then we can merge. 
I am just really missing a test file for the new service. This is now so nicely encapsulated...

src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
georgeDaskalakis: @msevestre yes this is  a bugfix here

src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
georgeDaskalakis: Concatenation was wrong anyway. Removed

src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
georgeDaskalakis: done

src/OSPSuite.Core/Services/ResidualsVsTimeChartService.cs
georgeDaskalakis: done

src/OSPSuite.Core/Services/ResidualsVsTimeChartService.cs
georgeDaskalakis: @msevestre working on this. Also there are a bunch of broken tests that I had not noticed ( no build on the PR, since it gets merged into a feature branch )

georgeDaskalakis: @msevestre the tests for the service have been added. This should be ready for merge.

msevestre (commented):
@georgeDaskalakis I am going to merge. I do not see where the OutputMapping is added in the simulation when an observed data is added to a simulation. I think this is missing?

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
```diff
@@ -159,9 +143,16 @@ public void InitializeSimulation(ISimulation simulation)
          _simulation = simulation;
       }
 
+      public void UpdateSimulationOutputMappings(SimulationOutputMappingDTO simulationOutputMappingDTO)
```
msevestre, line 146: Yeah this is not the right place. As I said, the presenter should not modify the object just because it is dislayign the object. I am going to create an issue for this

PR #1694, opened 2022-09-21
1691 initialize on batch create

rwmcintosh:
Cloning the simulation prevents changes to the original reference from affecting the batch once created

### DIFF
--- a/src/OSPSuite.Core/Domain/ModelCoreSimulation.cs
+++ b/src/OSPSuite.Core/Domain/ModelCoreSimulation.cs
@@ -11,7 +11,7 @@ namespace OSPSuite.Core.Domain
       /// <summary>
       ///    Build configuration used to create the simulation. May be null
       /// </summary>
-      IBuildConfiguration BuildConfiguration { get; }
+      IBuildConfiguration BuildConfiguration { get; set; }
 
       OutputSelections OutputSelections { get; }
 
--- a/src/OSPSuite.R/Domain/ConcurrentRunSimulationBatch.cs
+++ b/src/OSPSuite.R/Domain/ConcurrentRunSimulationBatch.cs
@@ -23,9 +23,9 @@ namespace OSPSuite.R.Domain
 
       internal int MissingBatchesCount => Math.Max(0, SimulationBatchRunValues.Count - _simulationBatches.Count);
 
-      public ConcurrentRunSimulationBatch(Simulation simulation, SimulationBatchOptions simulationBatchOptions)
+      public ConcurrentRunSimulationBatch(IModelCoreSimulation simulation, SimulationBatchOptions simulationBatchOptions)
       {
-         Simulation = Api.GetSimulationTask().Clone(simulation.CoreSimulation);
+         Simulation = Api.GetSimulationTask().Clone(simulation);
          SimulationBatchOptions = simulationBatchOptions;
          Id = generateId();
       }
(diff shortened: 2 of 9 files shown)


src/OSPSuite.Core/Domain/ModelCoreSimulation.cs
```diff
@@ -11,7 +11,7 @@ public interface IModelCoreSimulation : IObjectBase, IWithCreationMetaData, IWit
       /// <summary>
       ///    Build configuration used to create the simulation. May be null
       /// </summary>
-      IBuildConfiguration BuildConfiguration { get; }
+      IBuildConfiguration BuildConfiguration { get; set; }
```
msevestre, line 14: This may be a breaking change in pksim and Mobi FYI

src/OSPSuite.R/Domain/ConcurrentRunSimulationBatch.cs
```diff
@@ -23,9 +23,9 @@ public class ConcurrentRunSimulationBatch : IDisposable, IWithId
 
       internal int MissingBatchesCount => Math.Max(0, SimulationBatchRunValues.Count - _simulationBatches.Count);
 
-      public ConcurrentRunSimulationBatch(IModelCoreSimulation simulation, SimulationBatchOptions simulationBatchOptions)
+      public ConcurrentRunSimulationBatch(Simulation simulation, SimulationBatchOptions simulationBatchOptions)
```
msevestre, line 26: Why this change here?

msevestre (changes requested):
I am not sure that changing the public interface to simulation is necessary. Explain?

src/OSPSuite.R/Domain/ConcurrentRunSimulationBatch.cs
rwmcintosh: As far as I can tell, this is always called with R.Domain.Simulation, which isn't set up for cloning. In another spot, I need to access the property CoreSimulation to actually create a clone of it. So rather than cast, I think it's ok that we just use the object.

src/OSPSuite.Core/Domain/ModelCoreSimulation.cs
rwmcintosh: Meaning, we'll have to implement the setters once this is merged?

src/OSPSuite.R/Domain/ConcurrentRunSimulationBatch.cs
rwmcintosh: maybe I just need a default constructor - hold on

rwmcintosh: Yes, with a default constructor for R.Domain.Simulation the clone manager is happy.

msevestre: What about you don't create a clone then but just instantiate a new Simulation?? If there was not default constructor, there was probalbgy a reason

msevestre: so return new Simulation(_clonedModelCoreSimulation)

rwmcintosh: If this goes back to IModelCoreSimulation, then there's no time during the batch creation where I can get access to the underlying CoreSimulation without a cast. I need either the default constructor and clone, or cast to R.Domain.Simulation to get access to the underlying CoreSimulation so I can clone it and then instantiate an R.Domain.Simulation.

msevestre:
Merged. In order for the behavior to be available in R, we will need to update the dll

PR #1708, opened 2022-09-28
Deviation lines in plots

### LINKED ISSUE #1707: Deviation Lines for Predicted vs Observed Plots
It is  proposed to add deviation lines to PredictedVsObserved plots. The user will be able to introduce a new deviation line by right clicking on the plot and selecting “Add deviation line”. This will open a “New Deviation Line” Dialog window, where the user will be able to specify the fold factor of the deviation line (eg 2-fold, 3-fold and so on).

image

After confirming the fold input, corresponding deviation lines will be added to the plot.

image

The newly introduced deviation lines will appear in the “Curves and Axis Options” of the chart, like fe the “Dashes” currently do. They will be named after their fold value. There the user will be able to select or deselect the created deviation lines, making them visible or not in the chart.

image

georgeDaskalakis:
closes #1707


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -932,6 +934,16 @@ public static class ParameterIdentification
          public static readonly string TimeProfileConfidenceIntervalAnalysis = "Confidence Interval";
          public static readonly string TimeProfileVPCIntervalAnalysis = "Visual Predictive Check Interval";
 
+         public static string DeviationLineNameUpper(float foldValue)
+         {
```
msevestre, line 938: Please use string concatenation
```
public static string DeviationLineNameUpper(float foldValue)=> $"{foldValue}-fold deviation Upper";
```
Same with the other line

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1306,7 +1318,16 @@ public static class GroupRowFormat
             public static string Simulation = "Simulation";
             public static string Time = "Time";
             public static string Observation = "Observation";
-            public static string Undefined = "Undefined";
+            public static string IdentityCurve = "Identity";
+            public static string Undefined = "Undefined";
```
msevestre, line 1322: Let's not use undefined for this stuff. Did not we say we were going to create a special group for this?

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1306,7 +1318,16 @@ public static class GroupRowFormat
...
             public static string Observation = "Observation";
-            public static string Undefined = "Undefined";
+            public static string IdentityCurve = "Identity";
+            public static string Undefined = "Undefined"; 
+         }
+
+         public static class DeviationLines
+         {
+            public static string SpecifyFoldValue = "Specify deviation fold value";
+            public static string DeviationLineDescription = "Will create two deviation lines according to the given fold value which has to be >= 1.\n" +
```
msevestre, line 1328: hum. Can we not used + here again. Use the @ maybe or all in one line?

src/OSPSuite.Core/Domain/Data/DataInfo.cs
```diff
@@ -29,7 +29,12 @@ public enum ColumnOrigins
       /// <summary>
       ///    Column is based on calculation values but is used in a different context (for example ParameterIdentification plots)
       /// </summary>
-      CalculationAuxiliary
+      CalculationAuxiliary,
+
+      /// <summary>
+      ///    Column is an identity curve or deviation line to facilitate the viewing of a plot 
+      /// </summary>
+      IdentityCurve
```
msevestre, line 37: Should we create two entries or call them all DeviationLine (identity line is just a special DeviationLine)

src/OSPSuite.Core/Extensions/CurveExtensions.cs
```diff
@@ -37,5 +40,37 @@ public static void UpdateMarkerCurve(this Curve curve, string curveName)
...
+      {
+         curve.VisibleInLegend = false;
+         curve.LineStyle = LineStyles.Solid;
+         curve.Color = Color.Black;
+         curve.Name = curveName;
+         curve.LineThickness = 2;
+         curve.Name = curveName;
+      }
+
+      public static void UpdateDeviationCurve(this Curve curve, string curveName, int curveNumber)
```
msevestre, line 54: curveNumber? Can we create an enum for this (UPPER, MIDDLE, LOWER instead of 0 1 etc..)

src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
```diff
@@ -38,6 +38,12 @@ public interface IPredictedVsObservedChartService
       ///    <paramref name="observationColumns" /> and adjusts the axes names
       /// </summary>
       void ConfigureAxesDimensionAndTitle(IReadOnlyList<DataColumn> observationColumns, PredictedVsObservedChart chart);
+
+      /// <summary>
+      ///    Creates new deviation line with fold value <paramref name="foldValue" />
+      /// </summary>
+      IEnumerable<DataRepository> AddDeviationLine(float foldValue, List<DataColumn> observationColumns, PredictedVsObservedChart chart,
```
msevestre, line 45: not clear why you return an enumeration of DataREpository here. 
CAn you explain what this is?

src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
```diff
@@ -107,6 +114,77 @@ public void ConfigureAxesDimensionAndTitle(IReadOnlyList<DataColumn> observation
...
+      {
+         var deviationLineRepository= createDeviationRepository(foldValue, dataColumns, mergedDimension);
+         chart.AddCurvesFor(deviationLineRepository, x => x.Name, _dimensionFactory,
+            (column, curve) => curve.UpdateDeviationCurve(column.Name, numberOfDeviationLines));
+
+         return deviationLineRepository;
+      }
+
+      private DataRepository createDeviationRepository(float foldValue, IReadOnlyList<DataColumn> dataColumns,
+         IDimension mergedDimension)
```
msevestre, line 152: call this one dimension. DO we care that this is a mergedDimension?
Also why is this parameter on a new line? Is resharper formatting like this?

src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
```diff
@@ -107,6 +114,77 @@ public void ConfigureAxesDimensionAndTitle(IReadOnlyList<DataColumn> observation
...
+         var columnsWithMatchingDimension = dataColumns.Where(x => x.Dimension.IsEquivalentTo(mergedDimension)).ToList();
+
+         var columnsForIdentityRepository = columnsWithMatchingDimension.Where(column => columnNonZeroValues(column).Any()).ToList();
+         if (!columnsForIdentityRepository.Any())
+            return null;
+
+         var identityMinimum = getIdentityMinimum(columnsForIdentityRepository, mergedDimension);
+         var identityMaximum = getIdentityMaximum(columnsForIdentityRepository, mergedDimension);
+
+         var baseGridUpper = new BaseGrid(Captions.ParameterIdentification.Deviation, mergedDimension)
```
msevestre, line 163: this should be rename. IT's called baseGridUpper but it's use for all lines

src/OSPSuite.Presentation/Presenters/Charts/ChartDisplayPresenter.cs
```diff
@@ -240,6 +259,19 @@ public virtual void Clear()
 
       public void OnDragOver(IDragEvent dragEvent) => DragOver(this, dragEvent);
 
+      public void AddDeviationLines()
+      {
+         using (var deviationLinesPresenter = _applicationController.Start<IDeviationLinesPresenter>())
```
msevestre, line 264: nice and simple. Well done. Love it

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationPredictedVsObservedChartPresenter.cs
```diff
@@ -61,6 +64,7 @@ protected override void ClearChartAndDataRepositories()
       {
          base.ClearChartAndDataRepositories();
          _identityRepositories.Clear();
+         _deviationLineRepositories.Clear();
```
msevestre, line 67: do we nee dot have thow two repositories? Isn't identity just a specific deviation? I am not sure. Just asking

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationPredictedVsObservedChartPresenter.cs
```diff
@@ -83,6 +87,15 @@ private void plotAllCalculations(DataRepository simulationResult, Action<DataCol
          var calculationColumns = calculationColumnsToPlot(simulationResult);
          calculationColumns.Each(x => plotCalculationColumn(x, action));
       }
+      private void addDeviationLines(float foldValue)
```
msevestre, line 90: run resharper in this file please

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationPredictedVsObservedChartPresenter.cs
```diff
@@ -83,6 +87,15 @@ private void plotAllCalculations(DataRepository simulationResult, Action<DataCol
...
+      {
+         var observationColumns = _parameterIdentification.AllObservedData.Select(x => x.FirstDataColumn()).ToList();
+         _deviationLineRepositories.AddRange(_predictedVsObservedChartService.AddDeviationLine(foldValue, observationColumns, Chart, _deviationLineRepositories.Count/2));
```
msevestre, line 93: That i don't undertand at all. What's the count divided by 2 etc.
Stuff like this needs to be commented in code ad this is not obvious at all (at least to me)

src/OSPSuite.Presentation/Presenters/SimulationPredictedVsObservedChartPresenter.cs
```diff
@@ -112,6 +116,16 @@ private void plotCalculationColumn(DataColumn calculationColumn, Action<DataColu
          });
       }
 
+      private void addDeviationLines(float foldValue)
+      {
+         var observationColumns = getAllAvailableObservedData().Select(x => x.FirstDataColumn()).ToList();
+         _deviationLineRepositories.AddRange(_predictedVsObservedChartService.AddDeviationLine(foldValue, observationColumns, Chart, _deviationLineRepositories.Count));
```
msevestre, line 122: and here it's not devided by 2

Plus this code is so much the same as the other one,. I think we need to compose here

src/OSPSuite.UI/Views/Charts/DeviationLinesView.cs
```diff
@@ -0,0 +1,41 @@
...
+      {
+         _presenter = presenter;
+      }
+
+      public float GetFoldValue()
+      {
+         return foldValueTextEdit.EditValue.ConvertedTo<float>();
+      }
+
+      private void tablePanel_Paint(object sender, System.Windows.Forms.PaintEventArgs e)
```
msevestre, line 36: remove

src/OSPSuite.UI/Views/Charts/DeviationLinesView.cs
```diff
@@ -0,0 +1,41 @@
...
+   {
+      private IDeviationLinesPresenter _presenter;
+      public DeviationLinesView()
+      {
+         InitializeComponent();
+         foldValueInputControlItem.TextVisible = false;
+         foldValueTextLabelControl.Text = Captions.Chart.DeviationLines.SpecifyFoldValue.FormatForLabel();
+         deviationLineDescriptionLabelControl.AsDescription();
+         deviationLineDescriptionLabelControl.Text = Captions.Chart.DeviationLines.DeviationLineDescription.FormatForDescription();
+         //Regex: only numbers, with 2 points of decimal precision, and greater than 1
```
msevestre, line 20: regex..oh no please :)
Let's create a DTO with a number and a rule and the screenbinder. This is how we do it everywhere. Right now you are bypassing the whole databinding and implementing everything by hand. why?

msevestre (changes requested):
A few changes required
Some code duplication
DataBinding missing
No tests whatsoever (except one little test)}. We need more here as the logic seems to be quite advanced

src/OSPSuite.Assets/UIConstants.cs
georgeDaskalakis: done

src/OSPSuite.Assets/UIConstants.cs
georgeDaskalakis: The new special group is Identity. Undefined was there already - it is being shown as deleted and then added because Identity got added before it (for me it also makes sense to leave the Undefined for last, even in the constants).

src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
georgeDaskalakis: done

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationPredictedVsObservedChartPresenter.cs
georgeDaskalakis: done

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationPredictedVsObservedChartPresenter.cs
georgeDaskalakis: Should not be devided by two, it is wrong that it is there. removed.

src/OSPSuite.Presentation/Presenters/SimulationPredictedVsObservedChartPresenter.cs
georgeDaskalakis: About the code duplication. Both call the service to get the work done, the main difference I would say is getting the observation columns....

src/OSPSuite.UI/Views/Charts/DeviationLinesView.cs
georgeDaskalakis: removed, thanks was added from some double click on the designer and I never noticed I guess

src/OSPSuite.Assets/UIConstants.cs
georgeDaskalakis: I mean the main reason for me was readability here, since it is quite a long line. Still can easily let everything be in one line.

src/OSPSuite.Core/Extensions/CurveExtensions.cs
georgeDaskalakis: this is not upper, lower aso, this is literally the count of deviation lines in the chart (possibly CurveCount would be a better naming now that I think of it), since we are setting a different LineStyle to each new deviation line with a rotation of 3

src/OSPSuite.Core/Domain/Data/DataInfo.cs
georgeDaskalakis: we ll call them all deviation lines

src/OSPSuite.Core/Extensions/CurveExtensions.cs
georgeDaskalakis: renamed to curve count

src/OSPSuite.UI/Views/Charts/DeviationLinesView.cs
georgeDaskalakis: used a DTO instead

georgeDaskalakis:
@msevestre I would say all the points we have discussed are tackled: DTO, some refactoring to avoid too much code duplication between deviation lines and identity, serialization. This is ready for a final review.

PR #1709, opened 2022-09-28
Fixes #1672 Errors are not propagated on failed simulation run

### LINKED ISSUE #1672: No error is propagated when `SimulationBatch` run returns an error
See description in https://github.com/Open-Systems-Pharmacology/OSPSuite.SimModel/issues/146 - it seems that the error is thrown silently and is not propagated.

@abdelr mayber you can tell more where the error is actually thrown?

rwmcintosh:


src/OSPSuite.R/Domain/SimulationBatch.cs
```diff
@@ -148,12 +149,15 @@ public SimulationResults Run(SimulationBatchRunValues simulationBatchRunValues)
       {
          _simModelBatch.UpdateParameterValues(simulationBatchRunValues.Values);
          _simModelBatch.UpdateInitialValues(simulationBatchRunValues.MoleculeValues);
-         var simulationResults = _simModelBatch.RunSimulation();
+         var simulationRunResults = _simModelBatch.RunSimulation();
+
+         if (!simulationRunResults.Success)
+            throw new OSPSuiteException(simulationRunResults.Error);
```
rwmcintosh, line 155: Unsuccessful run must be signaled by exception

src/OSPSuite.Core/Domain/Services/SimModelBatch.cs
```diff
@@ -79,7 +80,10 @@ public SimulationRunResults RunSimulation()
          {
             updateSimulationValues();
             _simModelSimulation.RunSimulation();
-            return new SimulationRunResults(success: true, warnings: WarningsFrom(_simModelSimulation), results: getResults());
+            var hasResults = simulationHasResults(_simModelSimulation);
+
+            return new SimulationRunResults(success: hasResults, warnings: WarningsFrom(_simModelSimulation), results: getResults(),
+               error: hasResults ? null : Error.SimulationDidNotProduceResults);
```
msevestre, line 86: why don't you create a method (something around the lines of...)
```
createSimulationResults(string errorFromException = null) {
var hasResults = simulationHasResults(_simModelSimulation);
var warnings = WarningsFrom(_simModelSimulation);
var error = errorFromException ??  (hasResults ? null : Error.SimulationDidNotProduceResults)
if(error==null){
return SimulationRunResults(success: true , warnings, results: getResults())
}

//error
return SimulationRunResults(false, warnings, error)
}
```
and call that in both instances?

src/OSPSuite.R/Domain/SimulationBatch.cs
msevestre: yep

tests/OSPSuite.R.Tests/Services/ConcurrentSimulationRunnerSpecs.cs
```diff
@@ -87,6 +88,66 @@ public void should_complete_without_error()
...
+         _results = sut.RunConcurrently();
+      }
+
+      [Observation]
+      public void should_throw_exception_when_the_simulation_run_fails()
+      {
+         var id = _ids.First();
+         var simulationBatch = Api.GetSimulationBatchFactory().Create(_simulation, _concurrentRunSimulationBatch.SimulationBatchOptions);
+            
+         The.Action(() => simulationBatch.Run(_simulationBatchRunValues.FirstOrDefault(v => v.Id == id)))
```
msevestre, line 146: hum I don't get it. The action was called already in RunConcurrently
So i think this test should check that _results has en entry with success=false and an error

That tests here is a test for the _simulationBatch or sthg. No?

msevestre (changes requested):
I am wondering if we can have the implementation only once for the returning of the results
Also the tests is not what I would except. I feel you are testing something else (this test is valid  but for another sut)

src/OSPSuite.Core/Domain/Services/SimModelBatch.cs
```diff
@@ -93,6 +95,25 @@ public SimulationRunResults RunSimulation()
...
+         {
+            return new SimulationRunResults(success: true, warnings, results: getResults());
+         }
+
+         //error
+         return new SimulationRunResults(false, warnings, new DataRepository(), error);
+      }
+
+      private bool simulationHasResults(Simulation simModelSimulation)
+      {
```
msevestre, line 113: FYI: This is ok to use the short hand syntax (see below)


```
private bool simulationHasResults(Simulation simModelSimulation) => simModelSimulation.AllValues.Any();
```

tests/OSPSuite.R.Tests/Services/ConcurrentSimulationRunnerSpecs.cs
```diff
@@ -87,6 +88,64 @@ public void should_complete_without_error()
...
+         _concurrentRunSimulationBatch.AddSimulationBatchRunValues(_simulationBatchRunValues[0]);
+
+         _ids.AddRange(_concurrentRunSimulationBatch.SimulationBatchRunValues.Select(x => x.Id));
+         _results = sut.RunConcurrently();
+      }
+
+      [Observation]
+      public void should_return_a_result_with_success_set_to_false()
+      {
+         _results.All(x => x.Succeeded).ShouldBeFalse();
```
msevestre, line 145: Do we have a test that now verifies that this flag is sometimes true?  (e.g. can you set the parameter for one run instead as for all). Then one run should be success and the one where it fails should be false.

rwmcintosh: We do now.

PR #1710, opened 2022-09-28
Fixes #1633 comparison threshold for fractions

### LINKED ISSUE #1633: Comparison Threshold for Fractions
At the moment we get sometimes very low comparison thresholds for **fractions** from SimModel.
E.g. in some of our Batch Comparison projects we have tolerance 1E-17 for `Fraction absorbed into mucosa segment`.

In principle it should not be the problem. The automatically calculated threshold of the ODE variables are usually OK, and then we calculate the threshold of the `Fraction absorbed ` as `Threshold_of_ODE_variable / Total_Applied_oral_drug_mass`.

However in this particular case applied drug mass **changes during the simulation in case of multiple administrations**.
And then the threshold (calculated using the total drug mass at the end of the simulation) might be too high at the beginning of the simulation.
In order to calculate the threshold properly here, we must either make it time dependent (too complicated) or e.g. use just the **first** applied dose (still too complicated).

Instead, I would suggest the following:
* define new constant (MIN_FRACTION_COMPARISON_THRESHOLD or so) for the lowest value of comparison threshold for all fraction quantities. 
  * I would set it to 1E-4 ➡️ we ignore the differences if both fractions are below 0.01%
* when setting the threshold for a quantity: check if the quantity has dimension "Fraction". 
  * If yes: `set quantity.threshold = Max(calculated threshold; MIN_FRACTION_COMPARISON_THRESHOLD)`

 Either here
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/f741455db467c4449ddb28106adb5c51c4aaf2f7/src/OSPSuite.Core/Batch/Mappers/SimulationResultsToBatchSimulationExportMapper.cs#L56

or here 
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/f741455db467c4449ddb28106adb5c51c4aaf2f7/src/OSPSuite.Core/Domain/DataFactory.cs#L76

msevestre:


src/OSPSuite.Core/Domain/DataFactory.cs
```diff
@@ -97,11 +98,22 @@ private DataColumn createColumn(BaseGrid baseGrid, IQuantity quantity, VariableV
...
+         return new BaseGrid(TIME, _dimensionFactory.Dimension(Constants.Dimension.TIME))
          {
-            QuantityInfo = new QuantityInfo(new List<string> {Constants.TIME}, QuantityType.BaseGrid),
+            QuantityInfo = new QuantityInfo(new List<string> {TIME}, QuantityType.BaseGrid),
             Values = new List<float>(values.ToFloatArray())
          };
       }
+
+      //Exposing this one publicly so that we can test it (hard to test with SimModel being involved in the mix)
+      public float CalculateComparisonThreshold(IQuantity quantity, double defaultThreshold)
```
msevestre, line 109: @rwmcintosh @Yuri05 
We cannot mock SImModel stuff I think so I am exposting this one publically so that I can test without having to instantiate a simmodel that would create a threshold that is below this value

rwmcintosh (approved):

PR #1712, opened 2022-09-29
a sometimes failing concurrency test

rwmcintosh:
@Yuri05 @msevestre @PavelBal 

When I run this test it will fail similarly often to the R code that inspired it.

The SequenceEqual test is checking that all the sequences are identical besides any from failing runs.

### DIFF
--- a/src/OSPSuite.R/Services/SimulationTask.cs
+++ b/src/OSPSuite.R/Services/SimulationTask.cs
@@ -7,10 +7,10 @@ namespace OSPSuite.R.Services
    public interface ISimulationTask
    {
       /// <summary>
-      /// Return a clone of the simulation that can be used during batch run.
-      /// This simulation will contain the minimal information required to run. Configuration settings such as some building blocks will not be copied over
+      /// Returns a clone of the simulation that can be used during batch run.
+      /// This simulation will contain the minimal information required to run. Configuration settings (e.g. some building blocks) will not be copied over
       /// </summary>
-      /// <param name="simulation"></param>
+      /// <param name="simulation">Simulation to clone</param>
       IModelCoreSimulation CloneForBatchRun(IModelCoreSimulation simulation);
    }
 
--- a/src/OSPSuite.R/Services/SimulationTask.cs
+++ b/src/OSPSuite.R/Services/SimulationTask.cs
@@ -1,36 +1,47 @@
 ﻿using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Core.Domain.Services;
-using OSPSuite.R.Domain;
 
 namespace OSPSuite.R.Services
 {
    public interface ISimulationTask
    {
-      IModelCoreSimulation Clone(IModelCoreSimulation simulation);
+      /// <summary>
+      /// Return a clone of the simulation that can be used during batch run.
+      /// This simulation will contain the minimal information required to run. Configuration settings such as some building blocks will not be copied over
+      /// </summary>
+      /// <param name="simulation"></param>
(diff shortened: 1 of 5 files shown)


msevestre:
This is good. We removed R altogether from this issue.
@Yuri05 finally has an issue worth looking at;-)

Yuri05:
Nice 🙂

Yuri05:
fun to debug :)
I temporarily extended the code in `SimModelBatch.CreateAndFinalizeSimulation` with this:

grafik

This writes the xmls which are passed to SimModel for calculation.
In the simulation, exactly 1 output variable is set as persistable: 
grafik

Now when some simulations do not retrieve any results: this output variable is set as non-persistable!
grafik

Here the xmls of a failing run: simulation 27 is the one with persistable = 0
[xmls.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/9715068/xmls.zip)

So what happens: SimModel sometimes becomes an input with 0 required outputs, calculates and returns empty set of outputs, because none of the outputs is marked as to be returned.
SimModel should surely check for this and either produce a warning or stop with the error (not sure what's better) - but the **initial issue is apparently the SimModel-XML creation in core.** 
Something is not thread safe there...

msevestre:
SO GOOD @Yuri05 
debug master did it again.
Ok so we need to make SimModel generation thread safe!. WTF. THe service needs to be instantiated on each Manager instead of recup once and cached. This is probably an easy fix

rwmcintosh:
once again, bugs go in to SimModel and then bounce straight back out. Thanks @Yuri05

msevestre:
So we are already creating a new instance of the simulation exporter
```
   private SimulationExport createSimulationExport(IModelCoreSimulation simulation, SimModelExportMode simModelExportMode, IReadOnlyList<string> variableMoleculePaths)
      {
         var simulationExportCreator = _simulationExportCreatorFactory.Create();
         var simulationExport = simulationExportCreator.CreateExportFor(simulation.Model, simModelExportMode, variableMoleculePaths);
         simulationExport.AddSimulationConfiguration(simulation.SimulationSettings);
         return simulationExport;
      }
```
So the quick fix that I was hoping is not happening

msevestre:
Ok I think I have it....

src/OSPSuite.R/Services/SimulationTask.cs
```diff
@@ -1,38 +1,49 @@
...
-         // None of the other properties are required to complete the simulation
-         var simulationBuildConfiguration = new BuildConfiguration
-         {
-            SimulationSettings = _cloneManagerForModel.Clone(simulationToClone.SimulationSettings)
-         };
-         simulation.BuildConfiguration = simulationBuildConfiguration;
+
+         //Once we prepare a simulation for a batch, we are assuming that it won't be changed by the caller.
+         //We update the persistable at this stage to avoid any possible thread problems
+         _simulationPersistableUpdater.UpdateSimulationPersistable(simulation);
```
Yuri05, line 44: so the problem was that cloning did not properly copy the simulation settings and thus the curve was not set as persistable?
But why did it happen only sometimes and not every time?

msevestre:
Damm I had written all those comments yesterday in the PR and forgot to submit review. I HATE THIS FEATURE IN GITHUB

msevestre:
going to merge now. I think this will solve all our issues 
FYI @PavelBal

src/OSPSuite.R/Services/SimulationTask.cs
Yuri05: ok, I see. cool, cool :)

PavelBal:
Awesome. Once PK-Sim is updated, we should also update Core in R and should be able to use simulation batches again.

PR #751, opened 2020-03-16
WIP #750 add more pk

msevestre:

PR #1715, opened 2022-10-06
Fixes #751 comparison not including container

msevestre:


src/OSPSuite.Core/Comparison/ParameterDiffBuilder.cs
```diff
@@ -28,6 +28,8 @@ public override void Compare(IComparison<IParameter> comparison)
          else
             CompareValues(x => x.RHSFormula != null, Captions.Diff.IsStateVariable, comparison);
 
+         CompareValues(x => x.ContainerCriteria, x => x.ContainerCriteria, comparison);
```
msevestre, line 31: this is the chnage. All the rest is typo correction and test

rwmcintosh (approved):

PR #1716, opened 2022-10-06
Fixes try to work around 4k issue

msevestre:


src/OSPSuite.UI/UIConstants.cs
```diff
@@ -77,6 +77,9 @@ private static double createScaleFactor()
          {
             using (var graphics = Graphics.FromHwnd(IntPtr.Zero))
             {
+               if (graphics.DpiX >= 250)
```
msevestre, line 80: @rwmcintosh This is the place that I wanted to try out. If the ration is slightly bigger, I bet that most of @PavelBal issues will disappear. I do not know what the values is with a 4k resolution and scalign 200%
Can you check what graphics.Dpix you have?
Maybe send Pavel a small test program this he can start to know what HE has locally? That way, by just adding a bit more, we should be good to go. Every hard code height is using this factor so it will be done everywhere at once

rwmcintosh: I have this one that I'll try to reproduce first then see if this fixes it.
https://github.com/Open-Systems-Pharmacology/MoBi/issues/740

rwmcintosh: That one is better 
https://github.com/Open-Systems-Pharmacology/MoBi/issues/757

rwmcintosh: When I use 4k @ 200% scaling, Dpix = 192 and I can reproduce this issue.

Note: you have to restart MoBi after you set the scaling.

rwmcintosh: An observation here - these cut-off text boxes are all of the same specific type in DevExpress - ValueEdit. The ButtonEdit, TextEdit, and ComboBox are not affected

msevestre: Ok so this does not fix anything because we were already under 200
Hum. Maybe we should reduce the first cut off then to be lower than 200?

rwmcintosh: It's this line here:
https://github.com/Open-Systems-Pharmacology/MoBi/blob/0b2ee77da3abb00cf795bf6b9fd5ea1cc32d432a/src/MoBi.UI/Views/EditDistributedParameterView.cs#L175

Commenting that out gives nice result

image

msevestre: I don't get It. What's this tablepropeties stuff? 
There are some hidden roe showing up based on what you show and hide. Let's review quickly in a few minutes?

rwmcintosh: Not sure exactly what's the code for, but it's added with core update from here.
https://github.com/Open-Systems-Pharmacology/MoBi/pull/708

msevestre: You wrote that code ...;-)

rwmcintosh: Really? It was brought into MoBi from somewhere else before?

msevestre: it's in MOBI? Why?

msevestre: oh I thought you meant the scaling stuff sorry

rwmcintosh: Lol. yeah. There's another case of it that I'm looking at, but commenting it out makes the problem worse, not better.

https://github.com/Open-Systems-Pharmacology/MoBi/issues/740
and the line
https://github.com/Open-Systems-Pharmacology/MoBi/blob/7e894acf391be4e9c299f57550cb732785b047aa/src/MoBi.UI/Views/EditParameterView.cs#L114

rwmcintosh:
I guess we can close this PR. We found out nothing is simple after all!

PR #1718, opened 2022-10-17
Fixes #1636 show warning

### LINKED ISSUE #1636: Show a warning message when number of output points is high
As we have removed the limit points in the output interval, the software can sometimes fail to plot the specified number of points (or the calculation time explodes). Sometimes this happens when the user accidentially selects wrong unit for the resolution. I propose to show a warning before starting the simultation when the number of points exceeds a certain threshold.

Warning: "The selected output resolution will generate xxx points and may severy impact the software performance. Are you sure you want to run with these setting? If not, consider changing output resolution in simulations settings".

29 days, two output curves:

- 4 pts/h - ok
- 20 pts/h - ok 
- 60 pts/h - ok
- 120 pts/h - already slow

So I would propose to show a warning when the number of output points exceeds `2*29*24*60*100` = 8352000 points (8300000).

msevestre:


msevestre:
THis will need to be merged so that I can do the plumbing in MoBi and PKSim

rwmcintosh (approved):

PR #1722, opened 2022-10-19
Small improvements to deviation lines in plots

msevestre:


msevestre:
@georgeDaskalakis FYI.

PR #1723, opened 2022-10-19
Fixes #1635 Colour grouping button layout

### LINKED ISSUE #1635: "Apply Color Grouping" button layout issue
image

rwmcintosh:
I made a newer branch for this fix because there were some more changes since @georgeDaskalakis branch was created.


georgeDaskalakis (approved):

msevestre:
I don't understand this change.
tHe goal is to have consistent size button throughout the app.
Maybe this is not something we should do but we have other instances where we use this method. Aren't we going to have the same problem?

rwmcintosh:
Yes, I expect this will happen elsewhere. I checked the implemented layout with Spy++ when scaled and the dimensions reported are correct, but the button still does not show all the text. I'll open an issue about the general case. We will have to try to understand the relationship between pixels and scaling sometime soon.

PR #1747, opened 2022-11-07
Fixes #1746 Create typed way to use quantile extension methods

### LINKED ISSUE #1746: Create typed way to use quantile extension methods
It's not clear from the developers perspective that the float arrays need to be sorted before using the Median, Percentile, and Quantile extension methods.

Open-Systems-Pharmacology/PK-Sim#2410

rwmcintosh:


src/OSPSuite.Core/Maths/SortedFloatArray.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System.Collections.Generic;
+using OSPSuite.Core.Extensions;
+
+namespace OSPSuite.Core.Maths
+{
+   public class SortedFloatArray
```
msevestre, line 6: can we make this SortedArray of T ?

msevestre (approved):

PR #1751, opened 2022-11-08
1749 default scaling when changing output

### LINKED ISSUE #1749: 11.1 Observed data mapping in a simulation: Default Scaling
@Yuri05 commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2385)

When we add an observed data set in a Parameter IDentification: Default Scaling is Log:

But when adding obs data to a simulation: default scaling is Linear:

Should we make it consistent and set the **default scaling for obs data in a simulation to Log** as well?

---

@PavelBal commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2385#issuecomment-1301946117)

Consistent would be good, if we (@georgeDaskalakis ) understand how the defaults are selected. For PI, the default scaling is not always log but depends on the selected output. If we can easily re-use the code - than yes, it should also be used in the simulation.

---

@Yuri05 commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2385#issuecomment-1301969305)

>or PI, the default scaling is not always log but depends on the selected output. 

makes sense :)

It is defined here:
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/a3aa4aeaac5f62561d8c3f62f76e8a2a286599c0/src/OSPSuite.Core/Services/OutputMappingMatchingTask.cs#L63-L66

---

@PavelBal commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2385#issuecomment-1301994879)

@georgeDaskalakis I think this will a be quick fix, right?

---

@msevestre commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2385#issuecomment-1302046549)

This should absolutely be the same and the same code should be reused (not copy paste please).

---

@georgeDaskalakis commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2385#issuecomment-1302381379)

I cannot reproduce it tbh. The code that you are pointing to @Yuri05 is also the exact same that is being used for the mapping in Simulation: it was moved from Parameter Identification to this newly created OutputMappingMatchingTask, and the simulation mapping happens here:

https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/a3aa4aeaac5f62561d8c3f62f76e8a2a286599c0/src/OSPSuite.Core/Services/OutputMappingMatchingTask.cs#L59

when a new observed data is added. 

What does happen for sure is that when a matching output is not found the scaling is left to "linear", but this also happens when I add an output mapping to PI. Do you maybe have a project for me to reproduce?

---

@Yuri05 commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2385#issuecomment-1302455596)

>Do you maybe have a project for me to reproduce?

e.g. the project from https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2384

---

@Yuri05 commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2385#issuecomment-1302464338)

>What does happen for sure is that when a matching output is not found the scaling is left to "linear"

That's probably the reason. Of course, the scaling could be deduced either from the **observed** data and not from the mapped simulated output (simulated output mapped later on must have the same dimension anyway). Or as soon as a simulation output is mapped by user.

The latter happens actually in a PI. When adding new output mapping - it's set to linear. But when I e.g. select a concentration simulation output - it automatically changes to log.

Not a big issue for me though.

---

@georgeDaskalakis commented on [Fri Nov 04 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2385#issuecomment-1303584892)

@Yuri05 yeap this does makes sense: in Simulations the default scaling is being calculated and set only when automatically mapping newly added observed data to an output, not when changing the output selection. Should not be that hard to add this, if I cna implement it fast I will.

georgeDaskalakis:
closes #1749

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
@@ -6,7 +6,6 @@ using OSPSuite.Core.Domain.ParameterIdentifications;
 using OSPSuite.Core.Domain.Repositories;
 using OSPSuite.Core.Domain.Services;
 using OSPSuite.Core.Events;
-using OSPSuite.Core.Services;
 using OSPSuite.Presentation.DTO;
 using OSPSuite.Presentation.Mappers;
 using OSPSuite.Presentation.Views;
@@ -48,7 +47,6 @@ namespace OSPSuite.Presentation.Presenters
       private readonly ISimulationOutputMappingToOutputMappingDTOMapper _outputMappingDTOMapper;
       private readonly IQuantityToSimulationQuantitySelectionDTOMapper _simulationQuantitySelectionDTOMapper;
       private readonly List<SimulationQuantitySelectionDTO> _allAvailableOutputs = new List<SimulationQuantitySelectionDTO>();
-      private readonly IOutputMappingMatchingTask _outputMappingMatchingTask;
 
       private readonly NotifyList<SimulationOutputMappingDTO> _listOfOutputMappingDTOs;
 
@@ -66,7 +64,6 @@ namespace OSPSuite.Presentation.Presenters
          _observedDataRepository = observedDataRepository;
          _outputMappingDTOMapper = outputMappingDTOMapper;
          _simulationQuantitySelectionDTOMapper = simulationQuantitySelectionDTOMapper;
-         _outputMappingMatchingTask = new OutputMappingMatchingTask(_entitiesInSimulationRetriever);
          _listOfOutputMappingDTOs = new NotifyList<SimulationOutputMappingDTO>();
       }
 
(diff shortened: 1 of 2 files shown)


Yuri05 (approved):

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
```diff
@@ -149,6 +152,7 @@ public void UpdateSimulationOutputMappings(SimulationOutputMappingDTO simulation
          MarkSimulationAsChanged();
          if (!_simulation.OutputMappings.OutputMappingsUsingDataRepository(simulationOutputMappingDTO.ObservedData).Any())
             _simulation.OutputMappings.Add(simulationOutputMappingDTO.Mapping);
+         simulationOutputMappingDTO.Scaling = _outputMappingMatchingTask.DefaultScalingFor(simulationOutputMappingDTO.Output.Quantity);
```
msevestre, line 155: there is a space missing here. I am going to add it by hand.

PR #1754, opened 2022-11-08
1752 deviation lines in legend

### LINKED ISSUE #1752: 11.1 GOF plots - "deviation lines": Problem #2
@Yuri05 commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2389)

x-fold range is not shown in the legend as per default:
grafik

I would suggest the following: 
* call both lines (upper and lower) just "x-fold deviation" or "x-fold range" (without "upper/lower").
* add one of those curves to the legend when adding a new deviation range

Thus it would look like this:
grafik

---

@georgeDaskalakis commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2389#issuecomment-1302361249)

> I would suggest the following:
> 
> * call both lines (upper and lower) just "x-fold deviation" or "x-fold range" (without "upper/lower").

@Yuri05 we had it that way and it was changed after discussion (@msevestre 's idea if I remember correctly). I also agree it is a bit confusing to have two lines named exactly the same. 

---

@georgeDaskalakis commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2389#issuecomment-1302363170)

about the legend, I guess we could keep the naming and add both? It is true that then the legend doesn't look that pretty, but the  functional part of the "Curve and Axis Settings" tab makes more sense.

---

@Yuri05 commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2389#issuecomment-1302470575)

or we rename only 1 and show it in the legend and keep the 2nd AS IS 😉 
grafik

---

@msevestre commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2389#issuecomment-1302475706)

@georgeDaskalakis No having the same name for **both** does not make sense to me so it's definitely not my idea unless I was on drugs

I think only one needs to be shown in the legend and I think what @Yuri05 is proposing makes sense until we have presenters that actually fit this use case

---

@PavelBal commented on [Fri Nov 04 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2389#issuecomment-1303144136)

>  I was on drugs

That explains a lot of stuff... No kidding :D What Georgios ment is that it was your idea to have the "upper" nad "lower" in the naming so that the names are **not** equal.

I think only **one** entry should be shown in the legend. I actually do not see a problem to have the same name for both of them, as they even have the same color and style, so it is kind of the same line but shown twice...

georgeDaskalakis:
closes #1752

### DIFF
--- a/src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
+++ b/src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
@@ -120,6 +120,11 @@ namespace OSPSuite.Core.Services
          deviationCurves.Where(repository => repository != null).Each(repository =>
             chart.AddCurvesFor(repository, x => x.Name, _dimensionFactory, (column, curve) => curve.UpdateDeviationCurve(column.Name, deviationLinesCount)));
 
+         //adding one of the deviation lines to the legend. Workaround for now until the presenter is adjusted for Predicted vs. Observed Chart.
+         var upperDeviationLine = chart.Curves.FirstOrDefault(curve => curve.Name.Equals(Captions.ParameterIdentification.DeviationLineNameUpper(foldValue)));
+         if (upperDeviationLine != null)
+            upperDeviationLine.VisibleInLegend = true;
+
          chart.UpdateAxesVisibility();
          return deviationCurves;
       }
--- a/src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
+++ b/src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
@@ -121,7 +121,7 @@ namespace OSPSuite.Core.Services
             chart.AddCurvesFor(repository, x => x.Name, _dimensionFactory, (column, curve) => curve.UpdateDeviationCurve(column.Name, deviationLinesCount)));
 
          //adding one of the deviation lines to the legend. Workaround for now until the presenter is adjusted for Predicted vs. Observed Chart.
-         var upperDeviationLine = chart.Curves.FirstOrDefault(curve => curve.Name.Equals(Captions.ParameterIdentification.DeviationLineNameUpper(foldValue)));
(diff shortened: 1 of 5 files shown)


Yuri05 (approved):

src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
```diff
@@ -120,6 +120,11 @@ public void ConfigureAxesDimensionAndTitle(IReadOnlyList<DataColumn> observation
          deviationCurves.Where(repository => repository != null).Each(repository =>
             chart.AddCurvesFor(repository, x => x.Name, _dimensionFactory, (column, curve) => curve.UpdateDeviationCurve(column.Name, deviationLinesCount)));
 
+         //adding one of the deviation lines to the legend. Workaround for now until the presenter is adjusted for Predicted vs. Observed Chart.
+         var upperDeviationLine = chart.Curves.FirstOrDefault(curve => curve.Name.Equals(Captions.ParameterIdentification.DeviationLineNameUpper(foldValue)));
```
msevestre, line 124: use FindByName here
Also please import statics  Captions.ParameterIdentification
Lastly, this DeviationLineUpper function should not be under the ParameterIdentfication "class" because it's user outside of it. If you are on it already?

msevestre (changes requested):
Simplify code by using our extension methods please

src/OSPSuite.Core/Services/PredictedVsObservedChartService.cs
georgeDaskalakis: My personal taste is to have the captions.... namespace by the call - we also have it like that all over the place, but if you insist I can change it.

Everything else is fixed

PR #1756, opened 2022-11-08
1748 output mapping wrong namespace

msevestre:


src/OSPSuite.Core/Services/ResidualsVsTimeChartService.cs
```diff
@@ -27,12 +30,19 @@ public interface IResidualsVsTimeChartService
...
       /// </summary>
       DataRepository CreateScatterDataRepository(string id, string repositoryName, OutputResiduals outputResidual);
+
+      void ConfigureChartAxis(AnalysisChartWithLocalRepositories chart);
+
+      /// <summary>
+      ///    Get or creates the DataRepository for the given residuals <paramref name="outputResidual" /> for the given chart.
+      ///    If it already exists, update the values with the output residuals values (update of chart)
+      /// </summary>
+      DataRepository GetOrCreateScatterDataRepositoryInChart(AnalysisChartWithLocalRepositories chart, OutputResiduals outputResidual, int? runIndex = null);
```
msevestre, line 40: There is so much code duplication between the two presenters. I don't quite understand why since we have a service exactly for this purpose. I started to move some so that I could fix the issue only in one place. But at the end of the day, the unit fo residual is too m uch work. I still suggest to keep my chanages

msevestre (commented):
please see changes

Yuri05 (approved):

PR #1765, opened 2022-11-10
1763 delete observed data entry 

### LINKED ISSUE #1763: 11.1: Observed data mapping / Delete Entry
@Yuri05 commented on [Wed Nov 02 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2380)

The tooltip for **X** says "Delete entry", but it just clears the mapping to the simulation output.
Under "Delete Entry" I would usually understand that the entry is completely removed from the table (meaning the observed data set is removed from the simulation).
So we should either change the tooltip to something like "Clear mapped simulation output" or the click on **X** should trigger "Remove observed data set from the simulation" use case.
I would vote for the latter ("Remove obs data set")
grafik

---

@Yuri05 commented on [Wed Nov 02 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2380#issuecomment-1300559201)

@msevestre @PavelBal @rwmcintosh FYI

---

@PavelBal commented on [Wed Nov 02 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2380#issuecomment-1300654334)

- Delete entry
- "Unmapping" should be done via selecting "<None>" from the "Simulation Ouputs"-drop-down
- ONLY implement if max. 1 days

---

@georgeDaskalakis commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2380#issuecomment-1301799788)

The existing way of removing observed data from the simulation (right click on the list and remove), should still be available, right?
 

---

@PavelBal commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2380#issuecomment-1301852599)

> The existing way of removing observed data from the simulation (right click on the list and remove), should still be available, right?

Yes

georgeDaskalakis:
closes #1763

### DIFF
--- a/src/OSPSuite.Core/Domain/ISimulation.cs
+++ b/src/OSPSuite.Core/Domain/ISimulation.cs
@@ -13,5 +13,6 @@ namespace OSPSuite.Core.Domain
 
       //ResultsDataRepository will be null for Population Simulations and should never be called for them
       DataRepository ResultsDataRepository { get; set; }
+      void RemoveUsedObservedData(DataRepository dataRepository);
    }
 }
--- a/src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
@@ -49,6 +49,7 @@ namespace OSPSuite.Presentation.Presenters
       private readonly IQuantityToSimulationQuantitySelectionDTOMapper _simulationQuantitySelectionDTOMapper;
       private readonly List<SimulationQuantitySelectionDTO> _allAvailableOutputs = new List<SimulationQuantitySelectionDTO>();
       private readonly IOutputMappingMatchingTask _outputMappingMatchingTask;
+      private readonly IEventPublisher _eventPublisher;
 
       private readonly NotifyList<SimulationOutputMappingDTO> _listOfOutputMappingDTOs;
 
@@ -60,12 +61,14 @@ namespace OSPSuite.Presentation.Presenters
          IEntitiesInSimulationRetriever entitiesInSimulationRetriever,
          IObservedDataRepository observedDataRepository,
          ISimulationOutputMappingToOutputMappingDTOMapper outputMappingDTOMapper,
-         IQuantityToSimulationQuantitySelectionDTOMapper simulationQuantitySelectionDTOMapper) : base(view)
+         IQuantityToSimulationQuantitySelectionDTOMapper simulationQuantitySelectionDTOMapper,
+         IEventPublisher eventPublisher) : base(view)
       {
          _entitiesInSimulationRetriever = entitiesInSimulationRetriever;
          _observedDataRepository = observedDataRepository;
          _outputMappingDTOMapper = outputMappingDTOMapper;
          _simulationQuantitySelectionDTOMapper = simulationQuantitySelectionDTOMapper;
+         _eventPublisher = eventPublisher;
          _outputMappingMatchingTask = new OutputMappingMatchingTask(_entitiesInSimulationRetriever);
          _listOfOutputMappingDTOs = new NotifyList<SimulationOutputMappingDTO>();
       }
@@ -163,8 +166,12 @@ namespace OSPSuite.Presentation.Presenters
 
       public void RemoveOutputMapping(SimulationOutputMappingDTO outputMappingDTO)
       {
-         outputMappingDTO.Output = null;
+         _simulation.RemoveUsedObservedData(outputMappingDTO.ObservedData);
          _simulation.OutputMappings.Remove(outputMappingDTO.Mapping);
+
+         _eventPublisher.PublishEvent(new ObservedDataRemovedFromAnalysableEvent(_simulation, outputMappingDTO.ObservedData));
+         _eventPublisher.PublishEvent(new SimulationStatusChangedEvent(_simulation));
+
          _view.RefreshGrid();
       }
 
@@ -177,5 +184,16 @@ namespace OSPSuite.Presentation.Presenters
       {
          updateOutputMappingList();
       }
+
+      /*
+      private void removeUsedObservedDataFromSimulation(SimulationOutputMappingDTO outputMappingDTO)
+      {
+         _simulation.RemoveUsedObservedData(outputMappingDTO.ObservedData);
+         _simulation.OutputMappings.Remove(outputMappingDTO.Mapping);
+
+         _eventPublisher.PublishEvent(new ObservedDataRemovedFromAnalysableEvent(_simulation, outputMappingDTO.ObservedData));
+         _eventPublisher.PublishEvent(new SimulationStatusChangedEvent(_simulation));
+      }
+*/
    }
 }
--- a/src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
@@ -184,16 +184,5 @@ namespace OSPSuite.Presentation.Presenters
       {
          updateOutputMappingList();
       }
-
-      /*
-      private void removeUsedObservedDataFromSimulation(SimulationOutputMappingDTO outputMappingDTO)
-      {
-         _simulation.RemoveUsedObservedData(outputMappingDTO.ObservedData);
-         _simulation.OutputMappings.Remove(outputMappingDTO.Mapping);
-
-         _eventPublisher.PublishEvent(new ObservedDataRemovedFromAnalysableEvent(_simulation, outputMappingDTO.ObservedData));
-         _eventPublisher.PublishEvent(new SimulationStatusChangedEvent(_simulation));
-      }
-*/
    }
 }
--- a/src/OSPSuite.R/Domain/Simulation.cs
+++ b/src/OSPSuite.R/Domain/Simulation.cs
@@ -22,6 +22,10 @@ namespace OSPSuite.R.Domain
       public IEnumerable<CurveChart> Charts { get; } = new List<CurveChart>();
       public OutputMappings OutputMappings { get; set; }
       public DataRepository ResultsDataRepository { get; set; }
+      public void RemoveUsedObservedData(DataRepository dataRepository)
+      {
+         // nothing to do in R
+      }
 
       public void RemoveAnalysis(ISimulationAnalysis simulationAnalysis)
       {
--- a/src/OSPSuite.UI/Views/SimulationOutputMappingView.cs
+++ b/src/OSPSuite.UI/Views/SimulationOutputMappingView.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
 using DevExpress.Utils;
 using DevExpress.XtraEditors.Repository;
 using DevExpress.XtraGrid.Views.Base;
@@ -111,13 +112,9 @@ namespace OSPSuite.UI.Views
 
       private RepositoryItem allOutputsRepository(SimulationOutputMappingDTO dto)
       {
-         return RepositoryItemFor(_presenter.AllAvailableOutputs, _outputRepository);
-      }
-
-      protected RepositoryItem RepositoryItemFor<T>(IEnumerable<T> allItems, UxRepositoryItemComboBox listRepositoryItems)
-      {
-         listRepositoryItems.FillComboBoxRepositoryWith(allItems);
-         return listRepositoryItems;
+         var outputsList = _presenter.AllAvailableOutputs.Concat(null); 
+         _outputRepository.FillComboBoxRepositoryWith(_presenter.AllAvailableOutputs);
+         return _outputRepository;
       }
 
       public override string Caption => Captions.SimulationUI.ObservedDataSelection;
--- a/tests/OSPSuite.Presentation.Tests/Presentation/SimulationOutputMappingPresenterSpecs.cs
+++ b/tests/OSPSuite.Presentation.Tests/Presentation/SimulationOutputMappingPresenterSpecs.cs
@@ -14,6 +14,7 @@ using OSPSuite.Presentation.DTO;
 using OSPSuite.Presentation.Mappers;
 using OSPSuite.Presentation.Presenters;
 using OSPSuite.Presentation.Views;
+using OSPSuite.Utility.Events;
 
 namespace OSPSuite.Presentation.Presentation
 {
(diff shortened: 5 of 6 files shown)


georgeDaskalakis:
@msevestre this is ready for review

msevestre:
It's still draft.can you update?

src/OSPSuite.Core/Domain/ISimulation.cs
```diff
@@ -12,5 +12,6 @@ public interface ISimulation : ILazyLoadable, IAnalysable, IUsesObservedData, IM
 
       //ResultsDataRepository will be null for Population Simulations and should never be called for them
       DataRepository ResultsDataRepository { get; set; }
+      void RemoveUsedObservedData(DataRepository dataRepository);
```
msevestre, line 15: Shouldn't that be part of IUsesObservedData?
More of a question here.

src/OSPSuite.Presentation/Mappers/SimulationOutputMappingToOutputMappingDTOMapper.cs
```diff
@@ -22,6 +22,9 @@ public SimulationOutputMappingDTO MapFrom(OutputMapping outputMapping, IReadOnly
 
       private static bool matches(OutputMapping outputMapping, SimulationQuantitySelectionDTO simulationQuantitySelectionDTO)
       {
+         if (simulationQuantitySelectionDTO == null)
```
msevestre, line 25: How can this be null? It's a DTO

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
```diff
@@ -23,7 +26,7 @@ public interface ISimulationOutputMappingPresenter : IPresenter<ISimulationOutpu
    {
       void EditSimulation(ISimulation simulation);
       IReadOnlyList<SimulationQuantitySelectionDTO> AllAvailableOutputs { get; }
-      void RemoveOutputMapping(SimulationOutputMappingDTO outputMappingDTO);
+      void RemoveObservedData(SimulationOutputMappingDTO outputMappingDTO);
```
msevestre, line 29: Hum...Aren't we remoeving an OutputMapping? Why did we rename this?

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
```diff
@@ -79,6 +88,10 @@ private void updateOutputLists()
          _allAvailableOutputs.Clear();
          var outputs = _entitiesInSimulationRetriever.OutputsFrom(_simulation);
          _allAvailableOutputs.AddRange(outputs.Select(x => mapFrom(_simulation, x)).OrderBy(x => x.DisplayString));
+
+         //adding a none DTO output, in order to be able to select it and delete the outputMapping
+         var noneEntry = new SimulationQuantitySelectionDTO(_simulation, null, Captions.SimulationUI.NoneEditorNullText);
```
msevestre, line 93: I tnink this should be a member of the presenter so that you can reuse it later instead of a local variable

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
```diff
@@ -127,24 +140,64 @@ private IEnumerable<DataRepository> allAvailableObservedDataUsedBySimulation()
       public void UpdateSimulationOutputMappings(SimulationOutputMappingDTO simulationOutputMappingDTO)
       {
          MarkSimulationAsChanged();
+
+         if (simulationOutputMappingDTO.Output.DisplayString.Equals(Captions.SimulationUI.NoneEditorNullText))
```
msevestre, line 144: yes. Use the variable instead of comparing display string. 
e.g. if(Equals(simulationOutputMappingDTO.Output, _noneEntry))

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
```diff
@@ -127,24 +140,64 @@ private IEnumerable<DataRepository> allAvailableObservedDataUsedBySimulation()
...
+         {
+            removeOutputMapping(simulationOutputMappingDTO);
```
msevestre, line 146: I absolutely don't get it. Why would we have to remove a NULL entry from the simulation? This does not make sense to me

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
```diff
@@ -127,24 +140,64 @@ private IEnumerable<DataRepository> allAvailableObservedDataUsedBySimulation()
...
       public void MarkSimulationAsChanged()
       {
          _simulation.HasChanged = true;
       }
 
-      public void RemoveOutputMapping(SimulationOutputMappingDTO outputMappingDTO)
+      public void RemoveObservedData(SimulationOutputMappingDTO outputMappingDTO)
       {
-         outputMappingDTO.Output = null;
+         var parameterIdentifications = findParameterIdentificationsUsing(outputMappingDTO.ObservedData).ToList();
```
msevestre, line 170: Why is this done in a presneter? This looks like code that should be done at the task level. This presenter has nothing to do with ParameterIdentification. Why is this dealing with this concerns suddenly?

msevestre (changes requested):
I think the code is good, but not in the right places.

msevestre:
@georgeDaskalakis Happy to review with you if you want

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
georgeDaskalakis: when we click the "x" button we remove the observed data. to remove the output mapping, the <None> entry should be selected from the output dropdown.

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -201,6 +201,7 @@ public static class Captions
       public static readonly string ChartExportOptions = "Chart Export Options";
       public static readonly string No = "No";
       public static readonly string Yes = "Yes";
+      public static readonly string ReallyRemoveObservedDataFromSimulation = $"Really remove {ObjectTypes.ObservedData} from simulation?\nHint: {ObjectTypes.ObservedData} will not be deleted from project";
```
Yuri05, line 204: I think it should be "... from **the** simulation" and "...from **the** project"

georgeDaskalakis:
> @georgeDaskalakis Happy to review with you if you want

@msevestre sure, let's do that tomorrow

src/OSPSuite.Presentation/Mappers/SimulationOutputMappingToOutputMappingDTOMapper.cs
georgeDaskalakis: true, removed

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
georgeDaskalakis: done

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
georgeDaskalakis: when the output is set to null, the outputMapping should not be in the simulation (if you remember we have already discussed about this), so we have to explicitely delete it in order not to have OutputMappings with null outputs in the simulation.

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
georgeDaskalakis: You are right, I have overlooked this and I totally agree. Then the task will also be used in PK-Sim and MoBi - since we are going to keep the existing way of removing observed data from simulations.

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
msevestre: I see. Discutable logic but this is what the issue is about so I understand the change. Thanks

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
georgeDaskalakis: true, setting the DTO output to null is not anymore necessary (come sfrom the way the code was before), so I am removing. But about removing the outputMapping from the Simulation, why not? The Presenter was doing it till now anyway (just in a different place, when hitting the "x" button). I mean, it is the SimulationOutputMappingPresenter, and removing it from the simulation is an one-liner, I definately do not see a need for a service or something like that here.

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
georgeDaskalakis: true, but then I can find no way to unit test this. we can discuss if you want.

src/OSPSuite.Assets/UIConstants.cs
georgeDaskalakis: done

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
georgeDaskalakis: moving the OutputMapping removal code to the simulation itself.

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
georgeDaskalakis: this was moved to the service

georgeDaskalakis:
@msevestre I finished moving the code to the task and tackling everything other than the two unresolved comments above.

src/OSPSuite.Core/Domain/Services/ObservedDataTask.cs
```diff
@@ -166,11 +178,73 @@ public void AddImporterConfigurationToProject(ImporterConfiguration configuratio
          _executionContext.AddToHistory(new AddImporterConfigurationToProjectCommand(configuration).Run(_executionContext));
       }
 
+      public void RemoveUsedObservedDataFromSimulation(IReadOnlyList<UsedObservedData> usedObservedDataList)
+      {
+         if (!usedObservedDataList.Any())
+            return;
+
+         foreach (var usedObservedData in usedObservedDataList)
```
msevestre, line 186: this loop is weird. Feels like this is not a loop at all because you return after the first time the dialog is shown
It feels more like a query to see of this is used and then a message ?
My first instinct was to say: BEWARE, showing a message in a loop. But there is the return on line 195

src/OSPSuite.Core/Domain/Services/ObservedDataTask.cs
```diff
@@ -166,11 +178,73 @@ public void AddImporterConfigurationToProject(ImporterConfiguration configuratio
...
+               Captions.ParameterIdentification.CannotRemoveObservedDataBeingUsedByParameterIdentification(observedDataFrom(usedObservedData).Name,
+                  parameterIdentifications.AllNames().ToList()));
+            return;
+         }
+
+         var viewResult = _dialogCreator.MessageBoxYesNo(Captions.ReallyRemoveObservedDataFromSimulation);
+         if (viewResult == ViewResult.No)
+            return;
+
+         usedObservedDataList.GroupBy(x => x.Simulation).Each(x => removeUsedObservedDataFromSimulation(x, x.Key.DowncastTo<ISimulation>()));
```
msevestre, line 202: x.Key.DowncastTo<ISimulation>()

Why is this so? Are we sure of this at this stage?

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
```diff
@@ -54,14 +57,17 @@ public class SimulationOutputMappingPresenter : AbstractSubPresenter<ISimulation
...
          ISimulationOutputMappingToOutputMappingDTOMapper outputMappingDTOMapper,
-         IQuantityToSimulationQuantitySelectionDTOMapper simulationQuantitySelectionDTOMapper) : base(view)
+         IQuantityToSimulationQuantitySelectionDTOMapper simulationQuantitySelectionDTOMapper,
+         IObservedDataTask observedDataTask) : base(view)
       {
          _entitiesInSimulationRetriever = entitiesInSimulationRetriever;
          _observedDataRepository = observedDataRepository;
          _outputMappingDTOMapper = outputMappingDTOMapper;
          _simulationQuantitySelectionDTOMapper = simulationQuantitySelectionDTOMapper;
+         _noneEntry = new SimulationQuantitySelectionDTO(null, null, Captions.SimulationUI.NoneEditorNullText);
```
msevestre, line 67: Is this going to be a problem if the simulation is null?

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
msevestre: hum... Not sure what you mean.  You need to test the behaviour, not the implementation .We can check it out tomorrow

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
```diff
@@ -138,11 +155,15 @@ public void MarkSimulationAsChanged()
...
 
-      public void RemoveOutputMapping(SimulationOutputMappingDTO outputMappingDTO)
+      public void RemoveObservedData(SimulationOutputMappingDTO outputMappingDTO)
       {
-         outputMappingDTO.Output = null;
-         _simulation.OutputMappings.Remove(outputMappingDTO.Mapping);
-         _view.RefreshGrid();
+         var usedObservedData = new UsedObservedData
+         {
+            Id = outputMappingDTO.ObservedData.Id,
```
msevestre, line 162: Why are we using an id here an not the ObservedData itself. This USedObservedData does not feel like the right structure to me (this might be used like this somewhere else but in this case..I don't get it. Later, you are going through the project to retrieve the observed data by id. all of that is not required if we just have a reference to the observed data?

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
```diff
@@ -138,11 +155,15 @@ public void MarkSimulationAsChanged()
...
+            Simulation = _simulation
+         };
+
+         _observedDataTask.RemoveUsedObservedDataFromSimulation(new List<UsedObservedData>() { usedObservedData });
```
msevestre, line 166: you can just pass new []{} . This is also an IReadOnlyList

msevestre (commented):
Yeah this looks much better. Let's reivew this test stuff tomorrow

msevestre (approved):

PR #1766, opened 2022-11-10
Fixes #1764 refresh mapped observed data

### LINKED ISSUE #1764: 11.1 Observed data tab in a simulation not refreshed if previously mapped simulation output was removed
@Yuri05 commented on [Fri Nov 04 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2402)

If I map a simulation output, then re-run the simulation via "Define settings and run" and remove this simulation output: it is still shown as mapped in the Observed Data Tab.
After Closing and Reopen the simulation: the mapping is set to <None> (as expected)

---

@msevestre commented on [Mon Nov 07 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2402#issuecomment-1305622534)

This needs to be done via Pub Sub events.

rwmcintosh:


src/OSPSuite.Core/Events/AnalysableEvents.cs
```diff
@@ -64,6 +64,14 @@ protected SimulationEvent(ISimulation simulation) : base(simulation)
       }
    }
 
+   public class SimulationOutputSelectionsChangedEvent : SimulationEvent
```
rwmcintosh, line 67: Planning to publish this event in PK-Sim when the settings are re-defined

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
```diff
@@ -177,5 +178,11 @@ public void Handle(ObservedDataRemovedFromAnalysableEvent eventToHandle)
       {
          updateOutputMappingList();
       }
+
+      public void Handle(SimulationOutputSelectionsChangedEvent eventToHandle)
```
msevestre, line 182: This all looks good

But where it the event actually thrown?
This is also a core UI and presenter stuff

msevestre (changes requested):
Event not thrown?

src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
rwmcintosh: It will be right after this line of code
https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/eb50ba0d6970bf7286cc7452e6048aecb344eb39/src/PKSim.Core/Services/InteractiveSimulationRunner.cs#L56

msevestre: not sure about this. This will have to be done in MoBi as well
Ihave the feeling this could be in core

rwmcintosh: I tried in MoBi and it already refreshes somehow. But there is a similar spot for it in MoBi too.

rwmcintosh: There is a common presenter in Core, but this little section of code in PKSim means only PKSim knows if the simulation will actually be updated.

https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/eb50ba0d6970bf7286cc7452e6048aecb344eb39/src/PKSim.Presentation/Presenters/Simulations/SimulationOutputSelectionPresenter.cs#L72-L73

rwmcintosh: Same in MoBi

https://github.com/Open-Systems-Pharmacology/MoBi/blob/e52361e2094b3bfabf13a5d34428d0339d2ae28c/src/MoBi.Presentation/Presenter/OutputSelectionsPresenter.cs#L58-L59

PR #1768, opened 2022-11-11
1767 return empty plot when observed data without nonZero values

georgeDaskalakis:


src/OSPSuite.Presentation/Presenters/SimulationPredictedVsObservedChartPresenter.cs
```diff
@@ -50,6 +50,11 @@ protected override void UpdateAnalysis()
 
          _identityRepositories.AddRange(_predictedVsObservedChartService.AddIdentityCurves(observationColumns, Chart));
 
+         //if no identity repository has been added, resulting in the chart not having been initialized, just return an empty plot
+         //this has been the case for having only one plot with exclusively 0 values
+         if (!Chart.Axes.Any())
```
msevestre, line 55: This does not feel to me like the right check. Isn't it because the line above 
`_predictedVsObservedChartService.AddIdentityCurves(observationColumns, Chart)` does nto return anything ?

Almost feel like we know why this is crashing way down the line. While this check should be in place somewhere else I agree (probably where axis is being used and it's empty), I am wondering if there is not something higher level to represent the change here

msevestre (commented):
not sure about this change

src/OSPSuite.Presentation/Presenters/SimulationPredictedVsObservedChartPresenter.cs
msevestre: probably better. Not ideal, but definitely better

msevestre:
The build seems to be failing

georgeDaskalakis:
> The build seems to be failing

@msevestre  failing test fixed

PR #1776, opened 2022-11-11
Fixes some code update to remove unused method, abide to convention f…

msevestre:
…or naming, and remove hard core query in a loop

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
@@ -2,7 +2,6 @@
 using System.Linq;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Data;
-using OSPSuite.Core.Domain.ParameterIdentifications;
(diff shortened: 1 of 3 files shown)


src/OSPSuite.Presentation/Presenters/SimulationOutputMappingPresenter.cs
```diff
@@ -75,18 +68,20 @@ public void Refresh()
...
             updateOutputMappingList();
             _view.BindTo(_listOfOutputMappingDTOs);
          });
       }
 
-      public void UpdateCache()
+      private void updateOutputLists()
       {
          _allAvailableOutputs.Clear();
+         var outputs = _entitiesInSimulationRetriever.OutputsFrom(_simulation);
```
msevestre, line 80: this code was called in a loop for each output. The cache was not used at all really. I am sure we had discussed creating a cache to avoid calling in a loop so probably this was started but not finihsed

msevestre (commented):
@rwmcintosh I am going to merge so that I can update PKSim and MoBi and fix the remaining issues that I have

PR #1782, opened 2022-11-14
1781 grouping fixed in code

### LINKED ISSUE #1781: 11.1 Chart Editor default grouping
@Yuri05 commented on [Fri Nov 11 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2421)

At the moment the default grouping is **Repository/Category**

I suggest to make it other way around: **Category/Repository**

---

@Yuri05 commented on [Fri Nov 11 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2421#issuecomment-1311641829)

 @georgeDaskalakis @PavelBal @msevestre @rwmcintosh

---

@georgeDaskalakis commented on [Fri Nov 11 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2421#issuecomment-1311645035)

@Yuri05 good point. actually do we need repository as a grouping at all?

---

@Yuri05 commented on [Fri Nov 11 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2421#issuecomment-1311659159)

>actually do we need repository as a grouping at all?

I asked myself the same question :)
We need the info about the repository (otherwise it would be hard to distinguish between different obs data sets or e.g. between different simulations in the simulation comparison charts) - but this info could be shown just as a column, and not as grouping.

So I would prefer:
* group by category only
* show repository as the first column in the table

---

@msevestre commented on [Fri Nov 11 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2421#issuecomment-1311661709)

In the picutre above, it woild be very hard to find the curve. The group span over the whole row so you can actually see what obs data and sim you are using . I'd like to see a picture without grouping by data repo to see how it looks like
I am pretty sure you CANNOT differentiate between the first 4 entries
If you have multiple simulation starting with the same 8-10 name, this would be the same

---

@Yuri05 commented on [Fri Nov 11 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2421#issuecomment-1311676147)

>  I'd like to see a picture without grouping by data repo to see how it looks like

grafik

The repo still should be shown (as column) - so you can differentiate. Just not as grouping

---

@msevestre commented on [Fri Nov 11 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2421#issuecomment-1311679351)

yes but look how wide it is. I am asking for the SAME width as above. 

---

@Yuri05 commented on [Fri Nov 11 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2421#issuecomment-1311691985)

sure, with the small(er) width you will not see the full repo name immediately.
The question is, what more convenient for the user. If e.g. I have 20 repos with observed data sets, each repo with 2 data sets, and I need to select just one data set from each repo: 
* without grouping, I need e.g. to make my chart window wider (1 drag action) and/or make the repo column wider (1 click)
* with grouping: I need to expand all 20 repos (20 clicks)

---

@msevestre commented on [Fri Nov 11 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2421#issuecomment-1311697788)

I don't mind your change. I am just saying that with the default UI, this example will be very hard to use because each line will look the same. This a a trade off

---

@PavelBal commented on [Fri Nov 11 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2421#issuecomment-1311704204)

@georgeDaskalakis Default grouping by category is what we already agreed on, right? I don't understand why this is not the case any more...

georgeDaskalakis:
closes #1781

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
@@ -271,7 +271,7 @@ namespace OSPSuite.Presentation.Presenters.Charts
 
       protected override void SetDefaultColumnSettings()
       {
-         AddColumnSettings(BrowserColumns.RepositoryName).WithCaption(Captions.Chart.DataBrowser.RepositoryName).GroupIndex = 0;
+         AddColumnSettings(BrowserColumns.RepositoryName).WithCaption(Captions.Chart.DataBrowser.RepositoryName).GroupIndex = -1;
          AddColumnSettings(BrowserColumns.Simulation).WithCaption(Captions.SimulationPath);
          AddColumnSettings(BrowserColumns.TopContainer).WithCaption(Captions.TopContainerPath);
          AddColumnSettings(BrowserColumns.Container).WithCaption(Captions.ContainerPath);


msevestre:
@georgeDaskalakis Yes we should reset those. 
I am actually believing we should remove these altogether. This was a good idea but I don't think it's used and the maintenance of these is quite high I find. @Yuri05 thoughts?

src/OSPSuite.Presentation/Presenters/Charts/DataBrowserPresenter.cs
```diff
@@ -271,7 +271,7 @@ private void updateDataSelection(IReadOnlyList<DataColumnDTO> selectedDataColumn
 
       protected override void SetDefaultColumnSettings()
       {
-         AddColumnSettings(BrowserColumns.RepositoryName).WithCaption(Captions.Chart.DataBrowser.RepositoryName).GroupIndex = 0;
+         AddColumnSettings(BrowserColumns.RepositoryName).WithCaption(Captions.Chart.DataBrowser.RepositoryName).GroupIndex = -1;
```
Yuri05, line 274: `-1` means "no grouping"? And what would happen if we just would say 
`AddColumnSettings(BrowserColumns.RepositoryName).WithCaption(Captions.Chart.DataBrowser.RepositoryName);` ?
Looking at the columns below: no grouping

georgeDaskalakis: hehe true, I just wanted to make extra sure. -1 actually actively removes the grouping for this column if there was one till now - specifying none leaves everything as is. Of course the user can then change the groupings in the UI.

Yuri05: Ok. can you add your explanation "-1 actively removes the grouping for this column if there was one till now - specifying none leaves everything as is." as a comment?
Then fine for me :)

georgeDaskalakis: done

Yuri05:
I am always cautious when removing existing functionality. Sometimes it's really justified (e.g. for the pdf reports, which were not maintable anymore).
But in this case I would say: let's keep. Particularly setting the group index of the repository name from 0 to -1 in the only xml where it is not -1 seems doable for me :)

Yuri05:
>I am not saying removing now, but for the next version unless there is a clear usage.

Well, we can create a poll in the forum

PR #1791, opened 2022-11-15
Fixes #1780 Clone of chart does not clone deviation lines

### LINKED ISSUE #1780: Clone of chart does not clone deviation lines
@georgeDaskalakis @rwmcintosh 
There is a method in chart called UpdatePropertiesFrom (sourcexx)

This method need to be implemented in all objects having state .This is used in cloning etc. 
I think this is missing for the Chart with folds being saved. I have fixed the issue with closing in PKSim but this is missing in core
Can you tale a look please?

rwmcintosh:


src/OSPSuite.Core/Chart/PredictedVsObservedChart.cs
```diff
@@ -0,0 +1,66 @@
...
+
+      public void AddToDeviationFoldValue(float foldValue)
+      {
+         if (DeviationFoldValues.Contains(foldValue))
+            return;
+
+         DeviationFoldValues.Add(foldValue);
+      }
+
+      public override void UpdatePropertiesFrom(IUpdatable source, ICloneManager cloneManager)
```
rwmcintosh, line 42: The missing method transfers the foldValues to the new chart but that introduced an issue where if the foldValue was present in the list then the service assumed the curve was present in the chart, but that's not necessarily the case now

src/OSPSuite.Core/Chart/PredictedVsObservedChart.cs
```diff
@@ -0,0 +1,66 @@
...
+      {
+         base.UpdatePropertiesFrom(source, cloneManager);
+         var sourceChart = source as PredictedVsObservedChart;
+
+         sourceChart?.DeviationFoldValues.Each(AddToDeviationFoldValue);
+      }
+
+      public bool HasDeviationCurveFor(float foldValue)
```
rwmcintosh, line 50: The cache keeps track of foldValues for which there is a curve. If the foldValue is in the list and no value in the cache then it needs to be plotted

src/OSPSuite.Core/Chart/PredictedVsObservedChart.cs
```diff
@@ -0,0 +1,66 @@
...
+            return;
+
+         DeviationFoldValues.Add(foldValue);
+      }
+
+      public override void UpdatePropertiesFrom(IUpdatable source, ICloneManager cloneManager)
+      {
+         base.UpdatePropertiesFrom(source, cloneManager);
+         var sourceChart = source as PredictedVsObservedChart;
+
```
msevestre, line 46: typically we exit if the object is not of the right type

src/OSPSuite.Core/Chart/PredictedVsObservedChart.cs
```diff
@@ -0,0 +1,66 @@
...
+using OSPSuite.Core.Domain.UnitSystem;
+using OSPSuite.Core.Extensions;
+using OSPSuite.Utility.Collections;
+using OSPSuite.Utility.Extensions;
+
+namespace OSPSuite.Core.Chart
+{
+   public abstract class PredictedVsObservedChart : AnalysisChart
+   {
+      private readonly Cache<float, IEnumerable<DataRepository>> _deviationRepositoryCache = new Cache<float, IEnumerable<DataRepository>>(onMissingKey: x => null);
```
msevestre, line 16: is there a reason why you are not storing an IReadOnlyList here?

msevestre (commented):
looks good to me. Just one remark for code sync with other UpdatePropertiesFrom and a question regarding usage of IEnumerable

src/OSPSuite.Core/Chart/PredictedVsObservedChart.cs
rwmcintosh: I need to know for each foldValue, if there was a data repository created. That's how I can tell the difference between what the chart is supposed to have, vs what it already has.

rwmcintosh: oh, I see what you mean. forget my comment. above

PR #1793, opened 2022-11-15
Fixes #1792 Deviation and Identity columns are not named in data browser

### LINKED ISSUE #1792: 11.1 Deviation and Identity columns are not named in data browser
@Yuri05 commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2388)

when adding the "deviation lines" to a GoF plot, curve name is set in the curve settings window (lower part) but not in the curves browser window (upper part)

grafik

If I deselect the x-Fold curves and then select them again, I get the following:

grafik

rwmcintosh:


src/OSPSuite.Core/Domain/Mappers/DataColumnToPathElementsMapper.cs
```diff
@@ -36,16 +36,18 @@ public PathElements MapFrom(DataColumn dataColumn, IContainer rootContainer)
...
 
          return pathElementsForUnknownColumns(dataColumn, quantityPath);
       }
 
-      private static PathElements pathElementsForBaseGrid(DataColumn dataColumn)
+      private static PathElements pathElementsForNameOnlyColumn(DataColumn dataColumn)
       {
-         var pathColumnValues = new PathElements();
-         pathColumnValues[PathElementId.Name] = new PathElement {DisplayName = dataColumn.Name};
+         var pathColumnValues = new PathElements
```
msevestre, line 47: Return right away if you are already changing it for the short form

msevestre (approved):
One change and good to go

PR #1795, opened 2022-11-16
Fixes #1794 residuals vs time plot not refreshing

### LINKED ISSUE #1794: 11.1 Residuals vs. Time plot: no refresh
@Yuri05 commented on [Fri Nov 04 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2401)

If I change the settings on the observed data tab in a simulation (scaling or weight): Residual vs. time Plots are not refreshed.
I have to close and reopen the simulation to see the changes.

And if I clear the mapping of observed data set: it is not removed from the plot even after reopening the simulation, so I need to create a new plot (this also applies to Predicted vs Observed plots)

---

@msevestre commented on [Mon Nov 07 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2401#issuecomment-1305623721)

Those are two issues here that I think we should split. 
/1 is not too much of a problem. It's a small bug but why not
/2 needst to be fixed for sure

---

@PavelBal commented on [Mon Nov 14 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2401#issuecomment-1313909430)

tbd, but if implementing 2., the residuals error must be updated, too.

---

@georgeDaskalakis commented on [Tue Nov 15 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2401#issuecomment-1315563032)

accidentally closed. reopening.

rwmcintosh:


src/OSPSuite.Core/Events/AnalysableEvents.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Data;
+using OSPSuite.SimModel;
```
msevestre, line 4: Why was this added? if htis is required, this is PROBABLY NOT The right ISimulation here..

msevestre (changes requested):

src/OSPSuite.Presentation/Presenters/SimulationVsObservedDataChartPresenter.cs
```diff
@@ -85,5 +87,11 @@ protected void AddUsedObservedDataToChart()
       }
 
       protected abstract void AddRunResultToChart();
+      
+      public void Handle(SimulationOutputMappingsChangedEvent eventToHandle)
+      {
+         if (eventToHandle.Simulation == _simulation)
```
msevestre, line 93: this can cras if Simulation is null (I think). so we usually do (Equals()) .
I may be wrong

src/OSPSuite.Core/Events/AnalysableEvents.cs
rwmcintosh: It got added at one point, but it's not used

msevestre (approved):

src/OSPSuite.Presentation/Presenters/SimulationVsObservedDataChartPresenter.cs
rwmcintosh: I'll change it to Equals, but I don't think it should crash. if one is null

msevestre: Ok . Maybe old behavior then

PR #1803, opened 2022-11-18
1802 inconsistent dimension fixed

### LINKED ISSUE #1802: 11.1 Predicted vs Observed plot
@Yuri05 commented on [Thu Nov 03 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2392)

If my default output unit differs from the observed data unit: predicted vs. observed plot is created with those 2 different units on the X- and Y-axis.
grafik

Should in such a case change one of the axis units in order to match the other one?
E.g. change the unit of the simulated data:
grafik

---

@georgeDaskalakis commented on [Fri Nov 11 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2392#issuecomment-1311527197)

Hmm...this seems like something we had during the development phase and I had at some point fixed. I ll take another look.

georgeDaskalakis:
closes #1802

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/SimulationPredictedVsObservedChartPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/SimulationPredictedVsObservedChartPresenter.cs
@@ -4,7 +4,6 @@ using System.Linq;
 using OSPSuite.Assets;
 using OSPSuite.Core.Chart;
 using OSPSuite.Core.Chart.Simulations;
-using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Data;
 using OSPSuite.Core.Domain.Repositories;
 using OSPSuite.Core.Services;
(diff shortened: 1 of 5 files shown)


georgeDaskalakis:
@msevestre @Yuri05 any comments on this one?

Yuri05:
@georgeDaskalakis I don't know this code good enough to judge on the change.
Spontaneously I would ask: why do we need `AddRunResultToChart()` and `UpdateAnalysis();` as separate functions? 
I see the only place where they both are called directly one after another:
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/74733f1e1cfecceba24400708d5fd4e41377e5f4/src/OSPSuite.Presentation/Presenters/SimulationVsObservedDataChartPresenter.cs#L38-L50

georgeDaskalakis:
> @georgeDaskalakis I don't know this code good enough to judge on the change. Spontaneously I would ask: why do we need `AddRunResultToChart()` and `UpdateAnalysis();` as separate functions? I see the only place where they both are called directly one after another:
> 
> https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/74733f1e1cfecceba24400708d5fd4e41377e5f4/src/OSPSuite.Presentation/Presenters/SimulationVsObservedDataChartPresenter.cs#L38-L50

actually @Yuri05 you are right: the AddRunResultToChart() function makes more sense as a private called in UpdateAnalysis(). I corrected that.

Yuri05 (approved):

PR #1824, opened 2022-12-09
Fixes #818 Refactor StartValues vs PathWithValueEntity

### LINKED ISSUE #818: Parameter Identification: When User close and reopen the PI the results section shown the new boundaries and not those used in the PI run
PI is showing a warning in the results section:
image
User set on Parameters tab Min value lower:
image

When User close and reopen the PI the results section shown the new boundaries and not those used in the PI run.
image

rwmcintosh:


rwmcintosh:
I'm also trying to remove all references to IStartValue.StartValue except for serialization...

src/OSPSuite.Core/Domain/Builder/MoleculeStartValue.cs
```diff
@@ -51,7 +51,10 @@ public bool IsPresent
 
       public bool IsEquivalentTo(IMoleculeStartValue moleculeStartValue)
       {
-         var isBaseEquivalent = base.IsEquivalentTo(moleculeStartValue);
+         var isBaseEquivalent = false;
```
msevestre, line 54: I don't quite get this code. Were do we use the isEquivalent
and why do we use suddenly an actual instance of PathAndValueEntity. this feels a bit hacky

src/OSPSuite.Core/Domain/Builder/ParameterStartValue.cs
```diff
@@ -31,7 +31,11 @@ public class ParameterStartValue : StartValueBase, IParameterStartValue
 
       public bool IsEquivalentTo(IParameterStartValue parameterStartValue)
       {
-         var isBaseEquivalent = base.IsEquivalentTo(parameterStartValue);
+         var isBaseEquivalent = false;
+
+         if (parameterStartValue is PathAndValueEntity pathAndValueEntity)
```
msevestre, line 36: same here. Don't understand

src/OSPSuite.Core/Domain/Builder/PathAndValueEntity.cs
```diff
@@ -62,23 +61,13 @@ public Unit DisplayUnit
...
 
-      public ValueOrigin ValueOrigin { get; }
 
       public IObjectPath Path
       {
          get => ContainerPath.Clone<IObjectPath>().AndAdd(Name);
          set => entityFullPathToComponents(value);
       }
 
-      public void UpdateValueOriginFrom(ValueOrigin sourceValueOrigin)
```
msevestre, line 73: Why don't they have value origin?
 Why was this removed?

msevestre (changes requested):
I don't understand those changes
I don't think the cast to an instance of the class is cohesive.
Why is value origin stuff gone?

src/OSPSuite.Core/Domain/Builder/MoleculeStartValue.cs
rwmcintosh: Yeah, the reason for the switch is that I don't want to run this dual inheritance of interfaces and objects. I did not create an IPathAndValueEntity interface at all. But rather than refactor the StartValue classes and tasks to deal with concrete implementations, I drew the line here. All the tasks and whatever are written to use IParameterStartValue.

src/OSPSuite.Core/Domain/Builder/PathAndValueEntity.cs
rwmcintosh: We didn't implement the binder in the view for ExpressionProfile so I didn't think it was necessary. If all PathAndValueEntity should have it then let me know, but then it probably makes sense to expose it in the view.

msevestre (approved):

PR #1831, opened 2022-12-13
Fixes #1357 individual building block for mobi

### LINKED ISSUE #1357: Create Individual Structure (Building block used in MoBi)
OriginData (Should probably be promoted from PKSim to Core)
Cache of Parameters (Path as key, ParameterS)

rwmcintosh:


src/OSPSuite.Core/Domain/Builder/IndividualBuildingBlock.cs
```diff
@@ -0,0 +1,31 @@
+﻿using OSPSuite.Core.Domain.Services;
+
+namespace OSPSuite.Core.Domain.Builder
+{
+   public class IndividualBuildingBlock : PathAndValueEntityBuildingBlockSourcedFromPKSimBuildingBlock<IndividualParameter>
```
rwmcintosh, line 5: I wasn't sure if CalculationMethodCache should also be part of the MoBi Individual. These are the common building block properties besides that one.

msevestre: That's a mouthful class name.

src/OSPSuite.Core/Domain/Builder/PathAndValueEntityBuildingBlockSourcedFromPKSimBuildingBlock.cs
```diff
@@ -0,0 +1,20 @@
+﻿using OSPSuite.Core.Domain.Services;
+
+namespace OSPSuite.Core.Domain.Builder
+{
+   public abstract class PathAndValueEntityBuildingBlockSourcedFromPKSimBuildingBlock<T> : PathAndValueEntityBuildingBlock<T> where T : class, IWithPath, IObjectBase
```
msevestre, line 5: Hum.....don't like that name

src/OSPSuite.Core/Domain/CoreOriginData.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace OSPSuite.Core.Domain
+{
+   public class CoreOriginData : BaseOriginData
```
msevestre, line 3: I don't get the baseOriginData, vs CoreOriginData.
Why do we need this?

rwmcintosh: The only thing that exists in CoreOriginData is Species as a string. The rest is all the same. In Core we want to be able to set and get that string, but not in PKSim (not as a string anyway)

rwmcintosh: generic would probably work here. TSpecies or something like that.

src/OSPSuite.Core/Domain/Builder/IndividualBuildingBlock.cs
rwmcintosh: What about PathAndValueEntityBuildingBlockFromPKSim

rwmcintosh:
Removed OriginData from Core. Now we'll just get an enumerable of properties as string that we can display somehow in the view when it is implemented.

src/OSPSuite.Core/Domain/Builder/IndividualBuildingBlock.cs
```diff
@@ -0,0 +1,11 @@
+﻿using OSPSuite.Utility.Collections;
+
+namespace OSPSuite.Core.Domain.Builder
+{
+   public class IndividualBuildingBlock : PathAndValueEntityBuildingBlockFromPKSim<IndividualParameter>
+   {
+      public Cache<string, string> OriginData { set; get; } = new Cache<string, string>();
```
msevestre, line 7: @rwmcintosh As mentiioned in the chat, I think this is not the right abstraction here

msevestre (approved):

PR #1835, opened 2022-12-15
1836 simulation error not propagated

rwmcintosh:


src/OSPSuite.Core/Domain/Services/SimModelBatch.cs
```diff
@@ -101,7 +111,8 @@ SimulationRunResults createSimulationRunResults(string errorFromException = null
          var warnings = WarningsFrom(_simModelSimulation);
          var error = errorFromException ?? (hasResults ? null : Error.SimulationDidNotProduceResults);
 
-         return new SimulationRunResults(success: error == null, warnings, getResults(), error);
+         var success = error == null;
+         return new SimulationRunResults(success, warnings, success ? getResults() : null, error);
```
rwmcintosh, line 115: We created a secondary error message trying to get results in this case. A warning about time was generated and overwrote the error message from the simulation run

src/OSPSuite.R/Domain/SimulationBatch.cs
```diff
@@ -77,8 +77,10 @@ ISimulationResultsCreator simulationResultsCreator
          _simulationResultsCreator = simulationResultsCreator;
       }
 
-      public void Initialize(IModelCoreSimulation simulation, SimulationBatchOptions simulationBatchOptions)
+      public void Initialize(IModelCoreSimulation simulation, SimulationBatchOptions simulationBatchOptions, SimulationRunOptions simulationRunOptions)
       {
+         _simModelBatch.CheckForNegativeValues = simulationRunOptions?.CheckForNegativeValues ?? true;
```
rwmcintosh, line 82: There was a difference in how this parameter was set for batch vs regular simulation runner. In the batch, there was actually no way to manipulate this option. It was hard-coded.
It seems like you should be able to manipulate it.

rwmcintosh:
@PavelBal FYI

src/OSPSuite.Core/Domain/Services/SimModelBatch.cs
msevestre: Let's just simplify the way we return here. it's hard to understand when what is set

PR #1839, opened 2022-12-16
Fixes #1821 cannot import with commas

### LINKED ISSUE #1821: Cannot import valid CSV file
[test1.csv](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/10188458/test1.csv)
In the csv file attached, column separator is the **blank**, and decimal separator is **comma**.

when importing, I set the separator to blank:
grafik

The import window opens and shows all the column mappings as expected and the data is correctly shown in the preview window (so comma was correctly recognized as decimal separator):

grafik

However, when I click on "Import current sheet" after that: I don't get any graph of the data set which could be imported and the import button is disabled:
grafik

This happens both in PK-Sim and MoBi

rwmcintosh:


msevestre:
Why is this closed?

rwmcintosh:
I made a PR that had this long list of commits that had already been merged, but then deleted it and created a new one.

PR #1842, opened 2022-12-17
Fixes #1841 use named tuple

### LINKED ISSUE #1841: Importer - remove CSVSeparators class and use named tuples

rwmcintosh:
Missed this on the last PR

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Services/ICsvSeparatorSelector.cs
+++ b/src/OSPSuite.Infrastructure.Import/Services/ICsvSeparatorSelector.cs
@@ -1,16 +1,10 @@
 ﻿namespace OSPSuite.Infrastructure.Import.Services
 {
-   public class CSVSeparators
-   {
-      public char ColumnSeparator { get; set; }
-      public char DecimalSeparator { get; set; }
-   }
-
    public interface ICsvSeparatorSelector
    {
       /// <summary>
       ///    Returns the separator to be used for the reading of the .csv file
       /// </summary>
-      CSVSeparators GetCsvSeparator(string fileName);
+      (char? ColumnSeparator, char DecimalSeparator) GetCsvSeparator(string fileName);
    }
 }
--- a/src/OSPSuite.R/Services/DataImporterTask.cs
+++ b/src/OSPSuite.R/Services/DataImporterTask.cs
@@ -105,7 +105,7 @@ namespace OSPSuite.R.Services
          string dataPath,
          char columnSeparator, char decimalSeparator = '.')
       {
-         _csvSeparatorSelector.CsvSeparators = new CSVSeparators { ColumnSeparator = columnSeparator, DecimalSeparator = decimalSeparator };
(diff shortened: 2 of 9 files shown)


src/OSPSuite.Infrastructure.Import/Services/ICsvSeparatorSelector.cs
```diff
@@ -1,16 +1,10 @@
...
-      public char DecimalSeparator { get; set; }
-   }
-
    public interface ICsvSeparatorSelector
    {
       /// <summary>
       ///    Returns the separator to be used for the reading of the .csv file
       /// </summary>
-      CSVSeparators GetCsvSeparator(string fileName);
+      (char? ColumnSeparator, char DecimalSeparator) GetCsvSeparator(string fileName);
```
msevestre, line 8: Naming is off I would say. When you retrieve then, they should be like local variables no?
So columnSeparator and decimalSeparator

src/OSPSuite.R/Services/DataImporterTask.cs
```diff
@@ -105,7 +105,7 @@ public bool IgnoreSheetNamesAtImport
          string dataPath,
          char columnSeparator, char decimalSeparator = '.')
       {
-         _csvSeparatorSelector.CsvSeparators = new CSVSeparators { ColumnSeparator = columnSeparator, DecimalSeparator = decimalSeparator };
+         _csvSeparatorSelector.CsvSeparators = (';', ',');
```
msevestre, line 108: What aren't you using the variables passed as parameters?

src/OSPSuite.UI/Services/CsvSeparatorSelector.cs
```diff
@@ -13,13 +13,13 @@ public CsvSeparatorSelector(IApplicationController applicationController)
...
 
-      public CSVSeparators GetCsvSeparator(string fileName)
+      public (char? ColumnSeparator, char DecimalSeparator) GetCsvSeparator(string fileName)
       {
          using (var csvSeparatorPresenter = _applicationController.Start<ICSVSeparatorSelectorPresenter>())
          {
             csvSeparatorPresenter.SetFileName(fileName);
 
-            return csvSeparatorPresenter.Canceled() ? null : csvSeparatorPresenter.GetCSVSeparators();
+            return csvSeparatorPresenter.Canceled() ? (null, '.') : csvSeparatorPresenter.GetCSVSeparators();
```
msevestre, line 22: This should come from the presenter returning . Shouldn't have to ask cancel or not. Tell don't ask

msevestre (commented):
I thi I there is the one place where you are not using the variables to create the tuple. 
I thought from the code changes earlier that there was only one method with this signature. It's already spreading everywheee so I am not convinced that tuple is actually the way to go by the way so I would just forget about this change all together lol and stick with the class instead. Thoughts?

rwmcintosh:
It's fine with me. I'll just close and delete the branch.

PR #1846, opened 2022-12-17
Fixes #1845 show relevant formula

### LINKED ISSUE #1845: Simulation comparison does not show relevant formula differences
@Yuri05 commented on [Fri Dec 09 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2469)

[NoyesWhitney_TID_Events.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/10195256/NoyesWhitney_TID_Events.zip)
PK-Sim 11.1.94: when comparing 2 simulations shown in the screenshot: no differences are shown
grafik

However there are relevant differences in some formulas (export simulations to the text report or to MoBi to check):

```
Parameter: NEW|Organism|Lumen|Duodenum|Absorption of liquid
	IsFixed: False
	Value: -0.858761430689952 [1/min]
	Formula: alpha*V > 0 ? (k_p*alpha_p*V_p+inflow-k*alpha*V)/(alpha*V) : 0
```

```
Parameter: TID_NoyesWhitney_Events|Organism|Lumen|Duodenum|Absorption of liquid
	IsFixed: False
	Value: -0.858761430689952 [1/min]
	Formula: (k_p*alpha_p*V_p+inflow-k*alpha*V)/(alpha*V)
```

---

@Yuri05 commented on [Fri Dec 09 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2469#issuecomment-1344323381)

even when deactivating the option "only compare properties relevant to simulation results": the formula differences are not shown (only things like name/description/icon/...)
grafik

@msevestre FYI

---

@msevestre commented on [Fri Dec 09 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2469#issuecomment-1344325972)

I love those bugs

---

@msevestre commented on [Fri Dec 09 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2469#issuecomment-1344444878)

@Yuri05 Those parameters are hiddden
If you click "compare hidden entities", you'll see them

image

---

@Yuri05 commented on [Fri Dec 09 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2469#issuecomment-1344492806)

mm, ok. not a bug.
We could think if the properties of a hidden entity, **which are relevant for the simulation results** (like values, formulas, etc.) should be shown even if the option "Compare hidden" is deactivated.

---

@msevestre commented on [Fri Dec 09 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2469#issuecomment-1344514494)

IT's hard to sya. Maybe we compare everything all the time and we remove this property. Hidden are always parameters
so they are always relevant. But if the user do not see the parameter, we did not want to show the parameter in the diff (e.g. we change the formula between two versions). 

How did the user manage to change the formula ?

---

@Yuri05 commented on [Fri Dec 09 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2469#issuecomment-1344605270)

The user didn't :)
The simulation was created with an older PK-Sim version. In the meanwhile, the equation was adjusted in the PK-Sim DB.

When opening the project now, the previously created simulation keeps the old equation, but when recreating it from BBs (or cloning/reconfiguring) - the equation is updated.

---

@msevestre commented on [Fri Dec 09 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2469#issuecomment-1344607175)

Ok. This flag is exactly meant for this.
But maybe we should just not use it anymore if this an issue. Otherwise, this behaves as expected

moBI does not care and shows everything obviously 

---

@Yuri05 commented on [Fri Dec 09 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2469#issuecomment-1344623808)

>But maybe we should just not use it anymore if this an issue. 

maybe... or at least set "show hidden"=true as per default

msevestre:


src/OSPSuite.Core/Comparison/ComparerSettings.cs
```diff
@@ -41,7 +41,7 @@ public class ComparerSettings
       public FormulaComparison FormulaComparison { get; set; }
 
       /// <summary>
-      /// Specifies if hidden entities for PKSim (such as hidden parameters or constant formula) should be part of the comparison. This settings is application dependant (default is <c>false</c>)
+      /// Specifies if hidden entities for PKSim (such as hidden parameters or constant formula) should be part of the comparison. This settings is application dependent (default is <c>false</c>)
       /// </summary>
       public bool CompareHiddenEntities { get; set; }
```
msevestre, line 46: @Yuri05 Interestingly, this property was not SERIALIZED (and therefor always set to the default of the app ever)

msevestre:
@Yuri05 Trivial, merging

PR #1847, opened 2022-12-19
Fixes #1832 individual serialization

### LINKED ISSUE #1832: Serialize Individual to and from PKML and in Mobi project
@rwmcintosh commented on [Tue Dec 06 2022](https://github.com/Open-Systems-Pharmacology/MoBi/issues/820)

rwmcintosh:
You previously mentioned having the wrong abstraction in `IndividualBuildingBlock` so I introduced another class which inherits Cache<string,string>.

Dunno if that's what you meant or not. It prevents me having another class which is basically just name and value.


src/OSPSuite.Core/Domain/Builder/IndividualBuildingBlock.cs
```diff
@@ -2,10 +2,18 @@
...
+   public class OriginDataCache : Cache<string, string>
    {
-      public Cache<string, string> OriginData { set; get; } = new Cache<string, string>();
+      public OriginDataCache() : base(onMissingKey: x => string.Empty)
+      {
+      }
+
       public ValueOrigin ValueOrigin { set; get; }
-      public CalculationMethodCache CalculationMethodCache { set; get; }
+      // public CalculationMethodCache CalculationMethodCache { set; get; }
```
rwmcintosh, line 12: The serializer in PK-Sim uses a CalcuationMethodRepository which is not shared with core. So, is this object shared with MoBi? Or is there just another serializer needed?

src/OSPSuite.Core/Serialization/Xml/OriginDataCacheXmlSerializer.cs
```diff
@@ -0,0 +1,64 @@
...
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Utility.Collections;
+using OSPSuite.Utility.Extensions;
+
+namespace OSPSuite.Core.Serialization.Xml
+{
+   public class OriginDataCacheXmlSerializer : OSPSuiteXmlSerializer<OriginDataCache>
+   {
+      // public class CategoryItemXmlSerializer<TItem> : ObjectBaseXmlSerializer<TItem> where TItem : CategoryItem
```
rwmcintosh, line 11: Depending what I need to do with CalculationMethodCache, I might need this to serialize.

src/OSPSuite.Core/Domain/Builder/IndividualBuildingBlock.cs
msevestre: I don't understand why we need this calculation method repository here.
This info is only used as display purpose only. Just add it to your ORiginDataCache

src/OSPSuite.Core/Serialization/Xml/OriginDataCacheXmlSerializer.cs
msevestre: you don't need this

src/OSPSuite.Core/Serialization/Xml/OriginDataCacheXmlSerializer.cs
```diff
@@ -0,0 +1,64 @@
...
+         base.TypedDeserialize(cache, element, serializationContext);
+         var cacheElement = element.Element(Constants.Serialization.VALUES);
+         var stringStringCacheSerializer = SerializerRepository.SerializerFor<Cache<string, string>>();
+         var values = stringStringCacheSerializer.Deserialize<Cache<string,string>>(cacheElement, serializationContext);
+         values.Keys.Each(x => cache.Add(x, values[x]));
+      }
+
+      protected override XElement TypedSerialize(OriginDataCache cache, SerializationContext serializationContext)
+      {
+         var element = base.TypedSerialize(cache, serializationContext);
```
msevestre, line 56: hum..Why do you have to do all of this? 
I think if the OridinDataCahce is derived from CAche (not sure what you gain here really), then use the Cahce serialzier
If your OriginData HAS a cache, serialize it 
The easiest would have been to dump everything as a string. Since we are creating objects, let's not use string and string but real objects?

OriginData has a _cache of OriginDataItem or whatever
OriginDataItem 
name (key)
displayName (==name if not set)
description = empty bydefault
icon? 

That way you can display your species with an icon, you can display a populationusing the display name coming from PKSim etc.

Either One string, or real structure. A Cache<String, string> (or an object derived from it), does not feel like a structure to use for an API exchange. THis is what I meant with the wrong abstraction and why I always advised against Cache in Public API (unless required as get for serialziation, which here also is more lazyness than anything)

rwmcintosh: Updated with this. I know what you mean about the string dump being the simplest, but I would like to keep the separation of all the details into separate fields until the BB reaches MoBi for viewing.

PR #1849, opened 2022-12-23
1848 vague invalid file error description fixed

### LINKED ISSUE #1848: Importer: Invalid file description vague
When a file contains a duplicate header, the importer does not import it. This is correct. The message given though is very abstract and does not spercify in which sheet the problem exists.

georgeDaskalakis:
closes #1848


georgeDaskalakis:
@msevestre @Yuri05 hmmm actually there is a test failing that fails also in develop (at least for me locally). I can check, but earlist 3 January since I am in vacation till then. Should I convert this PR to a drauft till then?

Yuri05 (approved):

PR #1857, opened 2023-01-05
1851 unnecessary message

georgeDaskalakis:


georgeDaskalakis:
@msevestre @Yuri05 @PavelBal  this fixes the described issue, but would have the following side effect:

When trying to laod an empty file, simply nothing would happen. Right now, when trying to laod an empty file, the user gets notified that the file is not valid. 

For me this is not a grave issue. Of course it is possible to have both possibilities covered, but this would require more extensive code changes, and I would not suggets doing them now, at the end of the code freeze.

Yuri05:
>When trying to laod an empty file, simply nothing would happen. 

Well, I guess when importing an empty file the user would notice it in the import configuration wizard. So fine for me.

georgeDaskalakis:
> > When trying to laod an empty file, simply nothing would happen.
> 
> Well, I guess when importing an empty file the user would notice it in the import configuration wizard. So fine for me.

well not really....the user would just select the empty file, then nothing would happen, meaning the importer dialog would not open at all.

rwmcintosh:
I think we should have a message for the user if there's an empty file. When they dismiss the dialog, then of course they aren't expecting a message, but in this case, they will get frustrated.

Can the file be checked earlier for empty?

PavelBal:
"Nothing happens" is bad.

rwmcintosh (changes requested):
We'll need a specific message for this case

georgeDaskalakis:
@msevestre this is now finalized and ready for review.

rwmcintosh (approved):

PR #1861, opened 2023-01-06
1860 unit test failing locally  fixed

### LINKED ISSUE #1860: Importer: unit test failing locally
I have a local failing unit test for the current version of develop. 

image

After looking into this I have identified that this is being caused by the raplacement of ther decimal separator with '.' for csv files. Since the langugae in my PC is german, the default decimal separator is ',' instead of '.'- which is also the reason why this test is only failing locally for me and not on appveyor.

georgeDaskalakis:
closes #1860

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Core/DataSourceFileReaders/CsvDataSourceFile.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataSourceFileReaders/CsvDataSourceFile.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using OSPSuite.Core.Services;
 using OSPSuite.Infrastructure.Import.Services;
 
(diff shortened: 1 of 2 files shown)


Yuri05 (approved):

src/OSPSuite.Infrastructure.Import/Core/DataSourceFileReaders/CsvDataSourceFile.cs
```diff
@@ -50,7 +51,8 @@ protected override void LoadFromFile(string path)
                while (csv.ReadNextRecord())
                {
                   csv.CopyCurrentRecordTo(currentRow);
-                  var rowList = currentRow.Select(x => x.Replace(csvSeparators.DecimalSeparator, '.')).ToList();
+                  var currentCultureDecimalSeparator = Convert.ToChar(Thread.CurrentThread.CurrentCulture.NumberFormat.NumberDecimalSeparator);
```
rwmcintosh, line 54: This is pretty nice.
I pulled the branch and tried it on my PC and it works (using the starter project). However, it occurs to me that on my machine the new code is identical to what was there since my NumberDecimalSeparator will be '.' anyway.

Could you make sure that you can import this file from on your machine:
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1821


You could probably also skip the Replace if the characters searched and replaced are the same anyway (ie both '.' or both ',')

PR #1865, opened 2023-01-10
Fixes #1864 Scaling for resizable views

### LINKED ISSUE #1864: Scaling for resizable views
For resizable views, the method to calculate scale factor works in steps.

https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/fcd42c611764e4af44f266285175ef66f706fdde/src/OSPSuite.UI/UIConstants.cs#L76-L89

Instead of doing it that way, we should do some interpolation on scale factor (at least for higher DPI values.

rwmcintosh:


rwmcintosh:
4K @ 200%
The fix will work for arbitrarily higher scale factors as well. However, I noticed some other unrelated issues when scaling that much (350%). However, at that scale factor you can't realistically work. 

image
image
image

src/OSPSuite.UI/UIConstants.cs
```diff
@@ -69,19 +69,18 @@ private static int computedExtraRowHeight()
...
          }
 
          private static double createScaleFactor()
          {
             using (var graphics = Graphics.FromHwnd(IntPtr.Zero))
             {
-               if (graphics.DpiX >= 200)
-                  return 2;
-
                if (graphics.DpiX > 120)
```
rwmcintosh, line 80: Maybe we should just always scale if DpiX is > 96. I didn't want to jump the gun though.

Here, my DPI reading was 192 - just below the 200 required to scale by a factor of 2.

I adjusted the threshold to 190 instead of 200 and it looked good.

I guess the DPI will always be the % value of scaling x 96. So it stands to reason we should just scale as % of 96.

msevestre: yeah.. I agree. I would feel it should be always that scaling /96 but cool

PR #1867, opened 2023-01-11
#1866 view and presenter for individual

rwmcintosh:
I needed this interface to create the ValueOriginDTO in MoBi


src/OSPSuite.Core/Domain/Builder/OriginDataItems.cs
```diff
@@ -5,14 +5,19 @@
 
 namespace OSPSuite.Core.Domain.Builder
 {
-   public class OriginDataItems
+   public class OriginDataItems : IWithValueOrigin
```
rwmcintosh, line 8: I needed this interface when I was implementing mappers in MoBi

PR #1868, opened 2023-01-12
1781 chart grouping inconsistent

### LINKED ISSUE #1781: 11.1 Chart Editor default grouping

georgeDaskalakis:
closes #1781 

@msevestre @rwmcintosh @Yuri05 I am updating the xmls to be consistent with what is written in code. I could not find where the "user settings" get applied - I am thinking maybe this comes from the devexpress panel itself somehow...


src/Common/ChartLayouts/Standard View.xml
```diff
@@ -2,7 +2,8 @@
...
     <DataBrowserColumnSettings>
-      <GridColumnSettings columnName="RepositoryName" caption="Repository" visible="1" groupIndex="0" visibleIndex="0" width="0" />
+      <GridColumnSettings columnName="Category" caption="Category" visible="1" groupIndex="1" visibleIndex="0" width="0" />
```
Yuri05, line 5: I wonder if it should be `groupIndex="1"` or `groupIndex="0"` ?
or is it not relevant (DevExpress just sorts by GroupIndex and the absolute value is not of importance)?

georgeDaskalakis: @Yuri05 you are right, actually "groupIndex = 0" makes more sense

src/Common/ChartLayouts/Standard View.xml
```diff
@@ -2,7 +2,8 @@
...
     <DataBrowserColumnSettings>
-      <GridColumnSettings columnName="RepositoryName" caption="Repository" visible="1" groupIndex="0" visibleIndex="0" width="0" />
+      <GridColumnSettings columnName="Category" caption="Category" visible="1" groupIndex="0" visibleIndex="0" width="0" />
+      <GridColumnSettings columnName="RepositoryName" caption="Repository" visible="1" groupIndex="-1" visibleIndex="0" width="0" />
       <GridColumnSettings columnName="Simulation" caption="Simulation" visible="0" groupIndex="-1" visibleIndex="-1" width="0" />
       <GridColumnSettings columnName="TopContainer" caption="Top Container" visible="0" groupIndex="-1" visibleIndex="-1" width="0" />
       <GridColumnSettings columnName="Container" caption="Organ" visible="1" groupIndex="1" visibleIndex="1" width="0" />
```
Yuri05, line 9: please also remove the "Organ" column from Grouping

georgeDaskalakis: done

Yuri05 (approved):

PR #1873, opened 2023-01-22
1872 importer metaData messed up fixed

georgeDaskalakis:
@msevestre @Yuri05 @rwmcintosh this error comes from the fact that when we are setting the mappings for the DataSource, we first set the mapping parameters and then add the grouping ones:

https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/2529fd708ac9237794497df4b49c568c696077d8/src/OSPSuite.Presentation/Presenters/Importer/ImporterPresenter.cs#L212-L239

whereas when creating the ParsedDataSet we would just get all the parameters in a row. That means that when the user would set the grouping and afterwards set a mapping parameter, all the mappings set after the grouping would get assigned wrong values.


src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -218,11 +218,18 @@ public IEnumerable<ParsedDataSet> Parse(DataSheet dataSheet, ColumnInfoCache col
          if (missingColumns.Any())
             throw new MissingColumnException(dataSheet.SheetName, missingColumns);
 
-         var groupingCriteria =
+         //we first isolate the mapping parameters
+         var mappingCriteria =
             Parameters
-               .Where(p => p.IsGroupingCriterion())
+               .Where(p => p.IsGroupingCriterion() && !p.IsAnImplementationOf<GroupByDataFormatParameter>())
```
rwmcintosh, line 224: Can you use the OrderBy to preserve this as a single step (or possibly OrderByDescending).
```
         var mappingCriteria =
            Parameters
               .Where(p => p.IsGroupingCriterion()).OrderBy(p => p.IsAnImplementationOf<GroupByDataFormatParameter>())
               .Select(p => p.ColumnName);

```

Then you wouldn't need to perform the union as step 2. Although, you still need the comment to describe why you are ordering at all.

rwmcintosh (approved):

Yuri05:
`2merge | ¬ 2merge` ?

rwmcintosh:
> `2merge | ¬ 2merge` ?

My comment is only a suggestion to improve the readability for @georgeDaskalakis.

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
georgeDaskalakis: good suggestion @rwmcintosh. I have adjusted the code.

georgeDaskalakis:
@Yuri05 I have implemented Rob's suggestion - since @msevestre has not had any comments I think we can merge yes

PR #1877, opened 2023-01-25
Fixes #1876 Comparison of MSV BBs - list "Molecule name" and "Path" se…

### LINKED ISSUE #1876: Comparison of MSV BBs - list "Molecule name" and "Path" separately
@PavelBal commented on [Tue Dec 13 2022](https://github.com/Open-Systems-Pharmacology/MoBi/issues/835)

Comparison of MSV BBs lists the entries as "Path" and "Molecule" combined in the column "Object"

image

Having separate columns for "Path" and "Molecule" would allow to filter for molecule or a container of interest.

---

@rwmcintosh commented on [Tue Jan 17 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/835#issuecomment-1385440983)

@PavelBal 

You can create some sophisticated filters using the grid control.

Right click on the 'Object' caption

image

Show the filter editor and then you can create filters for strings and you can combine them with logical operations

image

This doesn't allow sorting, but if the goal is to isolate the view to the named molecule than this might do the trick

---

@msevestre commented on [Fri Jan 20 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/835#issuecomment-1398654303)

Path structure (I think)

Simulation. Not required

Then assuming a standard path with at least 2 entries

Last: NamePathElement
First: TopContainerPathElement

IF you have more entries
* For Molecule and Expression
  => One before last is MoleculePathElement
* For other, 
 => One before last is BottomCompartmentPathElement

Then you take whatever is left and put in into ContainerPathElement (and in BottomCompartmentPathElement if you had a molecule)
IF more than one entries, concatenate using the separator used in the PathElementMapper

so Organism|Liver|PeriportalIntracellular|Drug|Concentration

Name: Concentration
Molecule:Drug
Top: Organism
BottomContainer: Intracellular
Container: Liver-Periportal

---

@rwmcintosh commented on [Tue Jan 24 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/835#issuecomment-1402195816)

I think in the case of MSV and Expression the 'Name' is the 'MoleculeName', so all the indexes still line up, but we can have this additional property 'MoleculeName' which is a duplicate of 'Name'

image

rwmcintosh:
…parately

### DIFF
--- a/src/OSPSuite.Presentation/Mappers/DiffItemToDiffItemDTOMapper.cs
+++ b/src/OSPSuite.Presentation/Mappers/DiffItemToDiffItemDTOMapper.cs
@@ -91,22 +91,10 @@ namespace OSPSuite.Presentation.Mappers
       {
          Func<PathElements> pathElementsMapper = () => mapPathElementsAsIEntity(propertyDiffItem.CommonAncestor);
 
-         if (propertyDiffItem.Object1.GetType() == propertyDiffItem.Object2.GetType() && propertyDiffItem.Object1.IsAnImplementationOf<PathAndValueEntity>())
-            pathElementsMapper = () => mapPathElementsAsPathAndValueEntity(propertyDiffItem);
+         if (propertyDiffItem.Object1.IsAnImplementationOf<PathAndValueEntity>())
+            pathElementsMapper = () => _pathAndValueEntityToPathElementsMapper.MapFrom(propertyDiffItem.Object1.DowncastTo<PathAndValueEntity>());
 
          updateDiffItem(propertyDiffItem.FormattedValue1, propertyDiffItem.FormattedValue2, propertyDiffItem.PropertyName, objectNameFrom(propertyDiffItem), itemIsMissing: false, pathElementsMapper);
-      }
-
-      private PathElements mapPathElementsAsPathAndValueEntity(PropertyValueDiffItem propertyDiffItem)
-      {
-         var object1 = propertyDiffItem.Object1 as PathAndValueEntity;
-         var object2 = propertyDiffItem.Object2 as PathAndValueEntity;
-
-         if (object1 == null || object2 == null || !string.Equals(object2.Path.PathAsString, object1.Path.PathAsString))
-            return new PathElements();
-
-
-         return _pathAndValueEntityToPathElementsMapper.MapFrom(object1);
       }
 
       private string objectNameFrom(DiffItem diffItem)
--- a/src/OSPSuite.Presentation/Mappers/DiffItemToDiffItemDTOMapper.cs
+++ b/src/OSPSuite.Presentation/Mappers/DiffItemToDiffItemDTOMapper.cs
@@ -8,6 +8,7 @@ using OSPSuite.Core.Domain.Mappers;
 using OSPSuite.Core.Services;
 using OSPSuite.Presentation.DTO;
 using OSPSuite.Utility;
+using OSPSuite.Utility.Extensions;
 using OSPSuite.Utility.Visitor;
 
 namespace OSPSuite.Presentation.Mappers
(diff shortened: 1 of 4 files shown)


rwmcintosh:
Here's the view with a MSV comparison. There's a bit of duplication because the object name is the molecule.

image

For expression parameters it's a little different because they have a molecule and a distinct name

image

PSV and Individual are very much the same
image

src/OSPSuite.Presentation/Mappers/DiffItemToDiffItemDTOMapper.cs
```diff
@@ -86,7 +89,24 @@ private string presentWithDetails(string presentDetails)
 
       public void Visit(PropertyValueDiffItem propertyDiffItem)
       {
-         updateDiffItem(propertyDiffItem.FormattedValue1, propertyDiffItem.FormattedValue2, propertyDiffItem.PropertyName, objectNameFrom(propertyDiffItem), propertyDiffItem.CommonAncestor, itemIsMissing: false);
+         Func<PathElements> pathElementsMapper = () => mapPathElementsAsIEntity(propertyDiffItem.CommonAncestor);
+
+         if (propertyDiffItem.Object1.GetType() == propertyDiffItem.Object2.GetType() && propertyDiffItem.Object1.IsAnImplementationOf<PathAndValueEntity>())
```
msevestre, line 94: propertyDiffItem.Object1.GetType() == propertyDiffItem.Object2.GetType()
can this ever not be the case?

src/OSPSuite.Presentation/Mappers/DiffItemToDiffItemDTOMapper.cs
```diff
@@ -86,7 +89,24 @@ private string presentWithDetails(string presentDetails)
...
+            pathElementsMapper = () => mapPathElementsAsPathAndValueEntity(propertyDiffItem);
+
+         updateDiffItem(propertyDiffItem.FormattedValue1, propertyDiffItem.FormattedValue2, propertyDiffItem.PropertyName, objectNameFrom(propertyDiffItem), itemIsMissing: false, pathElementsMapper);
+      }
+
+      private PathElements mapPathElementsAsPathAndValueEntity(PropertyValueDiffItem propertyDiffItem)
+      {
+         var object1 = propertyDiffItem.Object1 as PathAndValueEntity;
```
msevestre, line 102: downcast here. This methods required those objects to be PathAndValueEntity so this should never be the case

src/OSPSuite.Presentation/Mappers/DiffItemToDiffItemDTOMapper.cs
```diff
@@ -86,7 +89,24 @@ private string presentWithDetails(string presentDetails)
...
+         var object2 = propertyDiffItem.Object2 as PathAndValueEntity;
+
+         if (object1 == null || object2 == null || !string.Equals(object2.Path.PathAsString, object1.Path.PathAsString))
```
msevestre, line 105: And the second condition? it this check required? Isn't it how we create those objects in the first place. If teels like line 95 could be replaced with ()=> _pathAndValueEntityToPathElementsMapper.MapFrom(object1); and this method gone.

msevestre (approved):
Small minor suggestions. Nothing critical
Code is much easier to understand in the mapper.

src/OSPSuite.Presentation/Mappers/DiffItemToDiffItemDTOMapper.cs
rwmcintosh: I wanted to avoid the implied knowledge of how PathAndValueEntity equivalency is determined, but we do that with Name in other DiffItems all the time, so this is the same, but in this case, with Path.

PR #1879, opened 2023-01-27
Fixes #1878 Rename error description incorrect when renaming building block

### LINKED ISSUE #1878: Rename error description incorrect when renaming building block
The default renaming in MoBi uses this method.

```
      public static string NameAlreadyExistsInContainerType(string name, string containerType)
      {
         return $"'{name}' already exists in {containerType}.";
      }
```

When the containerType is null the string should not have the last word 'in'

Just say "WhateverName already exists" without giving the context in that case

rwmcintosh:


rwmcintosh:
This is a tiny detail in OSPSuite Core while I'm reworking Building Block renaming in MoBi to accommodate the special renaming of Expression Profiles.

PR #1892, opened 2023-02-07
1886 importer unit for fraction not recognized

### LINKED ISSUE #1886: Data importer: Unit for "Fraction" is not recognized
The default unit for the dimension "Fraction" is an empty string, which is not recognized by the importer.

image
[core-importer-fraction-unit.zip](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/files/10579777/core-importer-fraction-unit.zip)

georgeDaskalakis:
closes #1886

### DIFF
--- a/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
+++ b/src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
@@ -263,9 +263,10 @@ namespace OSPSuite.Infrastructure.Import.Core
                   //if unit comes from a column
                   if (column.Key.Column.Dimension == null)
                   {
-                     var firstNonEmptyUnit = column.Value.FirstOrDefault(x => x.Unit != null);
+                     var firstValueWithNonEmptyUnit = column.Value.FirstOrDefault(x => !string.IsNullOrEmpty(x.Unit));
+                     var firstNonEmptyUnit = firstValueWithNonEmptyUnit == null ? "" : firstValueWithNonEmptyUnit.Unit;
 
-                     var dimensionOfFirstUnit = columnInfo.SupportedDimensions.FirstOrDefault(x => x.FindUnit(firstNonEmptyUnit.Unit, ignoreCase: true) != null);
+                     var dimensionOfFirstUnit = columnInfo.SupportedDimensions.FirstOrDefault(x => x.FindUnit(firstNonEmptyUnit, ignoreCase: true) != null);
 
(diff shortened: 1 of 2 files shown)


src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
```diff
@@ -263,16 +263,10 @@ private ParseErrors validateUnitsSupportedAndSameDimension(Cache<string, ColumnI
                   //if unit comes from a column
                   if (column.Key.Column.Dimension == null)
                   {
-                     var firstNonEmptyUnit = column.Value.FirstOrDefault(x => !string.IsNullOrEmpty(x.Unit));
+                     var firstValueWithNonEmptyUnit = column.Value.FirstOrDefault(x => !string.IsNullOrEmpty(x.Unit));
+                     var firstNonEmptyUnit = firstValueWithNonEmptyUnit == null ? "" : firstValueWithNonEmptyUnit.Unit;
```
msevestre, line 267: FYI you can write it as 

`var firstNonEmptyUnit = firstValueWithNonEmptyUnit?.Unit??"";`

tests/OSPSuite.Infrastructure.Tests/Import/DataSourceSpecs.cs
```diff
@@ -164,6 +164,11 @@ protected override void Context()
          sut = new DataSource(_fakedImporter);
          sut.DataSets.Add("sheet1", _fakeDataSet);
       }
+
+      protected string getFirstMeasurementUnit()
+      {
+         return sut.DataSets.KeyValues.First().Value.Data.First().Data.ToList()[1].Value[1].Unit;
```
msevestre, line 170: This structure lol. Value.Data.Data.Value...

msevestre (approved):

src/OSPSuite.Infrastructure.Import/Core/DataSource.cs
georgeDaskalakis: I like the suggestion. Implemented.

tests/OSPSuite.Infrastructure.Tests/Import/DataSourceSpecs.cs
georgeDaskalakis: @msevestre Yes I know - and totally agree. The only reason I would leave this here is that it is only the test and in the actual code it is never called like taht - the unit is deduced further down in the processesing.

PR #1899, opened 2023-02-14
Fixes #603 extension of formula reference

### LINKED ISSUE #603: Extension of formula references concepts
We have a lot of de-facto duplicated formulas in our models just due to the fact that we cannot reference neighbourhoods in a generic way like containers.
Having constructs like `PARENT|NEIGHBOURHOOD_TO.Interstitial` and some more would help.

Another problem is definition of calculation methods formulas for newly added organs in MoBi.
At the moment the problem (or at least one of the problems) is: parameters defined in the molecule properties of the Neighborhood `Organ|Interstitial<=>Organ|Intracellular` refer to other parameters of the same molecule defined in the molecule properties of the neighborhood `Organ|Plasma<=>Organ|Interstitial` (same organ!).

Thus if we find a good way to refer something like `SameOrgan|Plasma<=>SameOrgan|Interstitial / MoleculeProperties` - we could define help parameters in a generic way, which would add them to a tumor automagically.

One difficulty here: somehow full paths (or parts of the full parts) must be then integrated into the description of NEIGHBOURHOOD_TO.

E.g. the following example: 
* parameter `P1 `is located under `Container1|Plasma`
* parameter `P2 `is located in the neighbourhood between `Container1|Plasma` and `Container 4|Container 3|Plasma`
* `P1 `depends on `P2 `
grafik

In such a case `PARENT|NEIGHBOURHOOD_TO.Plasma` would not be sufficient. Instead, something like `PARENT|NEIGHBOURHOOD_TO.Container 4.Container 3.Plasma` should be used.

How should we deal with this?
 * Alternative 1: NEIGHBOURHOOD_TO keyword cann be followed only by a simple container name (e.g. `PARENT|NEIGHBOURHOOD_TO.Plasma`). 
   * Keyword-Resolver must then check if such a neighbourhood is unique.
   * If yes: reference can be resolved
   * If no (like in example above): exception
   * Examples like above must be handled as we do it now: using the name of the neighbourhoud
* Alternative 2: NEIGHBOURHOOD_TO keyword cann be followed by a sequence of container names (e.g. `PARENT|NEIGHBOURHOOD_TO.Container 4.Container 3.Plasma` or `PARENT|NEIGHBOURHOOD_TO.Container 4|Container 3|Plasma`)
* Other alternatives?

msevestre:


src/OSPSuite.Core/Domain/Formulas/FormulaExtensions.cs
```diff
@@ -37,6 +37,15 @@ public static bool IsExplicit(this IFormula formula)
          return formula.IsAnImplementationOf<ExplicitFormula>();
       }
 
+      /// <summary>
+      ///    Returns true if the formula  contains at least one entry using neighborhood reference  otherwise false
+      /// </summary>
+      /// <param name="formula"></param>
+      public static bool IsReferencingNeighborhood(this IFormula formula)
+      {
+         return formula.ObjectPaths.Any(x => x.Contains(ObjectPathKeywords.NBH));
```
Yuri05, line 46: why `Contains `and not `Equals `?

src/OSPSuite.Core/Domain/ObjectPathKeywords.cs
```diff
@@ -61,6 +61,12 @@ public static class ObjectPathKeywords
       /// </summary>
       public static readonly string ALL_FLOATING_MOLECULES = addKeyword("ALL_FLOATING_MOLECULES");
 
+      /// <summary>
+      /// Represents the keywords allowing us to identify a neighborhood between two containers
+      /// For example CONTAINER1|/<NBH/>CONTAINER2/<NBH/> means the neighborhood between CONTAINER1 and CONTAINER2
+      /// </summary>
+      public static readonly string NBH = addKeyword("<NBH>");
```
Yuri05, line 68: Should we name the constant **in the code** NEIGHBORHOOD instead of NBH?
`public static readonly string NEIGHBORHOOD = addKeyword("<NBH>");`

src/OSPSuite.Core/Domain/Services/FormulaTask.cs
```diff
@@ -130,36 +142,85 @@ public bool FormulasAreTheSame(IFormula firstFormula, IFormula secondFormula)
...
+         var lastIndex = pathAsList.FindLastIndex(x => x == NBH);
+
+         //Only one occurrence of the marker, this is not a valid path. We do not change anything as it won't be resolved later on
+         if (firstIndex == lastIndex)
+            return;
+
+         //We retrieve the path to first container, and second container 
+         var pathToFirstContainer = pathAsList.Take(firstIndex).ToList();
+         //+1 and -1 in order to skip the NBH tags
+         var pathToSecondContainer = pathAsList.Skip(firstIndex + 1).Take(lastIndex - firstIndex - 1).ToList();
```
Yuri05, line 170: should we check here, that `pathToSecondContainer `does NOT contain `<NBH>`?
Theoretically we could have (invalid) path like this: `..|<NBH>|..|<NBH>|..|Interstitial|<NBH>|MOLECULE|P1`
Then `pathToSecondContainer `would be `..|<NBH>|..|Interstitial` - which is invalid.

Or will it just automatically crash somewhere later on, so that we don't need this additional check?

src/OSPSuite.Core/Domain/Formulas/FormulaExtensions.cs
msevestre: there are many object paths. (mutliple per formula). We want to get the one that contains NBH
== would be wrong because we are not comparing the same type of objects

src/OSPSuite.Core/Domain/ObjectPathKeywords.cs
msevestre: there is already NEIGHBORHOOD.... so

src/OSPSuite.Core/Domain/Services/FormulaTask.cs
msevestre: yes. the path will be invalid and when we validate later, it will say that this is not a correct path. That's why i don't care if there are 3 or 5 tags. I need TWO

src/OSPSuite.Core/Domain/Formulas/FormulaExtensions.cs
Yuri05: right. I was mistakenly thinking you are checking the entries of a single path, but in fact you are checking multiple paths.

src/OSPSuite.Core/Domain/ObjectPathKeywords.cs
Yuri05: hmm, ok

src/OSPSuite.Core/Domain/Services/FormulaTask.cs
Yuri05: But in case the path has only `one` NBH-keyword, you exit without throwing the exception (lines 163-165).
I don't quite get the difference between these 2 situations (in both cases, the path is invalid)

msevestre: One nbh , I exit , the path will be checked during resolve step
3nbh. I only remove 2. One will be left. And the resolve step will throw

So those two steps are the same really. We rely on the check coming later to warn about a path that's not valid

Yuri05: ok

Yuri05 (approved):

rwmcintosh (approved):

msevestre:
ok I was not expecting it to be merged. I just wanted to have a first review so that we can discuss about it tomorrow
The implementation is not done. The service is not called anywhere and I need some integration tests at the model construction level, I'll reopen and create another PR

PR #1901, opened 2023-02-16
837 sum formula parent

msevestre:
@Yuri05 @rwmcintosh 
This is not working yet. Just want to see @Yuri05 if this is what you had in mind
For some reason, the test fails and I am not sure why


Yuri05:
@msevestre The implementation looks good for me.
Don't quite understand why the test is failing. 
What is the default mode when creating criteria: AND or OR? If it's OR: that is the reason, if it is AND - hmm.

Also that means, that having IN_PARENT condition together with OR mode cannot work - we should thrown an exception then

src/OSPSuite.Core/Domain/Services/FormulaTask.cs
```diff
@@ -211,13 +214,33 @@ public void ExpandDynamicFormulaIn(IContainer container)
...
+      private DescriptorCriteria updateDynamicFormulaCriteria(DynamicFormula formula, IUsingFormula usingFormula)
+      {
+         //we need to replace IN PARENT criteria with actual criteria matching the parent of the usingFormula
+         var criteria = formula.Criteria;
+         var allInParentTags = criteria.Where(x => x.IsAnImplementationOf<InParentCondition>()).ToList();
+         var parent = usingFormula.ParentContainer;
+         if (!allInParentTags.Any() || parent == null)
+            return criteria;
+
+         //we clone the criteria and remove all instances of InParentCondition. Then we add the criteria to the parent specifically
```
Yuri05, line 231: here we should check that criteria mode is AND.
If it is OR: throw the exception

msevestre: ok. Default is AND but yes

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1725,7 +1725,9 @@ public static string TimeArrayValuesDoesNotMatchFirstIndividual(int id, int inde
       public static string UnitIsNotDefinedInDimension(string unit, string dimension) => $"Unit '{unit}' is not defined in dimension '{dimension}'.";
 
       public static string CouldNotFindNeighborhoodBetween(string container1, string container2) => $"Could not find neighborhood between '{container1}' and '{container2}'";
-      
+
+      public const string InParentTagCanOnlyBeUsedWithAndOperator = "IN PARENT tag can only be used with AND operator";
```
msevestre, line 1729: @Yuri05 FYI

src/OSPSuite.Core/Domain/Services/FormulaTask.cs
```diff
@@ -211,13 +214,35 @@ public void ExpandDynamicFormulaIn(IContainer container)
...
+         var modifiedCriteria = criteria.Clone();
+         allInParentTags.Each(x => modifiedCriteria.RemoveByTag<InParentCondition>(x.Tag));
+         if (modifiedCriteria.Operator != CriteriaOperator.And)
+            throw new OSPSuiteException(Error.InParentTagCanOnlyBeUsedWithAndOperator);
```
Yuri05, line 235: actually we could throw the exception earlier, before cloning the criteria.

src/OSPSuite.Core/Domain/Descriptors/NotMatchTagCondition.cs
```diff
@@ -18,6 +18,6 @@ public override bool IsSatisfiedBy(EntityDescriptor entityDescriptor)
          return !entityDescriptor.Tags.Contains(Tag);
       }
 
-      public override IDescriptorCondition CloneCondition() => new NotMatchTagCondition(Tag);
```
msevestre, line 21: @Yuri05 @rwmcintosh I have removed this interface. This may break mobi not sure. I'll update and see
This was definitely overkill on domain side

msevestre (commented):
Ready to review

src/OSPSuite.Core/Domain/Descriptors/NotMatchTagCondition.cs
Yuri05: ok

src/OSPSuite.Core/Domain/Services/FormulaTask.cs
msevestre: done

Yuri05 (approved):

Yuri05:
Looks good! So what was the problem with the test?

msevestre:
@Yuri05 I had added p1 in two different containers (parent, and sub container) so it was loaded twice ....

Yuri05:
>I had added p1 in two different containers (parent, and sub container) so it was loaded twice ....

so p2 in the wrong formula before was not p2 but p1 from another container... lol, I see

PR #1905, opened 2023-02-17
Remove interface for object path making stuff just too complicated fo…

msevestre:
…r nothing

I wanted to remove this interface for a long time so here we go.
I am going to also fix the SS change at the same time that I was going to work on

### DIFF
--- a/src/OSPSuite.Core/Domain/Container.cs
+++ b/src/OSPSuite.Core/Domain/Container.cs
@@ -24,6 +24,11 @@ namespace OSPSuite.Core.Domain
       /// </summary>
       IReadOnlyList<IEntity> Children { get; }
 
+      /// <summary>
+      /// Returns the path to the parent container.
+      /// It should only be set if the container has no parent in the hierarchy. Otherwise, it will be null
+      /// </summary>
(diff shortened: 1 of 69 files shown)


src/OSPSuite.Core/Domain/Container.cs
```diff
@@ -24,6 +24,11 @@ public interface IContainer : IEntity, IEnumerable<IEntity>
       /// </summary>
       IReadOnlyList<IEntity> Children { get; }
 
+      /// <summary>
+      /// Returns the path to the parent container.
+      /// It should only be set if the container has no parent in the hierarchy. Otherwise, it will be null
+      /// </summary>
+      ObjectPath ParentPath { get; set; }
```
msevestre, line 31: @Yuri05 I am not quite sure about the behaviour of this ParentPath. I am not sure I want to synchronize with the ParentContainer of the Entity. I think it should be null all the time to indicate it should be displayed as default in mobi

msevestre (commented):
Please review

msevestre:
@Yuri05 Going to merge so that Ican update mobi. I can revisit those changes later

PR #1906, opened 2023-02-21
Fixes #1903 Resolve commands with dynamic container instead of static

### LINKED ISSUE #1903: Resolve commands with dynamic container instead of static
Context menus rely on ```IoC.Resolve<Command>()```  all over. Instead this should be done with ```container.Resolve<Command>()```

rwmcintosh:


rwmcintosh:
We need to change how the commands are resolved because when we use PK-Sim features from MoBi, the static IoC container is filled with MoBi registrations.

In this case, we really want to use an isolated container with PK-Sim registrations to resolve commands.

PR #1908, opened 2023-02-23
Fixes #1907 Imported expression profile is not populated from the database

### LINKED ISSUE #1907: Imported expression profile is not populated from the PKSim Database
@rwmcintosh commented on [Thu Feb 23 2023](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2546)

@rwmcintosh commented on [Wed Dec 14 2022](https://github.com/Open-Systems-Pharmacology/MoBi/issues/839)

Instead, the empty version of the building block is transferred.

We probably need a button similar to PK-Sim to populate. This would then query PK-Sim to transfer the values

rwmcintosh:
I have this object that I want to use to shuttle the data needed back and forth between MoBi and PK-Sim to allow the database query to run in PK-Sim and get the updated results back to MoBi.

I've called these objects DTO, which is fitting, I think; they are data transfer objects. Usually, we use that terminology for transferring data with the views.

This object is at the core level so you will eventually see PK-Sim and MoBi core tasks with reference to this object.


msevestre (commented):
Not convinced that DTO is the right name. Can't we always used this as a way to imply binding to a view as you said. For all other objects shared between Mobi and PKSim (which is all what ospuite.core) is about, we do not use DTO. So my vote is to drop this ending

rwmcintosh:
Yeah, I had that feeling too. It will cause confusion with that ending.

PR #1910, opened 2023-02-27

### LINKED ISSUE #1909: Gridview dark mode is hard to use

image

image

rwmcintosh:


msevestre:
Why did you close this?

PR #1911, opened 2023-02-28

### LINKED ISSUE #1909: Gridview dark mode is hard to use

rwmcintosh:
part 1 of 3 (MoBi and PKSim will be affected).

My proposal is to remove this DisabledColor from OSPSuite.Core. Within Core itself, it's only used in a couple of places besides ```UxGridView```.

```
PKAnalysisPivotGridControl.cs
WeightedDataRepositoryDataView.cs
```

For those user controls, the usage is adjusting for some other cell states besides readonly/editable. There is some color adjustments going on if cells contents meet certain conditions like null or below LLOQ. For those cases, I actually want to use the skin disabled colors instead of a constant color.

I think the usage in UxGridView itself is not needed. We generally want the skin to take care of drawing the cells.

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -2458,8 +2458,6 @@ namespace OSPSuite.Assets
       /// </summary>
       public static Color ChartDiagramBack = Color.White;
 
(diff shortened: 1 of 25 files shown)


rwmcintosh:
Solving this another way.

PR #1913, opened 2023-02-28
Fixes #1912 Rename class ExpressionParameterUpdate

### LINKED ISSUE #1912: Rename class ExpressionParameterUpdate
Names are not clear
https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/2548#issuecomment-1448088411

rwmcintosh:


src/OSPSuite.Core/Domain/Builder/ExpressionParameter.cs
```diff
@@ -2,5 +2,14 @@
 {
    public class ExpressionParameter : PathAndValueEntity
    {
+      public bool IsGlobalRelativeExpression()
```
msevestre, line 5: ok this is weird lol. A property that looks like a. method that calls an extension method?

msevestre: or is this because it's not a parameter maybe?

src/OSPSuite.Core/Domain/ParameterExtensions.cs
```diff
@@ -46,5 +46,18 @@ public static bool IsConstantParameter(this IParameter parameter)
       {
          return parameter.IsFixedValue || parameter.Formula.IsConstant() || parameter.Formula.IsDistributed();
       }
+
+      public static bool IsGlobalExpression(this IParameter parameter)
```
msevestre, line 50: do we agree that those could be changed to IWIthName and then you do not need the method in the Object Above?

PR #1917, opened 2023-03-08

### LINKED ISSUE #1909: Gridview dark mode is hard to use

rwmcintosh:
Moving Colors.Disabled from OSPSuite.Core.UIConstants to OSPSUite.UI.UIConstants so that we can get disabled colors from the DevExpress skin definition.

Updating uxGridView to use this value as well as the similar TextDisabled to set forecolor.

Now disabled cells in the grid look like disabled textedits. Some skins do not differentiate between disabled and enabled, but the default skin does differentiate.

### DIFF
--- a/src/OSPSuite.UI/UIConstants.cs
+++ b/src/OSPSuite.UI/UIConstants.cs
@@ -1,12 +1,24 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Drawing;
+using DevExpress.Skins;
 using OSPSuite.Assets;
 
 namespace OSPSuite.UI
 {
    public static class UIConstants
    {
+      public static class Colors
+      {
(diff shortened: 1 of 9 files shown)


src/OSPSuite.UI/UIConstants.cs
```diff
@@ -1,12 +1,24 @@
...
+using DevExpress.Skins;
 using OSPSuite.Assets;
 
 namespace OSPSuite.UI
 {
    public static class UIConstants
    {
+      public static class Colors
+      {
+         public static Color Disabled => skinColorFor("ReadOnly");
```
msevestre, line 13: maybe add a comment explaining why you are using a method get and not a property (so that it always gets the latest color from the theme)

msevestre (approved):

PR #1922, opened 2023-03-09
Fixes #1920 create neighborhood builder with path instead of container:

### LINKED ISSUE #1920: Spatial Structure: Allow creation of neighborhood via UI (not drag and drop)
@msevestre commented on [Fri Feb 17 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/898)

It should be possible to create neighborhoods by other means than drag-and-drop. E.g., a UI "Create neighborhood", where the user selects (or types in the full path to) the neighbors. In case of a tumor module, the user would select "PARENT|Tumor|Plasma" as one neighbor, and type in "Organism|VenousBlood|Plasma" as the second neighbor. the "PARENT" keyword is necessary as the path within the module is not complete. "PARENT" will be replaced by the container that is defined in the "Parent container" property.

image

---

@msevestre commented on [Wed Mar 08 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/898#issuecomment-1461205764)

progressing 
/cc @PavelBal @Yuri05 
image

---

@PavelBal commented on [Thu Mar 09 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/898#issuecomment-1461757904)

Looking good! 

But direct input of the path will also be supported, right?

---

@msevestre commented on [Thu Mar 09 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/898#issuecomment-1461947028)

One step at a time. A neighborhood ATM uses real containers so we'll have to break this logic if we allow to select a container that is not defined.
Should we allow any invalid path or only one stating with Parent can be invalid?

---

@PavelBal commented on [Thu Mar 09 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/898#issuecomment-1462176839)

> One step at a time. 

Sure, looks good already :)

> Should we allow any invalid path or only one stating with Parent can be invalid?

Not sure I understand... For the sake of simplicity I would say - allow any path, and think about restrictions later (when testing/working with the new feature, we might get a better feeling of what sanity checks should be performed).

---

@msevestre commented on [Thu Mar 09 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/898#issuecomment-1462179515)

well If I enter a path such as VenBlood|Plasma (typo), I will accept it. So basically, we do not check anything? 
This is going to be all a massive change...damn

---

@PavelBal commented on [Thu Mar 09 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/898#issuecomment-1462186718)

> So basically, we do not check anything?

This would be the expected behavior (not checking), because we could have a module that should create a neighborhood to a container that is maybe still not in the project (e.g., I have an extension module for tumor and it defines a neighborhood with venous blood, and open it in a project where I do not have a PK-Sim project yet).

msevestre:


msevestre:
@rwmcintosh @Yuri05 Tests are still failing but I'd like to put it out there already for review
In short: Neighborhood Builder now uses a path instead of reference to container. This will allow lose references in spatial structure but will come with lots of headache that we did not have before (rename etc..). Anyways, we need this

src/OSPSuite.Core/Domain/Builder/SpatialStructure.cs
```diff
@@ -154,40 +146,6 @@ public override void AcceptVisitor(IVisitor visitor)
...
 
-      /// <summary>
-      ///    Copy the references for first and second Neighbor from <paramref name="sourceNeighborhoodBuilder" /> to
-      ///    <paramref name="targetNeighborhoodBuilder" />.
-      /// </summary>
-      /// <param name="sourceNeighborhoodBuilder">The source neighborhood builder to copy neighbors from.</param>
-      /// <param name="targetNeighborhoodBuilder">The neighborhood builder to copy neighbors to.</param>
-      private void updateNeighborsReferences(INeighborhoodBuilder sourceNeighborhoodBuilder, INeighborhoodBuilder targetNeighborhoodBuilder)
-      {
-         var objectPathFactory = new ObjectPathFactory(new AliasCreator());
```
msevestre, line 165: all this god should be gone normally. Here we were trying to find a container based on the path etc.. well now let's just update the path

msevestre (commented):
I have also renamed the INeighborhood and INeighborhoodBuilder interface

rwmcintosh (approved):
The code looks great to me.

More of an application question rather than code review, but what is the difference between the two neighbors? Why is it important to know first vs. second instead of just a list of two.

```
      public ObjectPath FirstNeighborPath { get; set; }
      public ObjectPath SecondNeighborPath { get; set; }
```

PavelBal:
> More of an application question rather than code review, but what is the difference between the two neighbors? Why is it important to know first vs. second instead of just a list of two.

I would say, if we want to access parameters of the parent container from within the neighborhood, we must somehow differentiate between the two parents.

Yuri05:
Looks good for me

PR #1926, opened 2023-03-18
Fixes #1925

### LINKED ISSUE #1925: ApplicationIcon should convert implicitely to Image
@msevestre commented on [Sat Mar 18 2023](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2563)

@PavelBal commented on [Fri Sep 30 2022](https://github.com/Open-Systems-Pharmacology/MoBi/issues/929)

image

---

@msevestre commented on [Thu Mar 16 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/929#issuecomment-1474900229)

@PavelBal How do you get this bug? Is this here all the time?

---

@PavelBal commented on [Sat Mar 18 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/929#issuecomment-1474900230)

@msevestre I always have this now.

---

@msevestre commented on [Sat Mar 18 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/929#issuecomment-1474900232)

This is a MoBi only issue. I though this was in Core but I was wrong. PKSim does not use it

msevestre:


src/OSPSuite.Assets.Images/ApplicationIcon.cs
```diff
@@ -31,20 +31,14 @@ private static SvgImage bytesToImage(byte[] bytes)
 
       public static implicit operator SvgImage(ApplicationIcon icon) => icon.ToSvgImage();
 
+      public static implicit operator Image(ApplicationIcon icon) => icon.ToImage();
```
msevestre, line 34: @Yuri05 When we moved to svg, we lost the ability to convert implicitly to image. This was a breaking change that I did nto realize. The component that display in the grid only accepts image and hence the bug. Thsi is also a bug in a few other places in PKSim where we are returning an icon instead of an image (pop import from csv file for instance). Hence I decided to fix it here instead of everywhere

PR #1932, opened 2023-03-19
2566 fixed

### LINKED ISSUE #2566: 12.1 Persisting chart error messages
[Crash_Sim.zip](https://github.com/user-attachments/files/21344477/Crash_Sim.zip)
Try to open the simulation **S1** from the attached project: the error below comes multiple times.

If I run the simulation, the same error comes again several times. After that, the simulated curve is plotted.

```
Application:
MoBi® 12.1.0

Object reference not set to an instance of an object.

Stack trace:

at OSPSuite.Presentation.Presenters.Charts.ChartDisplayPresenter.Handle(ChartUpdatedEvent chartUpdatedEvent)
   at OSPSuite.Utility.Events.EventPublisher.<>c__DisplayClass4_1`1.<PublishEvent>b__1()
   at OSPSuite.Utility.Exceptions.ExceptionManagerBase.Execute(Action action)
```

georgeDaskalakis:
closes #2566

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
@@ -131,7 +131,7 @@ namespace OSPSuite.Presentation.Presenters.Importer
       private bool unitNotSet(Column column)
       {
          //when unit is not set and also columnName is not set
-         return (column.Unit.SelectedUnit == null || column.Unit.SelectedUnit == "?" ) 
+         return (column.Unit.SelectedUnit == null || column.Unit.SelectedUnit == "?")
                 && string.IsNullOrEmpty(column.Unit.ColumnName);
       }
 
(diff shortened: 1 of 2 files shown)


msevestre (approved):

src/OSPSuite.Presentation/Presenters/Importer/ColumnMappingPresenter.cs
```diff
@@ -732,6 +733,22 @@ private void invalidateErrorUnit()
...
+      {
+         if (errorColumn.Unit.SelectedUnit != null && errorColumn.Unit.SelectedUnit != UnitDescription.InvalidUnit
+                                                   && errorColumn.ErrorStdDev == null)
+            errorColumn.ErrorStdDev = Constants.STD_DEV_ARITHMETIC;
+      }
+
+      private static void checkErrorUnitAgainstMeasurementUnit(Column errorColumn, Column measurementColumn)
+      {
          if (errorColumn.ErrorStdDev != Constants.STD_DEV_GEOMETRIC &&
              ((errorColumn.Unit?.ColumnName.IsNullOrEmpty() != measurementColumn.Unit?.ColumnName.IsNullOrEmpty()) ||
```
msevestre, line 753: I feel that we are missing a few helper methods here that would make those lines more readable. It feels like we are never sure if the unit is going to bet set, of it's going to be equal to something.

I am glad we have tests to that refactoring should not ber hard

PR #1935, opened 2023-03-20
Fixes #1933 readonly cells are not easily readable

### LINKED ISSUE #1933: After update to using theme in table, content is not readable anymore
@msevestre commented on [Sat Mar 11 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/926)

@rwmcintosh @Yuri05 
image

@rwmcintosh We need a way to specify that readonly cell are NOT using the theme. The changes introduces to satisfy dark mode are creating a lot of UX issues 

Another example
image

So either we can make the text more readable (in this case the gray is fine) or  we do something else. Right now, this is a regression IMO using the default theme

---

@rwmcintosh commented on [Sat Mar 11 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/926#issuecomment-1465058530)

Is there any override colour for the text boxes happening? Eg. parent path? Those are readable. What we have now is just using the theme defaults in the grid so I would expect them to look the same.

---

@msevestre commented on [Sat Mar 11 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/926#issuecomment-1465080133)

No idea. I noticed this issue on all grid however 

---

@rwmcintosh commented on [Sun Mar 12 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/926#issuecomment-1465155441)

It probably just needs a different foreground text color. I picked "Readonly  Text", but if we used the default text it will probably be darker in light theme and lighter in dark theme.

---

@msevestre commented on [Wed Mar 15 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/926#issuecomment-1470905725)

image
lol

---

@Yuri05 commented on [Thu Mar 16 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/926#issuecomment-1471678189)

>lol

agree :)

---

@rwmcintosh commented on [Thu Mar 16 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/926#issuecomment-1471732273)

Really stretching the "Read" of "**Read**Only".

rwmcintosh:


src/OSPSuite.UI/UIConstants.cs
```diff
@@ -12,7 +12,6 @@ public static class Colors
       {
          // These colors change with the skin, so we use a method to read the latest up-to-date value from the skin
          public static Color Disabled => skinColorFor("ReadOnly");
-         public static Color TextDisabled => skinColorFor("DisabledText");
```
rwmcintosh, line 15: Removing this - it makes sense because 'DisabledText' we know reduces the contrast as an indicator. Now we are going to change the background of cells, but leave the foreground to default - that should retain most of the contrast for both dark and light themes.

PR #1936, opened 2023-03-20
Fixes #1934 create simulation configuration

### LINKED ISSUE #1934: Create SimulationConfiguration
as discussed. 
Creating a simulation will now use a module, one individual and one or more expression profile

msevestre:


msevestre:
@Yuri05 @rwmcintosh @georgeDaskalakis 
Tests are not running because we need to add a project conversion step for this. 
Hard to test at the moment so I do not want to spend too much time on implementing testing for this addition. 
BuildConfiguration is left for conversion purpose. We will need to be clever about how to handle it as I do not want to serialize it all the time and delete the node after the fact

src/OSPSuite.Core/Serialization/PKMLVersion.cs
```diff
@@ -6,8 +6,8 @@ public static class PKMLVersion
       public const int V5_4_1 = 5;
       public const int V5_5_1 = 6;
       public const int V5_6_1 = 7;
-      public const int NON_CONVERTABLE_VERSION= V5_6_1;
       public const int V6_0_1 = 8;
```
msevestre, line 9: @Yuri05 I have removed this version from supported version. This is the only place where we were doing something with the BuildCOnfiguraiton in core and this would become a nightmare to try to support it. Adios

src/OSPSuite.Core/Converters/v12/Converter110To120.cs
```diff
@@ -33,7 +35,64 @@ public Converter110To120(IObjectPathFactory objectPathFactory)
...
+         return (PKMLVersion.V12_0, _converted);
+      }
+
+      private void convertSimulation(XElement simulationElement)
+      {
+         //use to list because the node will be deleted
+         simulationElement.Descendants("BuildConfiguration").ToList().Each(convertBuildConfigurationToSimulationConfiguration);
+      }
+
+      private void convertBuildConfigurationToSimulationConfiguration(XElement buildConfigurationElement)
```
msevestre, line 49: @rwmcintosh @georgeDaskalakis 
A little bit nasty but I am converting one xml structure to another BEFORE the serializer is kicking in. While this structure is temporary, we can do sthg like this to ensure our old pkml are valid (we'll have to potentially create another node such as ModuleList or sthg like this but this is it)

msevestre (commented):
Note that I did not remove the buildconfiugration so that MoBi won't completely break. But this should not be used anymore

src/OSPSuite.Core/Converters/v12/Converter110To120.cs
msevestre: Also we do not need to define any constants for those string IMO. This is only use for project conversion and if we change one, the tests will fail

msevestre:
Ok merging
created  #1937 to discuss usage of OverrideFormulaWithValue

PR #1938, opened 2023-03-21
1930 Nunit adapter updated

### LINKED ISSUE #1930: Unit tests not running on latest Visual studio (2023 17.5.2)
Creating this issue since I struggled a bit with this:

After updating to the latest VS (17.5.2, although I think this was not working also a few versions before), the unit tests do not run. This seems to be a known issue  ( [](https://github.com/nunit/nunit3-vs-adapter/issues/987) )  and I got it solved by updating the NUnit3TestAdapter nuget package to 4.4.2

I would suggest pushing this change, if nobody is against it. I don't think that would cause any compatibility issues for anyone.

georgeDaskalakis:
closes #1930


georgeDaskalakis:
Ok, good point - sorry for that then, I will be keeping this in mind more for the future.

PR #1940, opened 2023-03-21
Fixes #1939 logic update do not override formula with constant

### LINKED ISSUE #1939: Logic Update: Do not override formula with constant
@PavelBal @Yuri05 @StephanSchaller
Moving quickly through the module implementation here to get a first simulation and there is a question that came up
At the moment, in MoBi, when a PSV has a value, we are OVERRIDING the formula of the parameter with a constant value

So say that my parameter  Organism|V had a formula P1+P2
and I have a PSV
Path = Organism|V, Value: 55

The resulting parameter in the simulation would lose the formula P1 + P2 and have a constant formula (55)
This becomes a problem now in the individual where we export values for the individual such as Volume. The formula is then removed and becomes constant. (not possible for Distributed ParameteR)

But in general, why would we want to do something like this? 
I Am proposing to change the implementation also to allow a PSV to have a formula AND a value. If both are set, the formula will be set and the value set after (thus winning the battle) and the parameter would look like the default value was changed to the value. This also make sense to me

The only drawback is that it would look like the parameter value differs from default. But this is what we do in PKSim anyways (try to set the volume in individual. The parameter in the simulation appears to have changed)

/cc @rwmcintosh @georgeDaskalakis

msevestre:


src/OSPSuite.Core/Domain/Services/QuantityValuesUpdater.cs
```diff
@@ -82,24 +82,24 @@ private void updateParameterValueFromStartValue(IModel model, PathAndValueEntity
...
+         //Formula is defined, we update in the parameter instance
          if (pathAndValueEntity.Formula != null)
          {
             parameter.Formula = _cloneManagerForModel.Clone(pathAndValueEntity.Formula);
+            
+            //ensures that the parameter is seen as using the formula
+            parameter.IsFixedValue = false;
             _keywordReplacerTask.ReplaceIn(parameter, model.Root);
          }
-         else
```
msevestre, line 90: no else here. we set the FORMULA and the value

src/OSPSuite.Core/Domain/Services/QuantityValuesUpdater.cs
```diff
@@ -82,24 +82,24 @@ private void updateParameterValueFromStartValue(IModel model, PathAndValueEntity
...
+
+         //If the value is defined, this will be used instead of the formula (even if set previously)
+         if (pathAndValueEntity.Value != null)
          {
-            var constantFormula = parameter.Formula as ConstantFormula;
-            var parameterValue = pathAndValueEntity.Value.GetValueOrDefault(double.NaN);
-            if (constantFormula == null)
-            {
-               parameter.Formula = _formulaFactory.ConstantFormula(parameterValue, parameter.Dimension);
```
msevestre, line 96: This is the line I don't get. Why would we replace the formula of a parameter with a constant. I think this does not make sense anymore

Yuri05: I have no idea why this was done. 
Keeping the formula and setting the fixed value makes more sense

msevestre:
MErging then

PR #1941, opened 2023-03-22
Missing properties

msevestre:


src/OSPSuite.Core/Domain/Builder/OriginDataItem.cs
```diff
@@ -1,16 +1,13 @@
 ﻿namespace OSPSuite.Core.Domain.Builder
 {
-   public class OriginDataItem
+   public class OriginDataItem : IWithName, IWithDescription
```
msevestre, line 3: @rwmcintosh Missing. Adding and merging

PR #1947, opened 2023-03-30
Fixes #1946 mobi cannot handler observed data sets with only 1 point

### LINKED ISSUE #1946: MoBi cannot handle observed data sets with only 1 point anymore
@Yuri05 commented on [Fri Feb 24 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/908)

When an observed data set contains only 1 data point and is added to a simulation: simulation cannot be runned anymore due to the error below:
grafik

Example: [Bug_Observed1P.zip](https://github.com/Open-Systems-Pharmacology/MoBi/files/10823577/Bug_Observed1P.zip)
grafik

Seems to be the problem in MoBi only, PK-Sim works.
Also worked in MoBi v 11.0

---

@msevestre commented on [Thu Mar 16 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/908#issuecomment-1473046835)

How can it work in MoBi but not in PKSim??

---

@msevestre commented on [Wed Mar 29 2023](https://github.com/Open-Systems-Pharmacology/MoBi/issues/908#issuecomment-1489516333)

This is because of the residual vs time plot not handling this properly, In MoBi this is done right at the beginning. PKSim on demand. Same problem in PKSim probably

msevestre:


msevestre:
@Yuri05 This fixes the issue in MoBi (and the same issue in PKSim that is when opening a Res vs obs plot)

Yuri05 (approved):

PR #1948, opened 2023-03-30
Fixes #1379

### LINKED ISSUE #1379: Import tool does not read UTF-8 encoded files as UTF-8
When importing data from an UTF-8 encoded data file, PK-Sim seems to read the data as another format (probably LATIN-1). This results in problems when the file contains a "µ" symbol, as this is encoded differently in UTF-8 in comparison to LATIN-1. The result looks like this:
image

msevestre:


src/OSPSuite.Infrastructure.Import/Services/CsvReaderDisposer.cs
```diff
@@ -15,7 +15,7 @@ public class CsvReaderDisposer : IDisposable
       public CsvReaderDisposer(string fileFullPath, char delimiter = ',')
       {
          _fsReader = new FileStream(fileFullPath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
-         _csv = new CsvReader(new StreamReader(_fsReader, Encoding.Default), hasHeaders: true, delimiter: delimiter);
+         _csv = new CsvReader(new StreamReader(_fsReader, Encoding.UTF8), hasHeaders: true, delimiter: delimiter);
```
msevestre, line 18: What the heck is Encoding,Default?

msevestre:
Pushing so that all tests are running. As expected, this was about 2 min works....(unless all tests are breaking..we'll see tomorrow)

Yuri05 (approved):

PR #1952, opened 2023-04-02
Fixes #1949 data importer parsing unit should be case insensitive

### LINKED ISSUE #1949: Data importer - parsing units should be case insensitive
Now "mg/mL" is not recognized, but "mg/ml" is.

msevestre:
@georgeDaskalakis The function hasUnit should not be used as it checked for actual value by keys. The code should always use SupportUnits In my opinion

I found a section that still uses hasUnit. The code is too complicated for me to review or change at this stage so I will ask you to do the modifications if required.
image

I really think the code in those functions (`validateUnitsSupportedAndSameDimension` or `validateErrorAgainstMeasurement` etc) is very complicated. We should extract some of the logic regarding error mapping, finding units etc...we repeat almost the same code throughout the Datasource file and this makes modification much harder than they should


msevestre (commented):
I ran resharper in a bunch of files. This should be done more often

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -76,7 +75,7 @@ private void setDimensionsForMappings(ColumnInfoCache columnInfos)
             else
             {
                var supportedDimensions = concreteColumnInfo.SupportedDimensions;
-               var dimensionForUnit = supportedDimensions.FirstOrDefault(x => x.HasUnit(mappedColumn.Unit.SelectedUnit));
+               var dimensionForUnit = supportedDimensions.FirstOrDefault(x => x.SupportsUnit(mappedColumn.Unit.SelectedUnit, ignoreCase: true));
```
msevestre, line 78: this is the one change that did not impact my failing test but should be done I believe?

src/OSPSuite.Infrastructure.Import/Core/DataFormat/AbstractColumnsDataFormat.cs
```diff
@@ -159,7 +158,12 @@ public UnitDescription ExtractUnitDescriptions(string description, IReadOnlyList
 
       protected string ValidateUnit(string unit, IReadOnlyList<IDimension> supportedDimensions)
       {
-         return supportedDimensions.Any(x => x.HasUnit(unit)) ? unit : UnitDescription.InvalidUnit;
+         var dimensionForUnit = supportedDimensions.FirstOrDefault(x => x.SupportsUnit(unit, ignoreCase: true));
```
msevestre, line 161: this was the one change that was required

tests/OSPSuite.Infrastructure.Tests/Import/DataSourceSpecs.cs
```diff
@@ -24,21 +21,20 @@ public abstract class concern_for_DataSource : ContextSpecification<IDataSource>
       protected IImporter _fakedImporter;
       protected IDataSet _fakeDataSet;
 
-
       protected override void Context()
       {
          _fakedTimeDimension = A.Fake<IDimension>();
-         _fakedConcentrationDimensionMolar = A.Fake<IDimension>(); 
+         _fakedConcentrationDimensionMolar = A.Fake<IDimension>();
```
msevestre, line 27: Why are we faking all the dimensions here? shouldn't we use real dimension instead?

msevestre (commented):
@georgeDaskalakis Please review

tests/OSPSuite.Presentation.Tests/Importer/Core/DataFormat/DataFormat_MetaData_CSpecs.cs
```diff
@@ -1,108 +1,98 @@
...
-            new MetaDataCategory() { Name = "Route" }
+            new MetaDataCategory() {Name = "Organ"},
+            new MetaDataCategory() {Name = "Compartment"},
+            new MetaDataCategory() {Name = "Species"},
+            new MetaDataCategory() {Name = "Dose"},
+            new MetaDataCategory() {Name = "Molecule"},
+            new MetaDataCategory() {Name = "Route"}
          };
 
-         A.CallTo(() => _fakedTimeDimension.HasUnit("min")).Returns(true);
```
msevestre, line 52: @georgeDaskalakis Another example why we should never fake dimension. Suddently other tests are failing because details of implementation has changed. In general, we should only fake presenter, view and services (last only if really required). Domain, as a rule of thumb, never

georgeDaskalakis:
@msevestre I am removing the remaining .HasUnit and am thinking a bit about how to restructure the functions. Should I commit directly to the PR?

msevestre:
@georgeDaskalakis I don't think we need to refactor RIGHT Now. But this needs refactoring for sure
As for the HasUnit. I can't tell. If this comes from the program, then hasUnit is the way to go. But if this comes from PARSING, then hasUnit is wrong

georgeDaskalakis (approved):

PR #1956, opened 2023-04-03
1953 ensures that the simulation configuration supports multiple modules)

msevestre:


msevestre:
@PavelBal @Yuri05 @rwmcintosh 
This is just a first draft trying to have a SImulation configuration using multiple Modules
I have no idea what we want to allow and what not
* CAn we have a transporter defined in Module B transport stuff defined in Module A
* What about Molecules. WHat happens if we have multiple molecule with the same name in different Modules? etc.

src/OSPSuite.Core/Domain/ModelCoreSimulationExtensions.cs
```diff
@@ -39,7 +39,7 @@ private static IReadOnlyList<IContainer> allApplicationsForMolecule(IModelCoreSi
          var allApplications = applicationEventGroup.SelectMany(x => x.GetAllChildren<IContainer>(c => c.ContainerType == ContainerType.Application))
             .ToList();
 
-         return getApplicationsForAppliedAncestorMolecule(simulation.Reactions, moleculeName, allApplications, new List<string>());
+         return getApplicationsForAppliedAncestorMolecule(simulation.Reactions.SelectMany(x=>x), moleculeName, allApplications, new List<string>());
```
Yuri05, line 42: Why do we need this:` .SelectMany(x=>x)` ??

src/OSPSuite.Core/Domain/Module.cs
Yuri05, line 1: (has nothing to do with the change in this PR)
I would find it more intuitive to name the building blocks that contain **multiple** entities **in plural**
- Molecules (or MoleculesBuildingBlock)
- Reactions
- PassiveTransports
- Observers
- EventGroups
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/bb5da48109703ebdc02cfb80fe3c4e09ca9794e9/src/OSPSuite.Core/Domain/Module.cs#L17-L22

PavelBal:
> CAn we have a transporter defined in Module B transport stuff defined in Module A

What do you mean? Molecule A defined in Module A and has an active transport by Transporter B, but Transporter B is defined in Module B?

Then transport is created only when Module A is combined with Module B...

> What about Molecules. WHat happens if we have multiple molecule with the same name in different Modules? etc.

Last module always wins - use the information about the molecule from the last applied module.

src/OSPSuite.Core/Domain/Services/SimulationConfigurationValidator.cs
```diff
@@ -16,20 +16,23 @@ class SimulationConfigurationValidator : ISimulationConfigurationValidator
...
       }
 
+      private void validateModule(Module module, ValidationResult validationResult)
+      {
+         validateBuildingBlockWithFormulaCache(module.Molecule, validationResult);
+         validateBuildingBlockWithFormulaCache(module.Reaction, validationResult);
+         validateBuildingBlockWithFormulaCache(module.SpatialStructure, validationResult);
+         validateBuildingBlockWithFormulaCache(module.PassiveTransport, validationResult);
+         validateBuildingBlockWithFormulaCache(module.Observer, validationResult);
+         validateEventGroupBuildingBlock(module.EventGroup, module?.Molecule, validationResult);
```
Yuri05, line 31: Why "?." in `module?.Molecule` ?

src/OSPSuite.Core/Domain/Services/TransportCreator.cs
```diff
@@ -43,19 +43,23 @@ public void CreateActiveTransport(ModelConfiguration modelConfiguration)
       {
          var (model, simulationConfiguration) = modelConfiguration;
          var allNeighborhoods = model.Neighborhoods.GetChildren<Neighborhood>().ToList();
-         var molecules = simulationConfiguration.Molecules;
-         foreach (var molecule in molecules.AllFloating())
+         //TODO can we have a transport define in one grabbing a molecule from another module?
```
Yuri05, line 46: Hard to imagine... but who knows

rwmcintosh (approved):

src/OSPSuite.Core/Domain/Services/SimulationConfigurationValidator.cs
rwmcintosh: Yeah, that's probably just an oversight. The null-propagation was removed in all the first parameters, but this one had a second parameter with null-prop

src/OSPSuite.Core/Domain/ModelCoreSimulationExtensions.cs
msevestre: Because we go from Enumerabtion of BB to enumeration of reactions. SelectMany flattens it all

src/OSPSuite.Core/Domain/Module.cs
msevestre: @rwmcintosh Can maybe comment on this but I am assuming it's because we simply took the name of the class and removed BuildingBlock (MoleculeBuildingBlock=>Molecule)

src/OSPSuite.Core/Domain/Services/SimulationConfigurationValidator.cs
msevestre: yeah I did not see it when I replaced

msevestre:
@rwmcintosh I don't mind renaming Observer=>Observers etc.. do you feel strongly about this?

src/OSPSuite.Core/Domain/ModelCoreSimulationExtensions.cs
Yuri05: ok

msevestre: not sure we need this. Right now I am combining all over all . This is probably not what we want. Just want to make the tests run again and they break everything again :)

src/OSPSuite.Core/Domain/Module.cs
rwmcintosh: That's exactly right, and they should be changed.

@msevestre if you want to do it now, that would be great. It won't get any easier lol.

rwmcintosh:
> @rwmcintosh I don't mind renaming Observer=>Observers etc.. do you feel strongly about this?

Yes, now is the easiest it's going to get! Thanks.

PR #1964, opened 2023-04-06
1963 only export species independent parameters to spatial structure

msevestre:


msevestre (commented):
Tests are not running yet but here ios the idea
1- so far we were not exporting in individual distributed parameters. Turns out that some are species specific and therefor, our new concepts break this
2-I am exporting distributed parameter as well in our PathAndValueEntity. How do I know? There is a new nullable prop called DsitributionType which will tell us if this is a distributed value or not. In this case, the caller is also responsible to add the required sub parmaetrs (Mean, STD etc.) that are also required for each distribution
3-if a value was set bu the user in PKSim for one of those parameter, a normal parameter is created. (we will lose the distribution here but so what)
4-Tests are still failing


Quesiton: too complicated or required? I thought about it a bit today before implementing and I don't see a way around

src/OSPSuite.Core/Domain/Formulas/DistributionFormula.cs
```diff
@@ -1,73 +1,59 @@
-﻿using System.Collections.Generic;
+﻿using OSPSuite.Core.Maths.Random;
 using OSPSuite.Utility.Extensions;
-using OSPSuite.Core.Maths.Random;
 
 namespace OSPSuite.Core.Domain.Formulas
 {
-   public interface IDistributionFormula : IFormula
```
msevestre, line 7: goodbuy interface with also abtract class

src/OSPSuite.Core/Domain/Formulas/DistributionFormulaFactory.cs
```diff
@@ -21,68 +25,85 @@ public interface IDistributionFormulaFactory
...
 
-      public DistributionFormulaFactory(IObjectPathFactory objectPathFactory,IObjectBaseFactory objectBaseFactory)
+      public DistributionFormulaFactory(IObjectPathFactory objectPathFactory, IObjectBaseFactory objectBaseFactory)
       {
          _objectPathFactory = objectPathFactory;
          _objectBaseFactory = objectBaseFactory;
       }
 
-      public NormalDistributionFormula CreateNormalDistributionFormulaFor(IDistributedParameter distributedParameter, IParameter meanParameter, IParameter deviationParameter)
+      public NormalDistributionFormula CreateNormalDistributionFormulaFor(IDistributedParameter distributedParameter, IParameter meanParameter, IParameter deviationParameter) =>
```
msevestre, line 42: those changes are pretty cool here. I suggest looking at them side by side to understand how i have use a func that returns a func to minimize the parameters that I had to pass around. Functional programming for the win

src/OSPSuite.Core/Domain/Services/ModelConstructor.cs
```diff
@@ -148,16 +148,22 @@ private void removeUndefinedLocalMoleculeParametersIn(IModel model)
...
       private ValidationResult validateModel(ModelConfiguration modelConfiguration)
       {
-         var (_, simulationConfiguration) = modelConfiguration;
          if (!modelConfiguration.ShouldValidate)
             return new ValidationResult();
 
+         var (_, simulationConfiguration) = modelConfiguration;
+
+         //Validation needs to happen at the very end of the process
+         var reactionAndTransportValidation = validate<ValidatorForReactionsAndTransports>(modelConfiguration);
```
msevestre, line 157: @Yuri05 I need to move this at a later stage since some parameters are only created int he last step. So validation of structure cannot be performed at each step as we had it before

tests/OSPSuite.Core.IntegrationTests/Helpers/AssertForSpecs.cs
```diff
@@ -1081,38 +1088,40 @@ public static void AreEqualApplicationBuilder(IApplicationBuilder x1, IApplicati
...
-         if (!AssertBothNotNull(x1, x2)) return;
-         AreEqualObjectPath(x1.Path, x2.Path);
-         AssertAreEqualNullableDouble(x1.Value, x2.Value);
-         AreEqualUnit(x1.DisplayUnit, x2.DisplayUnit);
-         AreEqualDimension(x1.Dimension, x2.Dimension);
-         AreEqualFormula(x1.Formula, x2.Formula);
+         AreEqualPathAndValueEntity(x1, x2);
+      }
+
+      public static void AreEqualIndividualParameter(IndividualParameter x1, IndividualParameter x2)
```
msevestre, line 1121: @rwmcintosh there was not test for serialization here. We need to remember to add tests for ALL our domain

tests/OSPSuite.HelpersForTests/ModelHelperForSpecs.cs
```diff
@@ -103,6 +103,40 @@ private IndividualBuildingBlock getIndividual()
             Value = 10,
             Dimension = amountPerTimeDimension
          });
+
+         individual.Add(new IndividualParameter
+         {
+            //new parameter that does not exist
+            Path = _objectPathFactory.CreateObjectPathFrom(Organism, ArterialBlood, "NEW_PARAM_DISTRIBUTED"),
+            DistributionType = DistributionType.Normal,
```
msevestre, line 111: example here of a dsitrbitued parameter with required sub parameters (MEAN, DEVIATION and PERCENTILE)

src/OSPSuite.Core/Domain/Formulas/DistributionFormulaFactory.cs
```diff
@@ -21,68 +25,85 @@ public interface IDistributionFormulaFactory
...
+         createDistribution<NormalDistributionFormula>(meanParameter.Dimension, configureNormalDistribution(meanParameter.Name, deviationParameter.Name));
+
+      private Action<DistributionFormula, IDimension> configureNormalDistribution(string meanParameterName = null, string deviationParameterName = null) => (distributionFormula, dimension) =>
```
Yuri05, line 45: I think `null `as default value for mean and deviation parameter names does not make sense.
Both parameters must be either mandatory or the default value should be the default names of those parameters 
(which are "Mean" and "Deviation" I think - but needs to be checked)

src/OSPSuite.Core/Domain/Formulas/DistributionFormulaFactory.cs
```diff
@@ -21,68 +25,85 @@ public interface IDistributionFormulaFactory
...
-      }
+         addObjectPathToDistributionFormula(distributionFormula, meanParameterName, MEAN, dimension);
+         addObjectPathToDistributionFormula(distributionFormula, deviationParameterName, DEVIATION, dimension);
+      };
 
-      public DiscreteDistributionFormula CreateDiscreteDistributionFormulaFor(IDistributedParameter distributedParameter, IParameter meanParameter)
+      public DiscreteDistributionFormula CreateDiscreteDistributionFormulaFor(IDistributedParameter distributedParameter, IParameter meanParameter) =>
+         createDistribution<DiscreteDistributionFormula>(meanParameter.Dimension, configureDiscreteDistribution(meanParameter.Name));
+
+      private Action<DistributionFormula, IDimension> configureDiscreteDistribution(string meanParameterName = null) => (distributionFormula, dimension) =>
```
Yuri05, line 54: same here (mandatory or default mean parameter name instead of null)

src/OSPSuite.Core/Domain/Formulas/DistributionFormulaFactory.cs
```diff
@@ -21,68 +25,85 @@ public interface IDistributionFormulaFactory
...
+         addObjectPathToDistributionFormula(distributionFormula, meanParameterName, MEAN, dimension);
+
+      public LogNormalDistributionFormula CreateLogNormalDistributionFormulaFor(IDistributedParameter distributedParameter, IParameter meanParameter, IParameter deviationParameter) =>
+         createDistribution<LogNormalDistributionFormula>(meanParameter.Dimension, configureLogNormalDistributionDistribution(meanParameter.Name, deviationParameter.Name));
+
+      private Action<DistributionFormula, IDimension> configureLogNormalDistributionDistribution(string meanParameterName = null, string deviationParameter = null) => (distributionFormula, dimension) =>
```
Yuri05, line 60: same

src/OSPSuite.Core/Domain/Formulas/DistributionFormulaFactory.cs
```diff
@@ -21,68 +25,85 @@ public interface IDistributionFormulaFactory
...
-      }
+         addObjectPathToDistributionFormula(distributionFormula, meanParameterName, MEAN, dimension);
+         addObjectPathToDistributionFormula(distributionFormula, deviationParameter, GEOMETRIC_DEVIATION, null);
+      };
+
+      public UniformDistributionFormula CreateUniformDistributionFormulaFor(IDistributedParameter distributedParameter, IParameter minParameter, IParameter maxParameter) =>
+         createDistribution<UniformDistributionFormula>(minParameter.Dimension, configureUniformDistributionDistribution(minParameter.Name, maxParameter.Name));
 
-      public LogNormalDistributionFormula CreateLogNormalDistributionFormulaFor(IDistributedParameter distributedParameter, IParameter meanParameter, IParameter deviationParameter)
+      private Action<DistributionFormula, IDimension> configureUniformDistributionDistribution(string minParameterName = null, string maxParameterName = null) => (distributionFormula, dimension) =>
```
Yuri05, line 69: same (null)

src/OSPSuite.Core/Domain/Formulas/DistributionFormulaFactory.cs
```diff
@@ -21,68 +25,85 @@ public interface IDistributionFormulaFactory
...
-         addObjectPathToDistributionFormula(distribution, distributedParameter,maxParameter,Constants.Distribution.MAXIMUM);
-         return distribution;
+         switch (distributionType)
+         {
+            case DistributionType.Normal:
+               return createDistribution<NormalDistributionFormula>(dimension, configureNormalDistribution());
+            case DistributionType.LogNormal:
+               return createDistribution<LogNormalDistributionFormula>(dimension, configureLogNormalDistributionDistribution());
+            case DistributionType.Uniform:
+               return createDistribution<UniformDistributionFormula>(dimension, configureLogNormalDistributionDistribution());
```
Yuri05, line 84: Must be `configureUniformDistributionDistribution`

src/OSPSuite.Core/Domain/Formulas/DistributionFormulaFactory.cs
```diff
@@ -21,68 +25,85 @@ public interface IDistributionFormulaFactory
...
+
+         configure(distribution, dimension);
+         return distribution;
       }
 
-      private void addObjectPathToDistributionFormula(IDistributionFormula distribution,
-                                                      IDistributedParameter distributedParameter,
-                                                      IParameter usedParameter,
-                                                      string alias)
+      private void addObjectPathToDistributionFormula(DistributionFormula distribution, string usedParameter, string alias, IDimension dimension)
```
Yuri05, line 101: looking at this function and all the configureXXXDistribution functions above:
* I would remove the `alias `argument here (and also rename `usedParameter `to `usedParameterName`)
* and change the signature of configurXXX so that the default alias is also the default value of the argument, e.g.
`configureNormalDistribution(string meanParameterName = MEAN, string deviationParameterName = DEVIATION)`

src/OSPSuite.Core/Domain/Formulas/DistributionFormulaFactory.cs
Yuri05: rename **configureLogNormalDistributionDistribution** to **configureLogNormalDistribution** :)

src/OSPSuite.Core/Domain/Formulas/DistributionFormulaFactory.cs
Yuri05: rename **configureUniformDistributionDistribution** to **configureUniformDistribution**

src/OSPSuite.Core/Domain/Formulas/DistributionFormulaFactory.cs
Yuri05: nice :)

src/OSPSuite.Core/Domain/Services/QuantityValuesUpdater.cs
```diff
@@ -64,6 +64,7 @@ private void updateParameterFromExpressionProfiles(ModelConfiguration modelConfi
 
       private void updateParameterFromIndividualValues(ModelConfiguration modelConfiguration)
       {
+         //Order by distribution to ensure that distributed parameter
```
Yuri05, line 67: the comment reminds me of:
<details><summary> 😉 </summary>

</details>

src/OSPSuite.Core/Domain/Services/QuantityValuesUpdater.cs
```diff
@@ -89,11 +90,20 @@ private IParameter getOrAddModelParameter(ModelConfiguration modelConfiguration,
...
+         var displayUnit = pathAndValueEntity.DisplayUnit;
+         var distributionType = pathAndValueEntity.DistributionType;
+         var value = pathAndValueEntity.Value;
 
-         simulationConfiguration.AddBuilderReference(parameter, pathAndValueEntity);
+         //if the distribution is undefined or the value is set, we create a default parameter to ensure that the value will take precedence.
+         //Otherwise, we create a distributed parameter and assume that required sub-parameters will be created as well
+         parameter = distributionType == null || value!=null ? 
+            _parameterFactory.CreateParameter(name, dimension: dimension, displayUnit: displayUnit) : 
+            _parameterFactory.CreateDistributedParameter(name, distributionType.Value, dimension: dimension, displayUnit: displayUnit);
```
Yuri05, line 103: I wouldn't use named arguments if the argumet name is the same as the parameter name (`dimension: dimension` etc.)

src/OSPSuite.Core/Serialization/Xml/ExtendedPropertyXmlSerializer.cs
```diff
@@ -45,7 +45,7 @@ public class BoolExtendedPropertyXmlSerializer : ExtendedPropertyXmlSerializer<b
 
    public class NullableDoubleExtendedPropertyXmlSerializer : ExtendedPropertyXmlSerializer<double?>
    {
-      protected NullableDoubleExtendedPropertyXmlSerializer() : base($"{"NullableDouble"}ExtendedProperty")
```
Yuri05, line 48: `$"{"NullableDouble"}ExtendedProperty"` lol

rwmcintosh (approved):

src/OSPSuite.Core/Domain/Formulas/DistributionFormulaFactory.cs
msevestre: alias need to be MEAN or DEVIATION etc. other wise methods won't work
    protected virtual double Mean(IUsingFormula refObject)
      {
         return GetReferencedEntityByAlias(Constants.Distribution.MEAN, refObject).Value;
      }

src/OSPSuite.Core/Domain/Formulas/DistributionFormulaFactory.cs
msevestre: yep. That's why some tests were failing :)

src/OSPSuite.Core/Domain/Formulas/DistributionFormulaFactory.cs
msevestre: all done

msevestre:
@Yuri05 Ready for review again. Test passing

Yuri05 (approved):

PR #1992, opened 2023-04-26
Fixes #1991 curve binder refresh performance

### LINKED ISSUE #1991: CurveBinder refresh performance

rwmcintosh:
We are frequently looking up the index of basegrid values so that we can use the index to get y value.
In the refresh, we can avoid searching the list over and over by using the index in the first place.

I had a project that was a bit sluggish. Profiling showed that we spent a lot of time finding the index of values in the basegrid

image

### DIFF
--- a/src/OSPSuite.UI/Binders/ArithmeticMeanAreaCurveBinder.cs
+++ b/src/OSPSuite.UI/Binders/ArithmeticMeanAreaCurveBinder.cs
@@ -14,9 +14,9 @@ namespace OSPSuite.UI.Binders
       {
       }
 
-      protected override bool AddRelatedValuesToRow(DataRow row, DataColumn yData, IDimension yDimension, Unit yUnit, double y, float baseValue, int baseIndex)
+      protected override bool AddRelatedValuesToRow(DataRow row, DataColumn yData, IDimension yDimension, Unit yUnit, double y, BaseGrid baseGrid, int baseIndex)
       {
-         var stdDev = yDimension.BaseUnitValueToUnitValue(yUnit, ValueInBaseUnit(Curve.yData, baseValue, baseIndex));
+         var stdDev = yDimension.BaseUnitValueToUnitValue(yUnit, ValueInBaseUnit(Curve.yData, baseGrid, baseIndex));
          if (!IsValidValue(stdDev))
             stdDev = 0;
 
--- a/src/OSPSuite.UI/Binders/ArithmeticMeanAreaCurveBinder.cs
+++ b/src/OSPSuite.UI/Binders/ArithmeticMeanAreaCurveBinder.cs
@@ -14,9 +14,9 @@ namespace OSPSuite.UI.Binders
       {
       }
 
-      protected override bool AddRelatedValuesToRow(DataRow row, DataColumn yData, IDimension yDimension, Unit yUnit, double y, float baseValue)
+      protected override bool AddRelatedValuesToRow(DataRow row, DataColumn yData, IDimension yDimension, Unit yUnit, double y, float baseValue, int baseIndex)
       {
-         var stdDev = yDimension.BaseUnitValueToUnitValue(yUnit, Curve.yData.GetValue(baseValue));
+         var stdDev = yDimension.BaseUnitValueToUnitValue(yUnit, ValueInBaseUnit(Curve.yData, baseValue, baseIndex));
          if (!IsValidValue(stdDev))
             stdDev = 0;
 
--- a/src/OSPSuite.UI/Binders/ArithmeticMeanAreaCurveBinder.cs
+++ b/src/OSPSuite.UI/Binders/ArithmeticMeanAreaCurveBinder.cs
@@ -14,9 +14,9 @@ namespace OSPSuite.UI.Binders
       {
       }
 
-      protected override bool AddRelatedValuesToRow(DataRow row, DataColumn yData, IDimension yDimension, Unit yUnit, double y, int baseIndex)
+      protected override bool AddRelatedValuesToRow(DataRow row, DataColumn yData, IDimension yDimension, Unit yUnit, double y, float baseValue)
       {
-         var stdDev = yDimension.BaseUnitValueToUnitValue(yUnit, Curve.yData.GetValueAt(baseIndex));
+         var stdDev = yDimension.BaseUnitValueToUnitValue(yUnit, Curve.yData.GetValue(baseValue));
          if (!IsValidValue(stdDev))
             stdDev = 0;
 
--- a/src/OSPSuite.UI/Binders/ArithmeticMeanAreaCurveBinder.cs
+++ b/src/OSPSuite.UI/Binders/ArithmeticMeanAreaCurveBinder.cs
@@ -14,9 +14,9 @@ namespace OSPSuite.UI.Binders
       {
       }
 
-      protected override bool AddRelatedValuesToRow(DataRow row, DataColumn yData, IDimension yDimension, Unit yUnit, double y, float baseValue)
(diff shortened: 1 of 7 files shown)


src/OSPSuite.UI/Binders/ArithmeticMeanAreaCurveBinder.cs
```diff
@@ -14,9 +14,9 @@ public ArithmeticMeanAreaCurveBinder(Curve curve, ChartControl chartControl, Cur
       {
       }
 
-      protected override bool AddRelatedValuesToRow(DataRow row, DataColumn yData, IDimension yDimension, Unit yUnit, double y, float baseValue)
+      protected override bool AddRelatedValuesToRow(DataRow row, DataColumn yData, IDimension yDimension, Unit yUnit, double y, int baseIndex)
       {
-         var stdDev = yDimension.BaseUnitValueToUnitValue(yUnit, Curve.yData.GetValue(baseValue));
+         var stdDev = yDimension.BaseUnitValueToUnitValue(yUnit, Curve.yData.GetValueAt(baseIndex));
```
rwmcintosh, line 19: GetValue calls baseGrid.IndexOf underneath.
We can bypass that if we have the index

msevestre: The code is absolutely not equivalent. GetValue interpolates while GetValueAt just gets. 
I am not sure if, in this specific instance, this makes a difference or not. Hard to say. But this is not equivalent code

src/OSPSuite.UI/Binders/CurveBinder.cs
```diff
@@ -309,35 +309,52 @@ private void refreshData()
...
          var xUnit = xDimension.Unit(_xAxis.UnitName);
          var yUnit = yDimension.Unit(_yAxis.UnitName);
-         var xData = Curve.xData;
-         var yData = ActiveYData;
-         var baseGrid = activeBaseGrid(xData, yData);
-
-         // works for different base grids
-         _dataTable.BeginLoadData();
-         foreach (var baseValue in baseGrid.Values)
+         for (var baseIndex = 0; baseIndex < baseGrid.Values.Count; baseIndex++)
```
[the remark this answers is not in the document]
msevestre: See a obove/ This is not equivalent and is wrong when multiple repos are used (with different basegrid unless the basegrids have the same values and length)

msevestre (changes requested):
I think this is wrong. Happy to be corrected
Also when benchmarking, I'd suggest to do a BEFORE and AFTER Time measurement to see the improvements (absolute time)

src/OSPSuite.UI/Binders/CurveBinder.cs
```diff
@@ -309,35 +309,52 @@ private void refreshData()
          if (!dimensionsConsistentToAxisUnits())
             return;
 
+         var baseGrid = activeBaseGrid(Curve.xData, ActiveYData);
```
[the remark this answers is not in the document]
rwmcintosh: Sure, if my understanding of the usage is not quite right, then an interpolation step should not be lost. I thought it looked like in this case of refreshData, there would always be a 1-1 relationship between basegrid and curve data.

If that's wrong, then I'll add an interpolation, but I'll still try to use the index whenever possible without having to look it up.

rwmcintosh:
I did a measurement before after and let the profiler collect all the data
I used the same simulation and configured the chart in the same way
I used the checkboxes in databrowser to:
- I selected all curves
- - turn off all curves
- - turn on all curves
- I selected one curve
- - turn off one curve
- - turn on one curve

The profiler measured total UI freeze time
before: 11.29 seconds
after: 7.45 seconds

per action
| action | before | after |
|--------|--------|--------|
| turn off all curves| none | none|
| turn on all curves | 4.4 | 2.573 |
| turn off one curve | 3.156 | 2.286 |
| turn on one curve | 3.667 | 2.591|

src/OSPSuite.UI/Binders/CurveBinder.cs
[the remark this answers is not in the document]
rwmcintosh: But it should be possible to use the index and value of basegrid together to prevent using IndexOf.
In this loop in refreshData we can compare the index and value of the iterated basegrid to the curve basegrid. If the curve base grid value at a given base grid index is the same, then we can use the value directly. 

That's what GetValue does anyway. If it happens to find an index of a basegrid value, then return the value directly without interpolation.

In this latest change, refreshData checks this directly instead of making any calls to IndexOf.

Check it out.

src/OSPSuite.UI/Binders/CurveBinder.cs
```diff
@@ -356,6 +355,19 @@ private void refreshData()
...
       }
 
+      /// <summary>
+      /// Compares the <paramref name="dataColumn"/> base grid value at the <paramref name="baseIndex"/>. If it
+      /// is the same as <paramref name="baseValue"/> then we can use indexing to retrieve the value, otherwise use interpolation
+      /// </summary>
+      /// <returns>The value of the data column in base units for the given <paramref name="baseValue"/></returns>
+      protected static float ValueInBaseUnit(DataColumn dataColumn, float baseValue, int baseIndex)
+      {
+         if ((dataColumn.BaseGrid.Count > baseIndex) && (Math.Abs(dataColumn.BaseGrid[baseIndex] - baseValue) < float.Epsilon))
```
rwmcintosh, line 365: is there an extension method for float epsilon checks?

msevestre: No idea

msevestre: What is this check doing.? This looks weird. So if i understand is correctly, you are checking that the value at the index is similar to the one that I have in the column basegrid. If yes , then we assume we can use index?

Does it make more sense to check if the basegrid are the same ? If yes, then you can return index. If not , then extrapolation?

rwmcintosh: Yeah, we could do that. What is the prevalence of cases where the base grid has the same values but is not the same object? We would lose the optimization in those cases.
Conversely just refactor a ```hasSameBaseValueAt(int baseIndex)``` method to make it more easy to understand.

rwmcintosh: What if we used the same for related columns? Would those have the same basegrid, or a different basegrid with the same values?

```AddRelatedValuesToRow```

Comment (rwmcintosh) [reply src/OSPSuite.UI/Binders/CurveBinder.cs:365]: So do we need to worry about a CurveBinder having references to different base grids?

**Comment by rwmcintosh** [comment]
Checking for baseGrid object equivalence instead of array index/value equivalence, and applying this optimization to related columns

File: src/OSPSuite.UI/Binders/CurveBinder.cs (hunk 4 above)
Comment (rwmcintosh) [reply src/OSPSuite.UI/Binders/CurveBinder.cs]: Yeah, it's a bit confusing both x and y data have reference to their own basegrid, which is probably the same one.

```
      private BaseGrid activeBaseGrid(DataColumn xData, DataColumn yData)
      {
         if (Curve.InterpolationMode == InterpolationModes.xLinear)
            return xData.BaseGrid;

         if (Curve.InterpolationMode == InterpolationModes.yLinear)
            return yData.BaseGrid;

         throw new ArgumentException("InterpolationMode = " + Curve.InterpolationMode);
      }
```

Anyway, I think we're down to checking if the base grids match, and if so, we return the column data value for that index. If the base grids don't reference match, revert to the previous strategy.

msevestre (approved):
makes sense to me.

PR #2018, opened 2023-05-26
Fixes #2017 Move Initial Conditions relating to an expression profile…

### LINKED ISSUE #2017: Move Initial Conditions relating to an expression profile into the expression profile object
Related to https://github.com/Open-Systems-Pharmacology/MoBi/issues/855

rwmcintosh:
… into the expression profile object

### DIFF
--- a/src/OSPSuite.Core/Domain/Builder/ExpressionProfileBuildingBlock.cs
+++ b/src/OSPSuite.Core/Domain/Builder/ExpressionProfileBuildingBlock.cs
@@ -1,10 +1,14 @@
-﻿using OSPSuite.Core.Domain.Services;
+﻿using System.Collections.Generic;
+using OSPSuite.Core.Domain.Services;
+using OSPSuite.Utility.Extensions;
+using OSPSuite.Utility.Visitor;
 using static OSPSuite.Core.Domain.Constants.ContainerName;
 
 namespace OSPSuite.Core.Domain.Builder
 {
    public class ExpressionProfileBuildingBlock : PathAndValueEntityBuildingBlockFromPKSim<ExpressionParameter>
    {
+      private readonly StartValueCache<InitialCondition> _initialConditions = new StartValueCache<InitialCondition>();
       public override string Icon => Type.IconName;
 
       public virtual string MoleculeName { get; private set; }
@@ -12,7 +16,6 @@ namespace OSPSuite.Core.Domain.Builder
       public string Species { get; private set; }
 
       public ExpressionType Type { set; get; }
-
 
       public virtual string Category { get; private set; }
 
@@ -37,6 +40,20 @@ namespace OSPSuite.Core.Domain.Builder
          }
       }
 
+      public void AddInitialCondition(InitialCondition initialCondition)
+      {
+         _initialConditions.Add(initialCondition);
+         initialCondition.BuildingBlock = this;
+      }
+
+      public override void AcceptVisitor(IVisitor visitor)
+      {
+         base.AcceptVisitor(visitor);
+         _initialConditions.Each(ic => ic.AcceptVisitor(visitor));
+      }
+
+      public IEnumerable<InitialCondition> InitialConditions => _initialConditions;
+
       public override void UpdatePropertiesFrom(IUpdatable source, ICloneManager cloneManager)
       {
          base.UpdatePropertiesFrom(source, cloneManager);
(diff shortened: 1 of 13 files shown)


src/OSPSuite.Core/Domain/Builder/SimulationBuilder.cs
```diff
@@ -92,7 +92,8 @@ private void performMerge()
          _observers.AddRange(allBuilder(x => x.Observers));
          _molecules.AddRange(allBuilder(x => x.Molecules));
          _parameterValues.AddRange(allStartValueBuilder(x => x.SelectedParameterValues));
-         _initialConditions.AddRange(allStartValueBuilder(x => x.SelectedInitialConditions));
+         _initialConditions.AddRange(allStartValueBuilder(x => x.SelectedInitialConditions)
```
rwmcintosh, line 95: Now we need to look in all the expression profiles to build the list of all InitialConditions

src/OSPSuite.Core/Domain/Builder/ExpressionProfileBuildingBlock.cs
```diff
@@ -37,6 +40,29 @@ public override string Name
...
+         initialCondition.BuildingBlock = this;
+      }
+
+      public override void AcceptVisitor(IVisitor visitor)
+      {
+         base.AcceptVisitor(visitor);
+         _initialConditions.Each(ic => ic.AcceptVisitor(visitor));
+      }
+
+      public IEnumerable<InitialCondition> InitialConditions => _initialConditions;
```
msevestre, line 64: IReadonly?

src/OSPSuite.Core/Domain/Builder/ExpressionProfileBuildingBlock.cs
```diff
@@ -48,6 +74,8 @@ public override void UpdatePropertiesFrom(IUpdatable source, ICloneManager clone
          // Name is required because our base objects will the private property
          // But in this case the name is decomposed and stored in 3 other properties
          Name = sourceExpressionProfile.Name;
+
+         sourceExpressionProfile.InitialConditions.Each(initialCondition => AddInitialCondition(cloneManager.Clone(initialCondition)));
```
msevestre, line 78: clear your initial condition probably before doing this?

src/OSPSuite.Core/Domain/Builder/PathAndValueEntityBuildingBlock.cs
```diff
@@ -50,7 +50,7 @@ IEnumerator IEnumerable.GetEnumerator()
       public override void AcceptVisitor(IVisitor visitor)
       {
          base.AcceptVisitor(visitor);
-         _allValues.Each(initialCondition => initialCondition.AcceptVisitor(visitor));
+         _allValues.Each(pathAndValueEntity => pathAndValueEntity.AcceptVisitor(visitor));
```
msevestre, line 53: Personal thoughts for discussion: not sure big long name in lambdas make it easier to read. Above you used ic, ok. x, for the win in this case (or v for value)

src/OSPSuite.Core/Domain/Builder/SimulationBuilder.cs
```diff
@@ -92,7 +92,8 @@ private void performMerge()
...
+            .Concat(_simulationConfiguration.ExpressionProfiles.SelectMany(expressionProfile => expressionProfile.InitialConditions)));
```
msevestre, line 96: consistency. Why not use x here where we have x everywhere. It feels like it;s something difference and it's not

msevestre (changes requested):

src/OSPSuite.Core/Domain/Builder/ExpressionProfileBuildingBlock.cs
rwmcintosh: It's converting a Cache, so I don't know if it's better or not. I would need to call ToList on it for each access. This is templated from the PathAndValueEntityBuildingBlock which does the same.

msevestre: There is another thin. IReadonlyDictionary or sthg?

rwmcintosh: image

Looks like IReadOnlyCollection is the one.

msevestre (approved):

PR #2027, opened 2023-06-02
Fixes #2022 Add common interface to indicate a building block contain…

### LINKED ISSUE #2022: Add common interface to indicate a building block contains initial conditions
This relates to https://github.com/Open-Systems-Pharmacology/MoBi/issues/855

rwmcintosh:
…s initial conditions

### DIFF
--- a/src/OSPSuite.Core/Comparison/ExpressionProfileDiffBuilder.cs
+++ b/src/OSPSuite.Core/Comparison/ExpressionProfileDiffBuilder.cs
@@ -1,5 +1,4 @@
-﻿using OSPSuite.Core.Domain;
-using OSPSuite.Core.Domain.Builder;
+﻿using OSPSuite.Core.Domain.Builder;
 
 namespace OSPSuite.Core.Comparison
 {
@@ -13,7 +12,7 @@ namespace OSPSuite.Core.Comparison
       {
          base.Compare(comparison);
          CompareValues(x => x.Type, x => x.Type, comparison);
-         _enumerableComparer.CompareEnumerables<ExpressionProfileBuildingBlock, InitialCondition, ObjectPath>(comparison, x => x, x => x.Path);
+         _enumerableComparer.CompareEnumerables(comparison, x => x.InitialConditions, x => x.Path);
       }
    }
 
--- a/src/OSPSuite.Core/Comparison/ExpressionProfileDiffBuilder.cs
+++ b/src/OSPSuite.Core/Comparison/ExpressionProfileDiffBuilder.cs
@@ -1,4 +1,5 @@
-﻿using OSPSuite.Core.Domain.Builder;
+﻿using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Builder;
 
 namespace OSPSuite.Core.Comparison
 {
@@ -12,7 +13,7 @@ namespace OSPSuite.Core.Comparison
       {
          base.Compare(comparison);
          CompareValues(x => x.Type, x => x.Type, comparison);
-         _enumerableComparer.CompareEnumerables(comparison, x => x.InitialConditions, x => x.Path);
+         _enumerableComparer.CompareEnumerables<ExpressionProfileBuildingBlock, InitialCondition, ObjectPath>(comparison, x => x, x => x.Path);
       }
    }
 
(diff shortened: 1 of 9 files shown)


src/OSPSuite.Core/Comparison/ExpressionProfileDiffBuilder.cs
```diff
@@ -12,7 +13,7 @@ public override void Compare(IComparison<ExpressionProfileBuildingBlock> compari
       {
          base.Compare(comparison);
          CompareValues(x => x.Type, x => x.Type, comparison);
-         _enumerableComparer.CompareEnumerables(comparison, x => x.InitialConditions, x => x.Path);
```
msevestre, line 15: I am not sure why this was removed? It makes sense for me to have a property InitialCoditions get at the ExpressionProfile level. Using the expression profile as an IEnumerable InitialCondition is not *intuitive*

src/OSPSuite.Core/Domain/Builder/ExpressionProfileBuildingBlock.cs
```diff
@@ -76,7 +59,25 @@ public override void UpdatePropertiesFrom(IUpdatable source, ICloneManager clone
          Name = sourceExpressionProfile.Name;
 
          _initialConditions.Clear();
-         sourceExpressionProfile.InitialConditions.Each(initialCondition => AddInitialCondition(cloneManager.Clone(initialCondition)));
+         sourceExpressionProfile.Each<InitialCondition>(initialCondition => Add(cloneManager.Clone(initialCondition)));
```
msevestre, line 62: Same here. Each with an <> is very unexpected code. I think I want the old code back

src/OSPSuite.Core/Domain/Builder/SimulationBuilder.cs
```diff
@@ -93,7 +93,7 @@ private void performMerge()
          _molecules.AddRange(allBuilder(x => x.Molecules));
          _parameterValues.AddRange(allStartValueBuilder(x => x.SelectedParameterValues));
          _initialConditions.AddRange(allStartValueBuilder(x => x.SelectedInitialConditions)
-            .Concat(_simulationConfiguration.ExpressionProfiles.SelectMany(x => x.InitialConditions)));
+            .Concat(_simulationConfiguration.ExpressionProfiles.SelectMany<ExpressionProfileBuildingBlock, InitialCondition>(x => x)));
```
msevestre, line 96: lol ok this is even more extreme. This code becomes unreadable

src/OSPSuite.Core/Serialization/Xml/BuildingBlockXmlSerializer.cs
```diff
@@ -118,7 +118,7 @@ public override void PerformMapping()
       {
          base.PerformMapping();
          Map(x => x.Type);
-         MapEnumerable(x => x.InitialConditions, x => x.AddInitialCondition);
+         MapEnumerable<InitialCondition>(bb => bb, bb => bb.Add).WithMappingName(Constants.Serialization.INITIAL_CONDITIONS);
```
msevestre, line 121: Please revert this also

msevestre (changes requested):
I do not like this change at all. I think adding the IBuidlignBlcok is one thing that we can agree upon, but removing the accessor to InitialCondition is going to be a harder sell for me and this was *not* what you explained. Each instance where it was removed results in code that is not idiomatic.
I am in favor of adding the InitialConditions accessor back, undoing all changes where it was removed in favor of some weird generics and go from there

rwmcintosh:
> I do not like this change at all. I think adding the IBuidlignBlock is one thing that we can agree upon, but removing the accessor to InitialCondition is going to be a harder sell for me and this was _not_ what you explained. Each instance where it was removed results in code that is not idiomatic. I am in favor of adding the InitialConditions accessor back, undoing all changes where it was removed in favor of some weird generics and go from there

Ok, that makes sense. That means that in places where you have the most derived version ```ExpressionProfileBuildingBlock``` you can use the accessor to get the initial conditions. You are not forced to get the enumerator of the building block, sometimes using weird generics.

msevestre (approved):

PR #2029, opened 2023-06-08
Fixes #2028 "Initial concentration" overwritten to 0 when creating a …

### LINKED ISSUE #2028: "Initial concentration" overwritten to 0 when creating a simulation from module from PK-Sim 11.2
- Create a simulation with PK-Sim V11.2 and export it to PKML
- Load in MoBi OSMOSES
- Create a simulation from the imported building blocks.
- Parameters "Initial concentration" of proteins are set to 0, even though in the PV BB there are properly defined by a formula

image

rwmcintosh:
…simulation from module from PK-Sim 11.2

### DIFF
--- a/src/OSPSuite.Core/Converters/v12/Converter110To120.cs
+++ b/src/OSPSuite.Core/Converters/v12/Converter110To120.cs
@@ -189,6 +189,10 @@ namespace OSPSuite.Core.Converters.v12
 
       public void Visit(ParameterValuesBuildingBlock buildingBlock)
       {
+         // In building blocks prior to v12 (parameter start values building blocks) it was possible to have parameter values
+         // with both formula and value. In those cases, the formula would always be used and the value ignored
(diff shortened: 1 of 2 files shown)


src/OSPSuite.Core/Converters/v12/Converter110To120.cs
```diff
@@ -186,6 +187,15 @@ public void Visit(SimulationConfiguration simulationConfiguration)
          simulationConfiguration.ModuleConfigurations.Each(Visit);
       }
 
+      public void Visit(ParameterValuesBuildingBlock buildingBlock)
+      {
+         buildingBlock.Each(x =>
+         {
+            if (x.Formula != null)
```
msevestre, line 194: I would add a comment here explaing this code .Eg 
"in v11 and prior, both values were saved to pkml bla bla"

msevestre (approved):
looks good to me. Maybe just a quick comment (and even a TEST??) checking that this is working

PR #2031, opened 2023-06-08
Fixes #2030 Weird "differences" when applying an expression profile

### LINKED ISSUE #2030: Weird "differences" when applying an expression profile
-	Re-create a PK-Sim project “Fluvoxamine-Omeprazole-DDI” from snapshot with PK-Sim V 11.2
-	Save project with V11.2
-	Export simulation “omeprazole 40mg po asian EM fluvoxamine 50mg po” to pkml
-	Open Simulation created with V11.2 in MoBi OSMOSES
-	Create new expression profile
o	CYP3A4
o	Database query

- Create a new simulation with expression profile.
- Compare simulations:

image

rwmcintosh:


src/OSPSuite.Core/Comparison/EventAssignmentBuilderDiffBuilder.cs
```diff
@@ -17,7 +17,7 @@ public override void Compare(IComparison<EventAssignmentBuilder> comparison)
       {
          _entityDiffBuilder.Compare(comparison);
          CompareValues(x => x.UseAsValue, x => x.UseAsValue, comparison);
-         CompareValues(x => x.Dimension, x => x.Dimension, comparison);
+         _objectComparer.Compare(comparison.ChildComparison(x => x.Dimension));
          _objectComparer.Compare(comparison.FormulaComparison());
```
msevestre, line 21: I would create an extension methold Like FormulaComparison here
so that you would have 
_objectComparer.Compare(comparison.DimensionComparison());

This would ensure that this is a smaller change, also when implementing in all other apps (MoBi and PKsim)

msevestre (approved):
Looks good. One suggestion to make it even terser

PR #2042, opened 2023-06-22
Fixes #2041 Create initial conditions for ExpressionProfileBuildingBlock

### LINKED ISSUE #2041: Create initial conditions for ExpressionProfileBuildingBlock
relatest to https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2669

rwmcintosh:


src/OSPSuite.Core/Domain/Services/InitialConditionsCreator.cs
```diff
@@ -54,21 +61,31 @@ public InitialConditionsBuildingBlock CreateFrom(SpatialStructure spatialStructu
...
-            var initialCondition = CreateInitialCondition(_objectPathFactory.CreateAbsoluteObjectPath(container), molecule.Name, molecule.Dimension, molecule.DisplayUnit);
-            setInitialCondition(molecule, initialCondition);
-            setInitialConditionFormula(molecule.DefaultStartFormula, initialCondition, initialConditionsBuildingBlock);
-            initialConditionsBuildingBlock.Add(initialCondition);
+            initialConditionsBuildingBlock.Add(createInitialCondition(initialConditionsBuildingBlock, container, molecule));
          }
       }
 
-      private void setInitialConditionFormula(IFormula formula, InitialCondition initialCondition, IBuildingBlock moleculesStartValues)
+      public void CreateForExpressionInContainers(ExpressionProfileBuildingBlock buildingBlock, IReadOnlyList<IContainer> containers, MoleculeBuilder molecule)
```
msevestre, line 80: The signature is a bit weird. It is call create but has a side effects on the given parameter. This methods also acts as a factory (creator factory, same stuff) and typically have no side effects. They just create. 

so I would expect the signature to be more  returning an IreadOnly of InitialCondition and the caller is adding ti t the expression profile. Of course there is the formula cache that I am not sure how to deal with

So maybe the method should be renamed instead= >AddToExpressionProfile so that this is clear that is has side effects. 

I would add a comment explaining also what the contaienrs are because this is not clear

msevestre (changes requested):

PR #2057, opened 2023-07-02
#2019 expanding unit-case-insensitive fix

### LINKED ISSUE #2019: Data importer - parsing units should be case sensitive
A while ago I created an issue that parsing of units during import should be case-insensitive https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1949

Unfortunately, I have a case where this leads to wrong dimension - when trying to import `µm` of the dimension `Lenght`, the importer regodnizes it as `µM` of the dimension `Concentration (molar)`.

georgeDaskalakis:
closes #2019


src/OSPSuite.Core/Domain/UnitSystem/DimensionFactory.cs
```diff
@@ -111,12 +112,7 @@ public IDimension GetOrAddRHSDimensionFor(IDimension dimension)
 
       public IDimension DimensionForUnit(string unitName)
       {
-         var matches = Dimensions.Where(x => x.SupportsUnit(unitName, ignoreCase: true)).ToList();
-         if (!matches.Any())
-            return null;
-
-         //Try to find the first one that matches EXACTLY 
-         return matches.FirstOrDefault(x => x.SupportsUnit(unitName, ignoreCase: false)) ?? matches.First();
+         return Dimensions.DimensionForUnit(unitName);
```
georgeDaskalakis, line 115: @msevestre I moved this code to the newly created extension method to avoid code duplication, since the same function is used in columnInfo.

src/OSPSuite.Infrastructure.Import/Core/ColumnInfo.cs
```diff
@@ -64,9 +65,9 @@ public bool IsMeasurement
          get => !IsBase && !IsAuxiliary;
       }
 
-      public bool SupportsDimension(IDimension dimensionForUnit)
+      public IDimension DimensionForUnit(string unitName)
       {
-         return SupportedDimensions.Contains(dimensionForUnit);
+         return SupportedDimensions.DimensionForUnit(unitName);
```
georgeDaskalakis, line 70: @msevestre I have removed the dependency to the DimensionFactory, since we have already the SupportedDimensions for the specific column, we do not need to search in all the Dimnsions in DimensionFactory for the correct one for this unit, and then check again that it is supported by the ColumnInfo - we just search directly in the SupportedDimensions. I think it is quite better a solution this way.

src/OSPSuite.Core/Extensions/DimensionsListExtensions.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using OSPSuite.Core.Domain.UnitSystem;
+
+namespace OSPSuite.Core.Extensions
+{
+   public static class DimensionsListExtensions
+   {
+      public static IDimension DimensionForUnit(this List<IDimension> dimensionList, string unitName)
```
msevestre, line 9: I dont get why we have two overloads. Cannot you just work with one using Ienumrable. Specifically because the first entry is a query with a tolist.

msevestre (commented):
Looks good to me. One question

msevestre (approved):

src/OSPSuite.Core/Extensions/DimensionsListExtensions.cs
georgeDaskalakis: good point - I was thinking the other way around (had already the function taking a List as an argument and wanted to avoid the ".ToList()" call on the "IEnumerable").  I have removed the redundant overload

PR #2062, opened 2023-07-23
#2043 misleading axes caption fix

### LINKED ISSUE #2043: x and y axis caption of importer graph misleading
image

Can we have the x and y axis show the caption in the table instead of the dimension? (e,g PMA [years] and Ontogey Factor) instead of Dimensionless?

or is there a `good` reason for this mismatch?

georgeDaskalakis:
closes #2043 

note: also contains a small refactoring for the codebase to use XAxis instead of AxisBy(AxisTypes.X) since it is available and the extension of the same logic for the yAxis

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/Charts/SimpleChartPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Charts/SimpleChartPresenter.cs
@@ -211,7 +211,7 @@ namespace OSPSuite.Presentation.Presenters.Charts
       private void setAxesCaptions(DataRepository observedData)
       {
          Chart.XAxis.Caption = observedData.BaseGrid.Name;
-         Chart.YAxis.Caption = observedData.AllButBaseGridAsArray.FirstOrDefault()?.Name;
+         Chart.YAxis.Caption = observedData.ObservationColumns().FirstOrDefault()?.Name;
       }
 
       private void addCurvesToChart(DataRepository observedData)
--- a/src/OSPSuite.Presentation/Presenters/Charts/SimpleChartPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Charts/SimpleChartPresenter.cs
@@ -200,12 +200,18 @@ namespace OSPSuite.Presentation.Presenters.Charts
 
       private Axis getXAxis(CurveChart chart)
       {
-         return chart.AxisBy(AxisTypes.X);
+         return chart.XAxis;
       }
 
       private Axis getYAxis(CurveChart chart)
       {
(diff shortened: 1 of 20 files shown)


src/OSPSuite.Presentation/Presenters/Charts/SimpleChartPresenter.cs
```diff
@@ -200,12 +200,18 @@ private void setScaleInView(CurveChart chart)
...
       {
-         return chart.AxisBy(AxisTypes.Y);
+         return chart.YAxis;
+      }
+
+      private void setAxesCaptions(DataRepository observedData)
+      {
+         Chart.XAxis.Caption = observedData.BaseGrid.Name;
+         Chart.YAxis.Caption = observedData.AllButBaseGridAsArray.FirstOrDefault()?.Name;
       }
```
Yuri05, line 215: Questions:
1) Will the unit be displayed then as part of the column caption?
2) What would happen if `observedData.AllButBaseGridAsArray.FirstOrDefault()` returns not the values column but e.g. the geoSD-column?
3) Don't we already have somewhere a method, which return the first data column which is not the base grid **and not a related column**?

georgeDaskalakis: 1. Yes. The `AxisBinder` takes care of that in `private void setAxisTitle()`, setting the dimension and unit instead of caption and unit if the caption is empty (AxisBinder.cs, line 327):
`_axisView.Title.Text = Constants.NameWithUnitFor(!string.IsNullOrEmpty(Axis.Caption) ? Axis.Caption : Axis.Dimension.DisplayName, Axis.UnitName);
`
also, if the dimension is dimensionless, no unit is diplayed - just the caption.

2. and 3. : Good point. We do not have a function that returns the first one, but we have a function that returns all the observation columns (and with the current state of the importer - and our observed data I would guess, since the same exact error occurs in opening the observed data and will be solved with this PR -  that would be exactly one).  I have replaced `.AllButBaseGridAsArray` with `.ObservationColumns()` in the code.

georgeDaskalakis: @Yuri05 I would say with the above fix the issue with the name possibly coming from the error is fixed. Should I merge?

Yuri05 (approved):

src/OSPSuite.Presentation/Presenters/Charts/SimpleChartPresenter.cs
Yuri05: ok for me :)

PR #2066, opened 2023-07-28
Fixes #2065 Add red/green icons for new building block types

### LINKED ISSUE #2065: Add red/green icons for new building block types
Relates to: https://github.com/Open-Systems-Pharmacology/MoBi/issues/1052

rwmcintosh:


src/OSPSuite.Core/Domain/Builder/EventGroupBuildingBlock.cs
```diff
@@ -6,7 +6,7 @@ public class EventGroupBuildingBlock : BuildingBlock<EventGroupBuilder>
    {
       public EventGroupBuildingBlock()
       {
-         Icon = IconNames.EVENT_GROUP;
+         Icon = IconNames.EVENT;
```
rwmcintosh, line 9: I'm not sure if there's a good reason that this hasn't been changed. I also added a project conversion for this icon change.

Formerly, in MoBi, there was an object called ```BuildingBlockInfo``` that had the icon names. In almost all cases, it was just the icon name from the ```BuildingBlock``` with this one exception where "EventGroup" was not used and instead, "Event" was used.

That meant, that we have these icons ```EventRed``` and ```EventGreen```.

I changed this just to be consistent - we always use the ```BuildingBlock``` icon name for our display icon.

Yuri05 (approved):

PR #2085, opened 2023-08-25
Fixes #2084 Module needs a version

### LINKED ISSUE #2084: Module needs a version
We have to calculate when a module has changed compared to the same module in a simulation

The version calculation will have to take into account the building block versions

rwmcintosh:
I don't think there are any properties of the module itself that we need to consider. ExtendedProperties don't affect the simulation and the property ReadOnly also won't affect the simulation (and I'm not really sure we are going to use it).

### DIFF
--- a/src/OSPSuite.Core/Domain/Module.cs
+++ b/src/OSPSuite.Core/Domain/Module.cs
@@ -54,7 +54,7 @@ namespace OSPSuite.Core.Domain
 
       private static string typedVersionFor(IBuildingBlock x)
       {
-         return x.GetType() + x.Version.ToString();
+         return $"{x.GetType()}{x.Version}";
       }
 
       public Module()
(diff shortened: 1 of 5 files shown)


src/OSPSuite.Core/Domain/Module.cs
```diff
@@ -32,6 +32,31 @@ public class Module : ObjectBase, IEnumerable<IBuildingBlock>
...
+      }
+
+      private string versionCalculation(IEnumerable<IBuildingBlock> buildingBlocks)
+      {
+         return string.Join(string.Empty, buildingBlocks.Select(typedVersionFor).ToArray()).GetHashCode().ToString();
+      }
+
+      private static string typedVersionFor(IBuildingBlock x)
+      {
+         return x.GetType() + x.Version.ToString();
```
msevestre, line 57: I think this deserves a string interpolation instead of +

msevestre (approved):
Looks good to me. One possible update

PR #2089, opened 2023-09-01
Fixes #2088 parameter may not be defined in spatial structure

### LINKED ISSUE #2088: Parameter may not be defined in sptatial strucutre before setting the value
https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2656

Problem is arising when a parameter formula was changed for a distributed parameter. This really happens only when we  have an aging simulation

Say Hematocrit. Mean and deviation as well as main parameter are defined in the Invividual BB. 
WE add the Hematocrit parameter dynamically first. But because we overwrite the formula,. we set the ISDefault to false and in this moment, the percentile parameter in not define (will be added after). We need to be a bit more cool in this instance :)

msevestre:


src/OSPSuite.Core/Domain/DistributedParameter.cs
```diff
@@ -132,9 +132,14 @@ public override bool IsFixedValue
...
             //percentile set first so that correct value is available if a value event is raised
-            PercentileParameter.IsFixedValue = value;
+            //Note: Percentile parameter may be null when constructing the distributed parameter dynamically
+            if (PercentileParameter != null)
+               PercentileParameter.IsFixedValue = value;
+
             base.IsFixedValue = value;
-            OnPropertyChanged(() => Percentile);
+
+            if (PercentileParameter != null)
```
msevestre, line 141: This is required to be able to solve all those small issues in PKSim. Merging. Please review
I don't want to check everywhere. Once one manipulated the parameter, those sub parameters SHOULD BE defined. This is just a weird use case of overriding the formula with a table formula aging

Yuri05 (commented):
looks good

PR #2100, opened 2023-09-08
Fixes #2099 Performance: creating simulation with individual takes 10…

### LINKED ISSUE #2099: Performance: creating simulation with individual takes 10 seconds more than without
- Re-create a PK-Sim project “Fluvoxamine-Omeprazole-DDI” from snapshot with PK-Sim V 11.2
- Save project with V11.2
- Export simulation “omeprazole 40mg po asian EM fluvoxamine 50mg po” to pkml
- Open Simulation created with V11.2 in MoBi OSMOSES
- Create new individual in MoBi
- Create a simulation with the new individual.

Just re-creating the simulation without the individual takes 30 seconds on my machine (already not so fast..), with the individual these are 40 seconds.

rwmcintosh:
… seconds more than without


rwmcintosh:
This method ```ResolveObjectPathsFor``` is only called during model creation, and for some reason during serialization (in 2 spots).
```PKMLPersistor``` and ```SimulationExportCreator``` where the comment is 
```
//resolve to ensure that mapping works as expected
amountKinetic.ResolveObjectPathsFor(process);
```

PR #2107, opened 2023-09-13
Fixes #2098 too many results with wildcards

### LINKED ISSUE #2098: Too many results with wildcards
E.g the search for `getAllMoleculesMatching("Organism|Liver|Periportal|**|CYP3A4*", sim)` returns also 
**Alfentanil-CYP3A4-1st order CL Metabolite** - which is wrong

When I look for `getAllMoleculesMatching("Organism|Liver|Periportal|**|CYP3A4", sim)` 
(removed the wildcard at the end of the search path) - the result is as expected

msevestre:


msevestre:
@Yuri05 So the problem was that the path `Organism|Liver|Periportal|**|CYP3A4*` was with a wildcard BEFORE the CYP". The implementation was wrong in this case as it was trying to match anything before, and the | was optional. That being said, I had to change some of the behavior (See comments)

tests/OSPSuite.R.Tests/Services/ContainerTaskSpecs.cs
```diff
@@ -165,8 +169,8 @@ public class When_resolving_all_parameters_recursively_from_a_container : concer
       [Observation]
       public void should_return_the_matching_parameters()
       {
-         sut.AllParametersMatching(_organism, Constants.WILD_CARD_RECURSIVE).ShouldOnlyContain(_organism.GetAllChildren<IParameter>());
```
msevestre, line 168: This is not ok anymore
** is meant to replace tree structure. so if I Want everything it should be **|* e.g. all possible level and all names). 
I think this was a mistake

tests/OSPSuite.R.Tests/Services/ContainerTaskSpecs.cs
```diff
@@ -129,7 +132,8 @@ public class When_resolving_all_parameters_of_a_container_with_a_container_name_
       public void should_return_the_matching_parameters()
       {
          sut.AllParametersMatching(_organism, pathFrom($"Liv{Constants.WILD_CARD}", INTRACELLULAR, $"Vol{Constants.WILD_CARD}")).ShouldOnlyContain(_volumeLiverCell);
-         sut.AllParametersMatching(_organism, pathFrom($"{Constants.WILD_CARD}Liv", INTRACELLULAR, Constants.Parameters.VOLUME)).ShouldBeEmpty();
+         sut.AllParametersMatching(_organism, pathFrom($"{Constants.WILD_CARD}Liv", INTRACELLULAR, VOLUME)).ShouldBeEmpty();
+         sut.AllParametersMatching(_organism, pathFrom(_liver.Name, Constants.WILD_CARD_RECURSIVE, $"{VOLUME}{Constants.WILD_CARD}")).ShouldOnlyContain(_volumeLiver, _volumeLiverCell);
```
msevestre, line 136: this is your use case here "**|Vol*"

tests/OSPSuite.R.Tests/Services/ContainerTaskSpecs.cs
```diff
@@ -165,8 +169,8 @@ public class When_resolving_all_parameters_recursively_from_a_container : concer
       [Observation]
       public void should_return_the_matching_parameters()
       {
-         sut.AllParametersMatching(_organism, Constants.WILD_CARD_RECURSIVE).ShouldOnlyContain(_organism.GetAllChildren<IParameter>());
-         sut.AllParametersMatching(_liver, Constants.WILD_CARD_RECURSIVE).ShouldOnlyContain(_liver.GetAllChildren<IParameter>());
+         sut.AllParametersMatching(_organism, pathFrom(Constants.WILD_CARD_RECURSIVE, Constants.WILD_CARD)).ShouldOnlyContain(_organism.GetAllChildren<IParameter>());
+         sut.AllParametersMatching(_liver, pathFrom(Constants.WILD_CARD_RECURSIVE, Constants.WILD_CARD)).ShouldOnlyContain(_liver.GetAllChildren<IParameter>());
```
Yuri05, line 173: yep, that's exactly as expected :)

Yuri05 (approved):

PR #2110, opened 2023-09-14
Fixes #2109 Module version calculation is order dependent

### LINKED ISSUE #2109: Module version calculation is order dependent
It depends on the order that building blocks are added to the module, but it should not be.

rwmcintosh:


tests/OSPSuite.Core.Tests/Domain/ModuleSpecs.cs
```diff
@@ -184,7 +184,40 @@ public void should_throw_an_exception()
...
+         sut = new Module
+         {
+            new ParameterValuesBuildingBlock(),
+            new InitialConditionsBuildingBlock(),
+            new EventGroupBuildingBlock()
+         };
+      }
+
+      [Observation]
+      public void the_version_should_not_match()
```
msevestre, line 215: version should match lol

src/OSPSuite.Core/Domain/Module.cs
```diff
@@ -49,7 +49,7 @@ public string VersionWith(ParameterValuesBuildingBlock selectedParameterValues,
 
       private string versionCalculation(IEnumerable<IBuildingBlock> buildingBlocks)
       {
-         return string.Join(string.Empty, buildingBlocks.Select(typedVersionFor).ToArray()).GetHashCode().ToString();
+         return string.Join(string.Empty, buildingBlocks.Select(typedVersionFor).OrderBy(x => x).ToArray()).GetHashCode().ToString();
```
msevestre, line 52: I'd explain this x=>x because it's not quite clear WHY this is done

msevestre (changes requested):

msevestre (approved):
comment not clear when I read it. Let's update to specify that we are ordering by type to ensure that the module with different building block ordering will have the same version. Not the WHAT but the WHY

PR #2127, opened 2023-10-03
Fixes #2126

### LINKED ISSUE #2126: Create simulation with two identical SS leads to crash
Name xxx is not unique in parent container.

Related to global parameters

msevestre:


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -2027,6 +2027,9 @@ public static class Warning
 
       public static string LargeNumberOfOutputPoints(int numberOfPoints) =>
          $"The selected output resolution will generate {numberOfPoints} points and may severely impact the software performance.\nAre you sure you want to run with these setting? If not, consider changing output resolution in simulations settings";
+
+      public static string GlobalParameterIsDefinedInMultipleSpatialStructures(string parameterName) =>
```
msevestre, line 2031: I think having the same parameter defined in two different spatial structure is a bit weird. Alternatively we could just take the Global Molecule parameter from the LAST spatial structure. Thoughts @PavelBal @Yuri05 @rwmcintosh

src/OSPSuite.Core/Domain/Services/MoleculePropertiesContainerTask.cs
```diff
@@ -53,15 +56,24 @@ public IContainer NeighborhoodMoleculeContainerFor(Neighborhood neighborhood, st
...
       {
          var globalMoleculeContainer = addContainerUnder(rootContainer, moleculeBuilder, moleculeBuilder.Name, simulationBuilder)
             .WithContainerType(ContainerType.Molecule);
 
          //Add global molecule dependent parameters
          if (moleculeBuilder.IsFloatingXenobiotic)
          {
-            simulationBuilder.SpatialStructures.Select(x => x.GlobalMoleculeDependentProperties).Each(x => { globalMoleculeContainer.AddChildren(addAllParametersFrom(x, simulationBuilder)); });
+            simulationBuilder.SpatialStructures
+               .Select(x => x.GlobalMoleculeDependentProperties)
```
msevestre, line 68: instead of this select and then SelctMany, I could just select the last `GlobalMoleculeDependentProperties` which would be similar to the what we do with other containers. But I am not sure if this would create a problem ??

src/OSPSuite.Assets/UIConstants.cs
Yuri05: Actually I still don't quite understand how it comes to this situation.
I thought that if we have >1 spatial structure with the global MoleculeProperties container - than the global MoleculeProperties container from the **last** spatial structure would completely **replace** the one from the previous spatial structure(s)?

Yuri05: >I feel that we should just use the last global container 

I would say: yes, this sounds more reasonable for me.

msevestre: see new implementation

src/OSPSuite.Core/Domain/Services/MoleculePropertiesContainerTask.cs
```diff
@@ -58,11 +61,11 @@ public IContainer CreateGlobalMoleculeContainerFor(IContainer rootContainer, Mol
          var globalMoleculeContainer = addContainerUnder(rootContainer, moleculeBuilder, moleculeBuilder.Name, simulationBuilder)
             .WithContainerType(ContainerType.Molecule);
 
+         var lastSpatialStructureGlobalMoleculeContainer = simulationBuilder.SpatialStructures.LastOrDefault()?.GlobalMoleculeDependentProperties;
```
msevestre, line 64: I am not sure if we always will have one container or not. The code before assumed yes but this may have been wrong

rwmcintosh (approved):

PR #2134, opened 2023-10-04
Fixes #2133: Ensure that we filter out null entry when adding to the …

### LINKED ISSUE #2133: V12: Error when calling popup menu for >2 simulations
see https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2792

msevestre:
…view

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/ContextMenus/ContextMenu.cs
+++ b/src/OSPSuite.Presentation/Presenters/ContextMenus/ContextMenu.cs
@@ -1,11 +1,11 @@
 ﻿using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
-using OSPSuite.Utility.Container;
-using OSPSuite.Utility.Extensions;
 using OSPSuite.Presentation.MenuAndBars;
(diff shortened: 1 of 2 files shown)


src/OSPSuite.Presentation/Presenters/ContextMenus/ContextMenu.cs
```diff
@@ -62,7 +60,7 @@ public abstract class ContextMenu<T> : ContextMenu
       protected ContextMenu(T objectRequestingContextMenu, IContainer container) : base(container)
       {
          _container = container;
-         AllMenuItemsFor(objectRequestingContextMenu).Each(_view.AddMenuItem);
+         AllMenuItemsFor(objectRequestingContextMenu).Where(x => x != null).Each(_view.AddMenuItem);
```
msevestre, line 63: @Yuri05 SOmetimes derives class return some items that are null if the entry should not be displayed. It's better to filter them here

msevestre (commented):
Merging

Yuri05 (commented):
approved

PR #2137, opened 2023-10-04
Fixed add extension method

msevestre:


src/OSPSuite.Core/Domain/ContainerExtensions.cs
```diff
@@ -184,5 +184,13 @@ public static void AddChildren(this IContainer container, IEnumerable<IEntity> c
       {
          return container?.GetSingleChild<TChild>(x => true);
       }
+
+
+      public static IEnumerable<(IUsingFormula usingFormula, FormulaUsablePath path)> GetPathsReferencing(this IContainer container, Func<FormulaUsablePath, bool> predicate)
```
msevestre, line 189: @Yuri05 Just extracting this guy because I need it in MoBi

PR #2142, opened 2023-10-13
Fixes #2141 Extract mapper to create parameters from individual parameters

### LINKED ISSUE #2141: Extract mapper to create parameters from individual parameters

rwmcintosh:
Just pulling some code into a task that can be reused

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,8 +1,8 @@
+Fixes #
+
 # Description
 
 Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.
-
-Fixes #
 
 ## Type of change
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,39 @@
+# Description
+
+Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.
+
+Fixes #
+
+## Type of change
+
+Please mark relevant options with an `x` in the brackets.
+
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- [ ] This change requires a documentation update
+- [ ] Algorithm update - updates algorithm documentation/questions/answers etc.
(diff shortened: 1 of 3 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,39 @@
...
+- [ ] Unit tests
+- [ ] Manual tests 
+- [ ] No tests required
+
+# Reviewer checklist
+
+Mark everything that needs to be checked before merging the PR.
+
+- [ ] Check if the code is well documented
+- [ ] Check if the behavior is what is expected
```
rwmcintosh, line 32: I think this one is ok if you can check it without having to pull the branch and test. That shouldn't be required of the reviewer don't you think?

I still think that it's good to check in PR through screenshots and description though.

rwmcintosh:
Sneaking in that PR template here. If we can agree on the content, then I'll add it to the other main repos. Maybe all the C# ones?

.github/PULL_REQUEST_TEMPLATE.md
msevestre: Yeah. Let's do this and see how we like it for the next few days and promotr

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,39 @@
+# Description
+
+Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.
+
+Fixes #
```
msevestre, line 5: Can you move this to the first line ? It's hard to find otherwise

src/OSPSuite.Core/Domain/Mappers/IndividualParameterToParameterMapper.cs
```diff
@@ -0,0 +1,35 @@
...
+      public IParameter MapFrom(IndividualParameter individualParameter)
+      {
+         var name = individualParameter.Name;
+         var dimension = individualParameter.Dimension;
+         var displayUnit = individualParameter.DisplayUnit;
+         var distributionType = individualParameter.DistributionType;
+
+         //if the distribution is undefined or the value is set, we create a default parameter to ensure that the value will take precedence.
+         //Otherwise, we create a distributed parameter and assume that required sub-parameters will be created as well
+         var distributedParameter = distributionType == null || individualParameter.Value != null ?
```
msevestre, line 28: Naming is off here. It's not necessarily a distributed parameter

msevestre (changes requested):

PR #2212, opened 2024-02-28
appveyor update

rwmcintosh:


appveyor.yml
```diff
@@ -4,15 +4,15 @@ image: Visual Studio 2019
...
+- ps: Update-AppveyorBuild -Version "$($env:ospsuite_nightly_version).$($env:appveyor_build_version)"
 
 dotnet_csproj:
   patch: true
   file: '**\*.csproj'
-  version: '11.3.$(build_number)'
-  assembly_version: '11.3.$(build_number)'
-  file_version: '11.3.$(build_number)'
-  informational_version: '11.3.$(build_number)'
+  version: '$(ospsuite_nightly_version).$(build_number)'
```
rwmcintosh, line 12: Just one PR before I duplicate this all over - is this ```$(ospsuite_nightly_version)``` the same as the one above that starts with ```$env::``` or is there some other variable with this value?

msevestre: @rwmcintosh It depends on where the `context` is used
Like 7 is executed in Powershell (ps)
the rest is executed in ruby? or yml environment

msevestre: so it's the same. Just diferent syntax

msevestre (approved):

PR #2215, opened 2024-03-19
2150 removing observed data entry from simulation should not show a warning

### LINKED ISSUE #2150: Removing observed data entry from simulation - do not show a warning
Removing observed data from the "Observed data" tab of a simulation shows a warning every time:

image

This is annoying and unnecessary. Add an option "Do not ask again".

peterdew:
Fixes #2150 

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

To make ik possible for the user to suppress the warning-messagesbox when he deletes an observed data entry from a simulation, there is added an confirmationmessagebox wich contains a 'do not show this again-checkbox. 

This choice is only persisted in memory, so the text of the checkbox says:'Do not show this again until restart' 

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

2024-03-19 14_47_46-MoBi® 12 0 0+53e21410768fb2c5dbf152c5e74e0a82e42d1bfa

# Questions (if appropriate):

### DIFF
--- a/src/OSPSuite.Core/ICoreUserSettings.cs
+++ b/src/OSPSuite.Core/ICoreUserSettings.cs
@@ -1,20 +1,20 @@
 ﻿namespace OSPSuite.Core
 {
-   public interface ICoreUserSettings
-   {
-      /// <summary>
-      ///    Specifies how many core can be use by the application when performing parallel computations
-      /// </summary>
-      int MaximumNumberOfCoresToUse { get; set; }
+    public interface ICoreUserSettings
+    {
+        /// <summary>
+        ///    Specifies how many core can be use by the application when performing parallel computations
+        /// </summary>
+        int MaximumNumberOfCoresToUse { get; set; }
 
-      /// <summary>
-      ///    Number of bin per plot
-      /// </summary>
-      int NumberOfBins { get; set; }
+        /// <summary>
+        ///    Number of bin per plot
+        /// </summary>
+        int NumberOfBins { get; set; }
 
-      /// <summary>
-      /// Number of individuals per bin
-      /// </summary>
-      int NumberOfIndividualsPerBin { get; set; }
-   }
+        /// <summary>
+        /// Number of individuals per bin
+        /// </summary>
+        int NumberOfIndividualsPerBin { get; set; }
+    }
 }
--- a/src/OSPSuite.Core/ICoreUserSettings.cs
+++ b/src/OSPSuite.Core/ICoreUserSettings.cs
@@ -1,26 +1,26 @@
 ﻿namespace OSPSuite.Core
 {
-    public interface ICoreUserSettings
-    {
-        /// <summary>
-        ///    Specifies how many cores can be used by the application when performing parallel computations
-        /// </summary>
-        int MaximumNumberOfCoresToUse { get; set; }
+   public interface ICoreUserSettings
+   {
+      /// <summary>
+      ///    Specifies how many cores can be used by the application when performing parallel computations
+      /// </summary>
+      int MaximumNumberOfCoresToUse { get; set; }
 
-        /// <summary>
-        ///    Number of bins per plot
-        /// </summary>
-        int NumberOfBins { get; set; }
+      /// <summary>
+      ///    Number of bins per plot
+      /// </summary>
+      int NumberOfBins { get; set; }
 
-        /// <summary>
-        /// Number of individuals per bin
-        /// </summary>
-        int NumberOfIndividualsPerBin { get; set; }
-        
-        /// <summary>
-        /// Suppresses the warning when the user removes an ObservedData-entry from a simulation
-        /// </summary>
-        /// <remarks>This property is purposely not included in the serializer so the suppression works for the duration of an application-run.</remarks>
-        bool SuppressWarningOnRemovingObservedDataEntryFromSimulation { get; set; }
-    }
+      /// <summary>
+      /// Number of individuals per bin
+      /// </summary>
+      int NumberOfIndividualsPerBin { get; set; }
+
+      /// <summary>
+      /// Suppresses the warning when the user removes an ObservedData-entry from a simulation
+      /// </summary>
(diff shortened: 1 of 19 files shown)


src/OSPSuite.UI/Services/DialogCreator.cs
```diff
@@ -80,26 +83,44 @@ public ViewResult MessageBoxYesNo(string message, ViewResult defaultButton = Vie
...
       }
 
-      private ViewResult showQuestionMessageBox(string message, IReadOnlyList<DialogResult> buttons, string yes, string no, string cancel, ViewResult defaultButton)
+      public ViewResult MessageBoxConfirm(string message, Action doNotShowAgain, ViewResult defaultButton = ViewResult.Yes)
+      {
+         return showQuestionMessageBox(message, new[] {
+            DialogResult.Yes, DialogResult.Cancel}, string.Empty, string.Empty, string.Empty, defaultButton, doNotShowAgain);
+      }
+
+      private ViewResult showQuestionMessageBox(string message, IReadOnlyList<DialogResult> buttons, string yes, string no, string cancel, ViewResult defaultButton, Action doNotShowAgain)
```
rwmcintosh, line 95: I think if you make this ```doNotShowAgain``` argument take a default argument of ```null``` then you would not need to add ```null``` args in method calls.

src/OSPSuite.UI/Services/DialogCreator.cs
```diff
@@ -80,26 +83,44 @@ public ViewResult MessageBoxYesNo(string message, ViewResult defaultButton = Vie
...
             Localizer.Active = currentLocalizer;
          }
       }
 
+      private void onMessageBoxClosed(XtraMessageBoxClosedArgs e, Action doNotShowAgain)
+      {
+          // e.Visible gets whether the user checked the 'Do not show this message again' checkbox. Checked makes e.Visible = false 
+          if (!e.Visible && e.DialogResult == DialogResult.Yes)
+          {
+              doNotShowAgain?.Invoke();
```
rwmcintosh, line 117: because here, is where it's used, and it's checked for ```null``` first

src/OSPSuite.UI/Services/DialogCreator.cs
peterdew: You're right. Fixed in commit 1ad367850ff74c4378df84d90be4dac856d619db

src/OSPSuite.UI/Services/DialogCreator.cs
peterdew: You're right. Fixed in commit 1ad367850ff74c4378df84d90be4dac856d619db

src/OSPSuite.Core/ICoreUserSettings.cs
```diff
@@ -1,20 +1,26 @@
...
+        /// <summary>
+        /// Number of individuals per bin
+        /// </summary>
+        int NumberOfIndividualsPerBin { get; set; }
+        
+        /// <summary>
+        /// Suppresses the warning when the user removes an ObservedData-entry from a simulation
+        /// </summary>
+        /// <remarks>This property is purposely not included in the serializer so the suppression works for the duration of an application-run.</remarks>
+        bool SuppressWarningOnRemovingObservedDataEntryFromSimulation { get; set; }
```
msevestre, line 24: This feels wrong to me. I can't really explain why but this is really granular to b in CoreUserSettings
Also it really does not scale. If we have more of those settings we ned to create one for all possible use cases. 
Lastly, this feels like it will be serialized. Bassed on the issue , it says something about per session. Or I did not understand what was meant

src/OSPSuite.UI/UIConstants.cs
```diff
@@ -22,6 +22,7 @@ private static Color skinColorFor(string state)
       public const int TOOL_TIP_INITIAL_DELAY = 500;
       public const int TOOL_TIP_INITIAL_DELAY_LONG = 1500;
       public const string EMPTY_COLUMN = " ";
+      public const string DO_NOT_SHOW_THIS_AGAIN_UNTIL_RESTART = "Do not show this again until restart";
```
msevestre, line 25: texts are not stored in this file

msevestre (changes requested):
indentation is also completely off

src/OSPSuite.Core/ICoreUserSettings.cs
rwmcintosh: Peter and I were discussing about where to put the flag. It's not serialized right now but I bet someone could make the mistake of adding it if they found this and thought it was left out by accident.

Peter originally had it as a static flag in the task. What do you think about leaving it there? With comment about why its static.

src/OSPSuite.UI/UIConstants.cs
PavelBal: Let's put this into the developer documentation, so new devs know directly where to put string constants. What do you think?

src/OSPSuite.Core/ICoreUserSettings.cs
PavelBal: Maybe a good use case to start a section in the dev docs, explaining what belongs where? E.g. a short description of what goes into `CoreUserSettings` und what should not.

rwmcintosh: @PavelBal do you think there are other nag boxes that could use a flag like this? I imagine there are, and if so, then we should plan for a kind of session limited user settings object. 

In this case, I'd put a collection of dialog box Id's in a collection and that collection could be checked before the dialog is shown.

src/OSPSuite.UI/UIConstants.cs
rwmcintosh: @peterdew place this in OSPSuite.Assets.Captions. You will be able to see the conventions there.

PR #2221, opened 2024-03-22
Fixes #2220

### LINKED ISSUE #2220: Make Add and AddAtFront handle multi element paths as strings
We keep creating new cases of the bug where we try to append a pathelement as a string and it gets added as a single entry, but it actually has the pipe separator in there.

The string overrides for adding elements to a path should handle the cases where the intention is to add a multi-part path rather than a single element. We just have to break it up by the ```PATH_DELIMITER```

image

rwmcintosh:
Fixes #2220 

# Description
Prevents us from making the same mistakes when appending paths together

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/OSPSuite.Core/Domain/ObjectPath.cs
+++ b/src/OSPSuite.Core/Domain/ObjectPath.cs
@@ -207,7 +207,6 @@ namespace OSPSuite.Core.Domain
       /// <summary>
       ///    Add path entries to the end of the path
       /// </summary>
-      /// <param name="pathToAdd">path entry to add</param>
       public virtual void Add(string pathToAdd)
       {
          var pathElements = splitToElements(pathToAdd);
@@ -217,7 +216,6 @@ namespace OSPSuite.Core.Domain
       /// <summary>
       ///    Add path entries to the front of the path
       /// </summary>
-      /// <param name="pathToAdd">path entry to add</param>
       public virtual void AddAtFront(string pathToAdd)
       {
          var pathElements = splitToElements(pathToAdd);
--- a/src/OSPSuite.Core/Domain/ObjectPath.cs
+++ b/src/OSPSuite.Core/Domain/ObjectPath.cs
@@ -63,12 +63,6 @@ namespace OSPSuite.Core.Domain
       }
 
       /// <summary>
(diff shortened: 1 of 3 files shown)


src/OSPSuite.Core/Domain/ObjectPath.cs
```diff
@@ -211,20 +205,35 @@ IEnumerator IEnumerable.GetEnumerator()
...
+      public virtual void AddAtFront(string pathToAdd)
+      {
+         var pathElements = splitToElements(pathToAdd);
+         pathElements.Reverse().Each(x => _pathEntries.Insert(0, x));
+      }
+
+      private static string[] splitToElements(string pathToInsert)
       {
-         _pathEntries.Insert(0, pathEntry);
+         var pathItems = pathToInsert.Split(new[] {PATH_DELIMITER}, StringSplitOptions.None);
```
msevestre, line 227: there is a ToPathArray method. . I would use this one

src/OSPSuite.Core/Domain/ObjectPath.cs
```diff
@@ -211,20 +205,35 @@ IEnumerator IEnumerable.GetEnumerator()
...
       }
 
       /// <summary>
-      ///   Add a multipart <paramref name="pathToAdd"/> at the front
+      ///    Add a multipart <paramref name="pathToAdd" /> at the front
       /// </summary>
       public virtual void AddAtFront(ObjectPath pathToAdd)
       {
-         pathToAdd.Reverse().Each(AddAtFront);
+         AddAtFront(pathToAdd.PathAsString);
```
msevestre, line 236: This is a bit weird. You have an array, you create a string to then create an array again. I think we need a lower level method tht takes an array of string

src/OSPSuite.Core/Domain/ObjectPath.cs
rwmcintosh: hmm, some tests break when using that method

PR #2246, opened 2024-05-11
Fixes #2240

### LINKED ISSUE #2240: Merge behavior Mode:
To check

What happens when merging two mdoules with formula names the same. Expectation is that the formula from the "merged" module wins and replaces the existing one by name

Decision

At the module level, a "Merge beahvior Mode"
 - Overwritte
  * what we have at the moment in v12 on April 18th
  * We replace the whole container strucutre by name 

- Extend
  * We add non existant containers
  * We add non existant parameters
  * We replace existing paramaters
  * We replace existing formulas by name (same behavio as in OVerwritte mode)
    * Either in the parmaeter itselfs or the PV

Errors
 * We are replacing a formula (for instance Volume_4Comp) with a new formula that requires a local parmaeter in each container. And we don't add this parameter everywhere

Notes
In order to NOT overwrite a formula, it needs to have a different name.

msevestre:
Fixes #2240

# Description

I am introducing a `MergeBehavior` that will have to be set. I don't think it belongs in the module but more in the configuration. @rwmcintosh @PavelBal thoughts?

This is not all done yet but a good chunk is implemented. This thinking about formula does not make sense because we create NEW INSTANCES anyways. (not one formula but N formulas)
## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/OSPSuite.Core/Domain/Services/SpatialStructureMerger.cs
+++ b/src/OSPSuite.Core/Domain/Services/SpatialStructureMerger.cs
@@ -49,18 +49,18 @@ namespace OSPSuite.Core.Domain.Services
 
       private void createMergedContainerStructureInRoot(IContainer root, SimulationBuilder simulationBuilder)
       {
-         var allSpatialStructures = simulationBuilder.SpatialStructures;
-         if (!allSpatialStructures.Any())
+         var allSpatialStructureAndMergeBehaviors = simulationBuilder.SpatialStructureAndMergeBehaviors;
+         if (!allSpatialStructureAndMergeBehaviors.Any())
             return;
 
-         var firstSpatialStructure = allSpatialStructures[0];
-         var allOtherSpatialStructures = allSpatialStructures.Skip(1).ToList();
+         var firstSpatialStructure = allSpatialStructureAndMergeBehaviors[0].spatialStructure;
+         var allOtherSpatialStructures = allSpatialStructureAndMergeBehaviors.Skip(1).ToList();
          var mapToModelContainer = mapContainerDef(simulationBuilder);
          var mergeTopContainerIntoModel = mergeTopContainerInStructure(root);
 
          // First step: We create the container structure.
          // This is done by adding all top containers defined in the FIRST spatial structure
-         // Then we merge all other top containers defined in the other spatial structures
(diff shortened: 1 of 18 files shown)


src/OSPSuite.Core/Domain/Services/SpatialStructureMerger.cs
```diff
@@ -90,32 +91,32 @@ private void createMergedContainerStructureInRoot(IContainer root, SimulationBui
...
-         addOrReplaceContainer(topContainer, parentContainer);
+         addOrReplaceContainer(topContainer, parentContainer, mergeBehavior);
       }
 
-      private void addOrReplaceContainer(IContainer containerToAdd, IContainer parentContainer)
+      private void addOrReplaceContainer(IContainer containerToAdd, IContainer parentContainer, MergeBehavior mergeBehavior)
       {
-         //we do have a special case to deal with when dealing with MoleculeProperties container that we merge instead of replacing
-         if (containerToAdd.IsNamed(Constants.MOLECULE_PROPERTIES))
+         //Merge behavior is merged or we have a special case to deal with when dealing with MoleculeProperties container that we merge instead of replacing
```
rwmcintosh, line 118: Did you mean 'Merge behavior is extend' instead of 'merged'

tests/OSPSuite.HelpersForTests/ModuleHelperForSpecs.cs
```diff
@@ -118,6 +133,16 @@ private SpatialStructure getSpatialStructureModule1()
 
 
          //LUNG
+         // Lung
```
rwmcintosh, line 136: 👍

rwmcintosh:
Looks very good. I think the module remains unaware how it will be used. This belongs to the stage when the user is building a simulation configuration. So, that means we'll have to create a new option there. It's not just a matter of adding a module to a simulation configuration, but also, how is it to be used.

src/OSPSuite.Core/Domain/Services/SpatialStructureMerger.cs
msevestre: yep

PR #2253, opened 2024-06-06
#2256 Add new Exception and change message to show Module

### LINKED ISSUE #2256: Add module name to the error message when parent container not found
Right now, we get this message:

image

- [ ] If possible, add the name of the module from which "Container A" is.
- [x] Open-Systems-Pharmacology/MoBi#1352
image

benjaperez1983:
Fixes #2256

# Description

Changed the way we display the message for this exception by adding the module and building block information to it.
## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

# Questions (if appropriate):

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -1383,8 +1383,8 @@ namespace OSPSuite.Assets
       public static readonly string FoldValueMustBeGreaterThanOne = "Fold value must be a number greater than one.";
       public static readonly string ImporterEmptyFile = "The file you are trying to load is empty.";
 
-      public static string CannotFindParentContainerWithPath(string parentPath, string containerName, string buildingBlockName, string moduleName) 
-          => $"Cannot find parent container '{parentPath}' defined as target of container '{containerName}' in '{buildingBlockName}' in module '{moduleName}'";
+      public static string CannotFindParentContainerWithPath(string parentPath, string containerName, string buildingBlockName, string moduleName) => 
+         $"Cannot find parent container '{parentPath}' defined as target of container '{containerName}' in '{buildingBlockName}' in module '{moduleName}'";
 
       public static  string NoUnitColumnValues(string mappingName) => $"No values for the unit were found in the excel column mapped for '{mappingName}' \n";
 
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -1384,7 +1384,7 @@ namespace OSPSuite.Assets
       public static readonly string ImporterEmptyFile = "The file you are trying to load is empty.";
 
       public static string CannotFindParentContainerWithPath(string parentPath, string containerName, string buildingBlockName, string moduleName) => 
-         $"Cannot find parent container '{parentPath}' defined as target of container '{containerName}' in '{buildingBlockName}' in module '{moduleName}'";
+         $"Cannot find parent container '{parentPath}' defined as target of container '{containerName}' in building block '{buildingBlockName}' in module '{moduleName}'";
 
       public static  string NoUnitColumnValues(string mappingName) => $"No values for the unit were found in the excel column mapped for '{mappingName}' \n";
 
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -1383,7 +1383,8 @@ namespace OSPSuite.Assets
       public static readonly string FoldValueMustBeGreaterThanOne = "Fold value must be a number greater than one.";
       public static readonly string ImporterEmptyFile = "The file you are trying to load is empty.";
 
-      public static string CannotFindParentContainerWithPath(string parentPath, string containerName)=> $"Cannot find parent container '{parentPath}' defined as target of container '{containerName}'";
+      public static string CannotFindParentContainerWithPath(string parentPath, string containerName, string buildingBlockName, string moduleName) 
+          => $"Cannot find parent container '{parentPath}' defined as target of container '{containerName}' in '{buildingBlockName}' in module '{moduleName}'";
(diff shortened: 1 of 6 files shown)


src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1383,7 +1383,8 @@ public static class Error
       public static readonly string FoldValueMustBeGreaterThanOne = "Fold value must be a number greater than one.";
       public static readonly string ImporterEmptyFile = "The file you are trying to load is empty.";
 
-      public static string CannotFindParentContainerWithPath(string parentPath, string containerName)=> $"Cannot find parent container '{parentPath}' defined as target of container '{containerName}'";
+      public static string CannotFindParentContainerWithPath(string parentPath, string containerName, string buildingBlockName, string moduleName) 
+          => $"Cannot find parent container '{parentPath}' defined as target of container '{containerName}' in '{buildingBlockName}' in module '{moduleName}'";
```
rwmcintosh, line 1387: Can you put the `=>` at the end of the line above

src/OSPSuite.Core/Domain/Services/SpatialStructureMerger.cs
```diff
@@ -67,9 +67,9 @@ private void createMergedContainerStructureInRoot(IContainer root, SimulationBui
 
          //Merge all other spatial structures
          //make sure we map the container to a model container so that we do not change the original containers
-
-         allOtherSpatialStructuresWithMergeBehavior.Select(x => new {x.mergeBehavior, topContainers = x.spatialStructure.TopContainers.Select(mapToModelContainer).ToList()})
-            .Each(x => x.topContainers.Each(topContainer => mergeTopContainerInStructure(topContainer, root, x.mergeBehavior)));
+         allOtherSpatialStructuresWithMergeBehavior
+             .Select(item => new { item.mergeBehavior, topContainers = item.spatialStructure.TopContainers.Select(mapToModelContainer).ToList(), item })
```
rwmcintosh, line 71: Could this be `item.spatialStructure` at the end instead of `item`. 

It's used in `tryMergeTopContainerInStructure` only to access the SpatialStructure (and the module from there).

src/OSPSuite.Core/Domain/Services/SpatialStructureMerger.cs
```diff
@@ -89,6 +89,18 @@ private void createMergedContainerStructureInRoot(IContainer root, SimulationBui
          }
       }
 
+      private void tryMergeTopContainerInStructure(IContainer root, IContainer topContainer, MergeBehavior mergeBehavior, (SpatialStructure spatialStructure, MergeBehavior mergeBehavior) item)
```
rwmcintosh, line 92: then we wouldn't need the Tuple in here, and instead we'd only have 

`private void tryMergeTopContainerInStructure(IContainer root, IContainer topContainer, MergeBehavior mergeBehavior, SpatialStructure spatialStructure)`

rwmcintosh (approved):

src/OSPSuite.Core/Domain/Exceptions.cs
```diff
@@ -149,4 +149,11 @@ public CannotLockFileException(Exception exception) : base(exception.Message)
       {
       }
    }
+
+   public class ContainerNotFoundException : OSPSuiteException
```
msevestre, line 153: hum. I don't understand why we generate a specific  exception that does nothing. It's like a marker exception somehow but it's sued only once and caught in the same file. I don't think we need this and we can throw the original exception by adding the building block name no?

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1383,7 +1383,8 @@ public static class Error
       public static readonly string FoldValueMustBeGreaterThanOne = "Fold value must be a number greater than one.";
       public static readonly string ImporterEmptyFile = "The file you are trying to load is empty.";
 
-      public static string CannotFindParentContainerWithPath(string parentPath, string containerName)=> $"Cannot find parent container '{parentPath}' defined as target of container '{containerName}'";
+      public static string CannotFindParentContainerWithPath(string parentPath, string containerName, string buildingBlockName, string moduleName) => 
+         $"Cannot find parent container '{parentPath}' defined as target of container '{containerName}' in '{buildingBlockName}' in module '{moduleName}'";
```
msevestre, line 1387: should we say in building block xxxx ?

msevestre (commented):
@rwmcintosh Thoughts? Not fond of generating lots of specific exception, especially without parameters

rwmcintosh:
> @rwmcintosh Thoughts? Not fond of generating lots of specific exception, especially without parameters

Yeah, we looked at this one together and it's just that we'd be passing the building block and/or module and/or the string down into where the exception is thrown (it's a couple calls lower in the stack) _just in case_ the parent container can't be found. 

This way, we just throw this specific exception and catch it where the domain objects are being held so that we can create a meaningful message.

msevestre:
Ok so no reason for me to make this exception public. Can it be declared internal , even with the file itself ? This should not be used anywhere else

Also is base() required ?

rwmcintosh:
Yeah good idea. Nested class even.

src/OSPSuite.Core/Domain/Exceptions.cs
benjaperez1983: Thought about it, but when this new exception is thrown, we do not have all the data we need, I. E SpatialStructure.
So the execption will not have this information that is critical to this task.
Therefore we could either send this parameter in, but it will only be used ONLY for this exception, which will not always be thrown.
The other thing we could do is throw it and retrhrow the same exception overwritting the message.
SO let me know your thoughts about this.

rwmcintosh (approved):

PR #2293, opened 2024-08-05
#2292 parent path differences

### LINKED ISSUE #2292: "Parent path" differences are not shown in "Comparison"

benjaperez1983:
Fixes #2292 

# Description

Parent path differences now showing on comparison

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):


msevestre (changes requested):
Unit test missing
Bug fix=> unit test. No exception to the rule

benjaperez1983:
> Unit test missing Bug fix=> unit test. No exception to the rule

Well, it was not a bug in fact, I`ve marked it as that but it is a new feature.
Having said that, challenge accepted, it should have it`s tests.

IMPORTANT: 
I added one file and renamed an existing, since the naming of the spec class was not accurate at all.

tests/OSPSuite.Core.IntegrationTests/DiffBuilders/ContainerDiffBuilderSpecs.cs
```diff
@@ -1,99 +1,46 @@
...
-         var container1 = new Container {Name = "O1"};
-         container1.Add(new Parameter {BuildMode = ParameterBuildMode.Global, Name = "P1"});
-
-         var container2 = new Container {Name = "O2"};
-         container2.Add(new Parameter {BuildMode = ParameterBuildMode.Local, Name = "P1"});
-
-         _object1 = container1;
-         _object2 = container2;
-      }
+         var entityDiffBuilder = A.Fake<EntityDiffBuilder>();
```
msevestre, line 18: You are moving from Integration tests actually testing the real behavior to using fake. I don't understand the reasoning. In fact, why are you not using concern_for_ObjectComparer ? This does everything for you. You just have to set _object1 and object2 in all your subtests and everything is done for you. RIght now, you are reinvitening the wheel

tests/OSPSuite.Core.IntegrationTests/DiffBuilders/ContainerObjectComparerSpecs.cs
```diff
@@ -0,0 +1,99 @@
+﻿using System.Linq;
+using OSPSuite.BDDHelper;
+using OSPSuite.BDDHelper.Extensions;
+using OSPSuite.Core.Domain;
+
+namespace OSPSuite.Core.DiffBuilders
+{
+   public class When_comparing_two_container_having_different_base_properties_and_container_properties : concern_for_ObjectComparer
```
msevestre, line 8: Why did you rename the file. We are testing the behavior of ContainerDiff. We do not have an object called ContainerObjectComparer

msevestre (changes requested):
I do not udnerstand the proposed changes
We just need to add one test to what we had before making sure the diff for ParentPath is reported. Do not use Fake unless absolutely required.

tests/OSPSuite.Core.IntegrationTests/DiffBuilders/ContainerObjectComparerSpecs.cs
benjaperez1983: you are right about that, but this was called "ContainterDiffBuilderSpecs", we do have a ContainterDiffBuilder class abd this tests are testing the "ContainerDiffBuilder" at all.

Since the concern class is concern_for_ObjectComparer : ContextForIntegration<IObjectComparer> , I thought it should be named after that.
Let me know what do you think

tests/OSPSuite.Core.IntegrationTests/DiffBuilders/ContainerDiffBuilderSpecs.cs
benjaperez1983: it does not in fact, since the change I did is on the ContainerDiffBuilder Class, which was not being tested at all.
The tests (even when the class was called "ContainerDiffBuilderSpecs" is testing the "ObjectComparer" class, not the ContainerDiffBuilder.
And since my changes are on the ContainerDiffBuilder class, I created this tests and changed the previous class name as well.

tests/OSPSuite.Core.IntegrationTests/DiffBuilders/ContainerObjectComparerSpecs.cs
msevestre: this is not correct. We are testing the behavior of the ContainerDiff Builder implictely by making the diff of two contaienrs. This is how it goes here. Instead of always duplicating the logic, we created one base class, and we always create in the context the two objects to compare. I want to test the ContainerDiffBuilder, I create two containers  etc..

Because it's integration tests, we are testing the whole system as a whole but we are focusing here on Container. So in our case, we ARE Testing container diff builder by creating TWO containers. You just need to update the context to have two different ParentPath and that's it

tests/OSPSuite.Core.IntegrationTests/DiffBuilders/ContainerDiffBuilderSpecs.cs
```diff
@@ -96,4 +98,23 @@ public void should_have_set_the_parent_to_the_common_3rd_level()
...
+
+         _object1 = new Container { Name = "O1", ParentPath = new ObjectPath("A", "B") };
+         _object2 = new Container { Name = "O1", ParentPath = new ObjectPath("A", "C") };
+      }
+
+      [Observation]
+      public void should_return_a_report_containing_the_container_path_difference()
+      {
+         _report.Count().ShouldBeEqualTo(1);
+         Assert.IsTrue(_report.First().Description.StartsWith("Parent Paths are not equal"));
```
msevestre, line 117: ok this is it. One last change. WE never use the ASsert but use our BDD stuff
can you do something along the line of _report.First().....ShouldBeEqualTo 
and remove the ref to NuNit at the top. 
also is there a constant that you can use instead of the magic string Parent Paths are not equal? Are we typically comparing to strings in the tests ?

benjaperez1983: Sure thing, just did          _report.First().Description.StartsWith("Parent Paths are not equal").ShouldBeTrue();
which is in line with our asserts.

Great catch mate!

PR #2298, opened 2024-08-08
Fixes #2297

### LINKED ISSUE #2297: Formulas are resolved before applying PV - unwanted behavior
[formula-resolved-before-pb.zip](https://github.com/user-attachments/files/16523552/formula-resolved-before-pb.zip)

In the attached project, the module `Test2` defines a parameter `Organism|Param` by an invalid formula (the referenced parameter `foo` is not present).

image

Additionally, the value of the parameter is defined in the PV BB

image

When trying to create a simulation, an error is thrown that `foo` is not found. This behavior is unwanted, as I want the value defined in the PV BB to be applied without the formula being evaluated. This is how it worked in MoBi 11, and it was a way to overwrite parameters that otherwise would not be resolved.

msevestre:
Fixes #2297

# Description

Need to replace the formula that is wrong with the constant as we add it before, BUT ONLY if the formula is not valid probably

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/OSPSuite.Core/Domain/Services/QuantityValuesUpdater.cs
+++ b/src/OSPSuite.Core/Domain/Services/QuantityValuesUpdater.cs
@@ -32,19 +32,22 @@ namespace OSPSuite.Core.Domain.Services
       private readonly IFormulaFactory _formulaFactory;
       private readonly IConcentrationBasedFormulaUpdater _concentrationBasedFormulaUpdater;
       private readonly IParameterValueToParameterMapper _parameterValueToParameterMapper;
+      private readonly ValidatorForForFormula _formulaValidator;
 
       public QuantityValuesUpdater(
          IKeywordReplacerTask keywordReplacerTask,
          ICloneManagerForModel cloneManagerForModel,
          IFormulaFactory formulaFactory,
          IConcentrationBasedFormulaUpdater concentrationBasedFormulaUpdater,
-         IParameterValueToParameterMapper parameterValueToParameterMapper)
+         IParameterValueToParameterMapper parameterValueToParameterMapper,
+         ValidatorForForFormula formulaValidator)
       {
          _keywordReplacerTask = keywordReplacerTask;
          _cloneManagerForModel = cloneManagerForModel;
          _formulaFactory = formulaFactory;
          _concentrationBasedFormulaUpdater = concentrationBasedFormulaUpdater;
          _parameterValueToParameterMapper = parameterValueToParameterMapper;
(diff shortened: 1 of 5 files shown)


tests/OSPSuite.Core.IntegrationTests/ModuleIntegrationTests.cs
```diff
@@ -231,4 +232,49 @@ public void should_have_kept_existing_container()
...
+   internal class When_constructing_a_simulation_based_on_an_invalid_parmaeter_overwritten_with_PV : concern_for_ModuleIntegration
+   {
+      [Observation]
+      public void should_be_able_to_construct_the_simulation()
+      {
+         _result.ValidationResult.ValidationState.ShouldBeEqualTo(ValidationState.Valid, _result.ValidationResult.Messages.Select(x=>x.Text).ToString("\n"));
+      }
+
+      protected override Func<ModuleHelperForSpecs, SimulationConfiguration> SimulationConfigurationBuilder()
+      {
```
msevestre, line 245: @PavelBal @Yuri05 This is the exampel that pavel created. I just want to get a failing test

src/OSPSuite.Core/Domain/Services/QuantityValuesUpdater.cs
```diff
@@ -138,14 +141,23 @@ private IParameter getParameter(ModelConfiguration modelConfiguration, PathAndVa
...
+            return;
          }
+
+         //now we have a non constant formula. Let's try to see if the reference can be resolved. If yes, we will simply set the value of the parameter
+         //Otherwise, we will create a new constant formula with the value
+
+         if (_formulaValidator.IsFormulaValid(parameter))
+            parameter.Value = actualParameterValue;
+         else
+            parameter.Formula = _formulaFactory.ConstantFormula(actualParameterValue, parameter.Dimension);
```
msevestre, line 160: @Yuri05 @PavelBal  This was this flag we had before
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/main/src/OSPSuite.Core/Domain/Services/QuantityValuesUpdater.cs#L70C1-L71C1
We used to always overwrite . Now we just set the value but of course does not work if the formula is not valid. It works now with this

so the order of call was the same as before. However, the logic had changed

Yuri05 (approved):

PR #2303, opened 2024-08-14
#2302 add serialization property IsPKSimModule

### LINKED ISSUE #2302: "Create proccess rate" property should not change a PK-Sim module to an extension module
Same as https://github.com/Open-Systems-Pharmacology/MoBi/issues/1443

benjaperez1983:
Fixes #2302

# Description

Add serialization property IsPKSimModule so that it can set on PkSim export and used in Mobi Import and change.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/OSPSuite.Core/Domain/Module.cs
+++ b/src/OSPSuite.Core/Domain/Module.cs
@@ -40,7 +40,7 @@ namespace OSPSuite.Core.Domain
       private IReadOnlyList<T> buildingBlocksByType<T>() where T : IBuildingBlock => _buildingBlocks.OfType<T>().ToList();
 
       /// <summary>
-      ///    Module is a PKSim module if created in PKSim
+      ///    Module is a PKSim module if created in PKSim 
       /// </summary>
       public bool IsPKSimModule { get; set; }
 
--- a/src/OSPSuite.Core/Domain/Module.cs
+++ b/src/OSPSuite.Core/Domain/Module.cs
@@ -40,8 +40,7 @@ namespace OSPSuite.Core.Domain
       private IReadOnlyList<T> buildingBlocksByType<T>() where T : IBuildingBlock => _buildingBlocks.OfType<T>().ToList();
 
       /// <summary>
-      ///    Module is a PKSim module if created in PKSim and the module version
-      ///    matches the version when it was first imported
+      ///    Module is a PKSim module if created in PKSim
       /// </summary>
       public bool IsPKSimModule { get; set; }
 
--- a/src/OSPSuite.Core/Domain/Module.cs
+++ b/src/OSPSuite.Core/Domain/Module.cs
@@ -36,7 +36,6 @@ namespace OSPSuite.Core.Domain
    public class Module : ObjectBase, IEnumerable<IBuildingBlock>
    {
       private readonly List<IBuildingBlock> _buildingBlocks = new List<IBuildingBlock>();
-
       private T buildingBlockByType<T>() where T : IBuildingBlock => _buildingBlocks.OfType<T>().SingleOrDefault();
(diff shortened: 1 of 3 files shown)


msevestre:
Can you add the number being fixed please ?
Otherwise the issue isn’t closed automatically

src/OSPSuite.Core/Domain/Module.cs
```diff
@@ -36,15 +36,14 @@ public enum MergeBehavior
...
-
       private T buildingBlockByType<T>() where T : IBuildingBlock => _buildingBlocks.OfType<T>().SingleOrDefault();
       private IReadOnlyList<T> buildingBlocksByType<T>() where T : IBuildingBlock => _buildingBlocks.OfType<T>().ToList();
 
       /// <summary>
       ///    Module is a PKSim module if created in PKSim and the module version
       ///    matches the version when it was first imported
       /// </summary>
-      public bool IsPKSimModule => !string.IsNullOrEmpty(PKSimVersion) && Equals(Version, ModuleImportVersion);
+      public bool IsPKSimModule { get; set; }
```
msevestre, line 46: Hum.  Why is this change required? We may end up with version not matching and the flag set to true

benjaperez1983: yeah well, from this story we are going to decouple the version from the IsPkSimModule, meaning that the version will be only for showing the simulation on Sync or not, but the PkSimModule will be just reflected by this property.
The reason for this, is that now, there are some propeties (the one mentioned on the original story) that should not change the Type of module (AKA IsPkSimModule) but should change the version.

msevestre: Well then update the property comment because it’s wrong

msevestre: Comment is misleading.  If the module was created in PKSim and something was changed then the flag is false but it was stjlll created in PK-Sim.   I don’t know what the first issue is about so I am not sure what constitute a PK-sim module 
@PavelBal can maybe explain

benjaperez1983: well would you say this comment is ok?
"Module is a PKSim module at the time of creation in PKSim, can be updated by changing some properties on Mobi" ?

PavelBal:
>  I don’t know what the first issue is about so I am not sure what constitute a PK-sim module

The original issue is about changing this property:

image

Should not convert a PK-Sim module to an extension module.

src/OSPSuite.Core/Domain/Module.cs
rwmcintosh: Yeah, the issue is that some commands in MoBi will change the building block but should not convert the module to extension.

We are going to need the commands themselves to know if they should unset this flag.

tests/OSPSuite.Core.Tests/Domain/ModuleSpecs.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
 using FakeItEasy;
+using NUnit.Framework;
```
rwmcintosh, line 4: can you check if this is really needed? Seems it should not be.

rwmcintosh (approved):

Yuri05:
I assume, with this change PK-Sim has to set `module.IsPKSimModule = true` before exporting a module?
Do we have an issue for this in the PK-sim repo?

benjaperez1983:
> I assume, with this change PK-Sim has to set `module.IsPKSimModule = true` before exporting a module? Do we have an issue for this in the PK-sim repo?

We do this is the issue:
github.com/Open-Systems-Pharmacology/PK-Sim/issues/2953

Yuri05 (approved):

PR #2309, opened 2024-09-04
#2308 Select a neighborhood consider parent path

### LINKED ISSUE #2308: Selecting a neighbor in the neighborhood view: Consider "parent path"
Consider this use case:

image

The second neighbor of the selected neighborhood points to `Organism|ArterialBlood_UmbilicalCord|BloodCells`. While `Organism`is not present in this module, it is set as the parent of `ArterialBlood_UmbilicalCord`, so the full path could be resolved and the container shown in the right pane.

benjaperez1983:
Fixes #2308

# Description

 Select a neighborhood consider parent path

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/OSPSuite.Core/Domain/ObjectPath.cs
+++ b/src/OSPSuite.Core/Domain/ObjectPath.cs
@@ -127,7 +127,10 @@ namespace OSPSuite.Core.Domain
             return resolvedEntity;
 
          var root = refEntity.RootContainer;
-         var firstEntry = _pathEntries[0];
+
+         removeMatchingPaths(root, usePath);
+
+         var firstEntry = usePath[0];
 
          //We have an absolute Path from the ref entity
          if (string.Equals(firstEntry, refEntity.Name))
@@ -144,6 +147,15 @@ namespace OSPSuite.Core.Domain
          }
 
          return resolvePath<T>(dependentObject, usePath);
+      }
+
+      private void removeMatchingPaths(IContainer root, List<string> usePath)
+      {
+         var matchingElements = !string.IsNullOrEmpty(root?.ParentPath?.PathAsString) &&
+                                new ObjectPath(_pathEntries).PathAsString.StartsWith(root?.ParentPath?.PathAsString ?? string.Empty);
+
+         if (matchingElements)
+            usePath.RemoveRange(0, root.ParentPath.Count);
       }
 
       public virtual T Clone<T>() where T : ObjectPath => new ObjectPath(_pathEntries).DowncastTo<T>();
--- a/src/OSPSuite.Core/Domain/ObjectPath.cs
+++ b/src/OSPSuite.Core/Domain/ObjectPath.cs
@@ -151,8 +151,8 @@ namespace OSPSuite.Core.Domain
 
       private void removeMatchingParentPath(IContainer root, List<string> usePath)
       {
-         var matchingElements = !string.IsNullOrEmpty(root?.ParentPath?.PathAsString) &&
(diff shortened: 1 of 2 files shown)


src/OSPSuite.Core/Domain/ObjectPath.cs
```diff
@@ -146,6 +149,15 @@ public virtual T Resolve<T>(IEntity refEntity) where T : class
          return resolvePath<T>(dependentObject, usePath);
       }
 
+      private void removeMatchingPaths(IContainer root, List<string> usePath)
+      {
+         var matchingElements = !string.IsNullOrEmpty(root?.ParentPath?.PathAsString) &&
+                                new ObjectPath(_pathEntries).PathAsString.StartsWith(root?.ParentPath?.PathAsString ?? string.Empty);
```
Yuri05, line 155: the part after `&&` could be simplified to 
`new ObjectPath(_pathEntries).PathAsString.StartsWith(root.ParentPath.PathAsString);`
(because the part before `&&` already ensures that `root.ParentPath.PathAsString` is not null and not empty)

rwmcintosh (approved):
I would just rename that one method a little more explicitly.

tests/OSPSuite.Core.Tests/Domain/ObjectPathSpecs.cs
```diff
@@ -116,6 +117,39 @@ public void should_return_the_desired_entity()
...
+         }
+      }
+
+      [Observation]
+      public void should_return_the_desired_entity()
+      {
+         sut.Resolve<IParameter>(_organ).ShouldBeEqualTo(_parameter);
+      }
+
+      private static string GenerateRandomString(Random random, int length)
```
msevestre, line 145: I have no idea why we need to generate random strings suddenly in a text. I have no idea what's going on.
Please create a path that we can analyze visually and not in a weird loop.

src/OSPSuite.Core/Domain/ObjectPath.cs
```diff
@@ -146,6 +149,15 @@ public virtual T Resolve<T>(IEntity refEntity) where T : class
          return resolvePath<T>(dependentObject, usePath);
       }
 
+      private void removeMatchingParentPath(IContainer root, List<string> usePath)
+      {
+         var matchingElements = !string.IsNullOrEmpty(root?.ParentPath?.PathAsString) &&
```
msevestre, line 154: same `root?.ParentPath?.PathAsString` locally in a variable  and check that it's not null or empty then return

src/OSPSuite.Core/Domain/ObjectPath.cs
```diff
@@ -146,6 +149,15 @@ public virtual T Resolve<T>(IEntity refEntity) where T : class
...
+                                new ObjectPath(_pathEntries).PathAsString.StartsWith(root?.ParentPath?.PathAsString ?? string.Empty);
```
msevestre, line 155: that ?? can never happen since you check before

msevestre: the code is kind of weird. we are using _pathEntries (global) to modify the parameter pass (usePath). 
this works because usePath is created using _pathEntries but this is really hard to follow. 
Also we have lists, we got to string, we compare strings and we then remove from list.
Should't we just use...no string at all?

Yuri05 (approved):

PR #2341, opened 2024-10-21
Fixes #2307 PV BB: "Add Protein expression" - wrong entries are added

### LINKED ISSUE #2307: PV BB: "Add Protein expression" - wrong entries are added?
The user can select organ(s) and molecules for which to add the expression profiles. Are the values taken from an existing expression profile (if so - from which)? If not - should the values be 0 by default (seems like they are)?

If I do this for CYP3A4 and "Liver" as organ, I get the following entries:

image

If I filter for "Liver" in the expression profile of CYP3A4, I see the following parameters:

image

- `Relative expression in vascular endothelium`, `Relative expression in plasma`, and `Relative expression in blood cells` should not be there
- `Initial concentration`, `Fraction expressed interstitial`, `Fraction expressed intracellular` are missing

rwmcintosh:
Fixes #2307

# Description
Changing how the 'Expression Related' parameters are extended in a parameter values building block. We are using the names from existing expression building blocks to decide which are 'Expression'.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):


src/OSPSuite.Core/Domain/Services/ParameterValuesCreator.cs
```diff
@@ -85,18 +89,72 @@ public ParameterValue CreateParameterValue(ObjectPath parameterPath, double valu
...
+         parameterValues.Each(formulaTarget =>
+         {
+            var formulaSource = formulaSourceFor(expressionParameters.AllByName(formulaTarget.Name).ToList(), formulaTarget);
+
+            if (formulaSource != null)
+               formulaTarget.Formula = _cloneManager.Clone(formulaSource.Formula, formulaCache);
+         });
+      }
+
+      private bool shouldCloneFormulaFor(ParameterValue parameterValue) => parameterValue.Formula == null || parameterValue.Formula.IsConstant();
```
Yuri05, line 111: I don't quite understand why the function returns `true` if `parameterValue.Formula = null`

src/OSPSuite.Core/Domain/Services/ParameterValuesCreator.cs
```diff
@@ -85,18 +89,72 @@ public ParameterValue CreateParameterValue(ObjectPath parameterPath, double valu
...
+         nameMatchedExpressionParameters.Where(hasCompartment).FirstOrDefault(x => Equals(compartmentFor(x), compartmentFor(formulaTarget)));
+
+      private static string compartmentFor(ParameterValue formulaTarget) => formulaTarget.Path[compartmentIndex(formulaTarget)];
+
+      private static int compartmentIndex(PathAndValueEntity x) => x.Path.Count - 3;
+
+      private static ExpressionParameter pathMatchedExpressionParameterFor(List<ExpressionParameter> nameMatchedExpressionParameters, ParameterValue formulaTarget) =>
+         nameMatchedExpressionParameters.FirstOrDefault(x => Equals(x.Path, formulaTarget.Path));
+
+      private static bool hasCompartment(ParameterValue formulaTarget) => formulaTarget.Path.Count >= 5;
```
Yuri05, line 133: uff
That's hard to understand and requires at least a comment what 5 means (ideally with a path example).

src/OSPSuite.Core/Domain/Services/ParameterValuesCreator.cs
```diff
@@ -85,18 +89,72 @@ public ParameterValue CreateParameterValue(ObjectPath parameterPath, double valu
...
+
+         return formulaSource;
+      }
+
+      private static ExpressionParameter compartmentMatchedExpressionParameterFor(ParameterValue formulaTarget, List<ExpressionParameter> nameMatchedExpressionParameters) =>
+         nameMatchedExpressionParameters.Where(hasCompartment).FirstOrDefault(x => Equals(compartmentFor(x), compartmentFor(formulaTarget)));
+
+      private static string compartmentFor(ParameterValue formulaTarget) => formulaTarget.Path[compartmentIndex(formulaTarget)];
+
+      private static int compartmentIndex(PathAndValueEntity x) => x.Path.Count - 3;
```
Yuri05, line 128: once I see a comment (s. below) I can judge if it's correct or not 😄

src/OSPSuite.Core/Domain/Services/ParameterValuesCreator.cs
Yuri05: And we have compartments on different levels, so hmm...
But waiting for the comment first :)

src/OSPSuite.Core/Domain/Services/ParameterValuesCreator.cs
rwmcintosh: Oh, that's because I don't want to overwrite it if the ParameterValue already has a formula due to the molecule parameter it's based on.

src/OSPSuite.Core/Domain/Services/ParameterValuesCreator.cs
rwmcintosh: Ok, I commented these. This seems to be by convention. We use this kind of convention elsewhere for Expression parameters. But, maybe 5 is not the right number. 

Eg. if we have a path with only 3 elements, then we could take the third-last as compartment maybe. That would be the first element.

src/OSPSuite.Core/Domain/Services/ParameterValuesCreator.cs
rwmcintosh: I added a comment where this is used.

src/OSPSuite.Core/Domain/Services/ParameterValuesCreator.cs
rwmcintosh: yeah, for example t 1/2 (liver) would return false. 

image

Yuri05 (approved):

PR #2364, opened 2024-11-11
Fixes #2358 It should not be possible to create a simulation with multiple identical neighborhoods

### LINKED ISSUE #2358: It should not be possible to create a simulation with multiple identical neighborhoods
Reported by @PavelBal https://github.com/Open-Systems-Pharmacology/MoBi/issues/1618

> [neighborhoods-diff-name-same-neighbors.zip](https://github.com/user-attachments/files/17630328/neighborhoods-diff-name-same-neighbors.zip)
> 
> 
> Currently, it is possible to create multiple neighborhoods between the same two containers but with different names. This can result in transport processes being created multiple times and unexpected model behavior (spent almost 5 hours debugging a project).
> 
> This should not be allowed; only one neighborhood between the same two containers should be allowed. Therefore, at the simulation creation step (when combining the modules), for each NEW neighborhood (i.e., the neighborhood with a new name), it should be checked whether a neighborhood between the specified containers already exists (disregarding what the first or the second neighbor is). 
> 
> To speed up the process, maintain a map of neighborhood paths that are already connected?
> 
> If a new neighborhood between two already connected containers is being added, show an error "Neighborhood \<NeighborhoodName> cannot be created because a neighborhood between the containers \<Container1> and \<Container2> already exists. Every pair of containers can be connected by one neighborhood only".
> 
> Attached is a project for testing. It contains three modules and three combinations of the modules (simulations). "Module 1" has three containers - A, B, and C, and the neighborhoods between containers A and B and between B and C, and the simulation "Valid_Simulation" can be created from this one module:
> 
> image
> 
> Module 2 defines a neighborhood between containers B and A, and it is combined with the first module to the simulation "Invalid_1", which has two neighborhoods between A and B. This should NOT be allowed:
> 
> image
> 
> Module 3 defines a neighborhood between A and B (the same as Module 1, but with a different name). It is combined with the first module to the simulation "Invalid_2", which again has two neighborhoods between A and B. This should NOT be allowed:
> 
> image
>

rwmcintosh:
Fixes #2358 

# Description
It's possible that two modules could define different neighborhoods that connect the same containers. This prevents the simulation from being created.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

# Questions (if appropriate):

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -1845,6 +1845,16 @@ namespace OSPSuite.Assets
       public static string TypeNotSupported(string typeName) => $"{typeName} is not currently been handled";
 
       public static string CannotAccessValueByPathUsingWildCard(string path) => $"Accessing value by path is not supported for path containing wildcard ({path})";
+
+      public static string EquivalentNeighborhoodsAreDefinedFor(string firstNeighbor, string secondNeighbor)
+      {
+         return $"More than one neighborhood is defined connecting '{firstNeighbor}' and '{secondNeighbor}'";
+      }
+
+      public static string NeighborhoodDefinition(string neighborhoodName, string firstNeighbor, string secondNeighbor, string spatialStructure)
+      {
+         return $"Neighborhood '{neighborhoodName}' connects '{firstNeighbor}' and '{secondNeighbor}' in '{spatialStructure}'";
+      }
    }
 
    public static class Validation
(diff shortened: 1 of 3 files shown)


rwmcintosh:
actually, I think the full path is needed rather than just the names.

rwmcintosh:
> actually, I think the full path is needed rather than just the names.

ok, it is this way already, in this simple example, the paths are only one element deep.

rwmcintosh:
image

Yuri05 (approved):

PR #2377, opened 2024-12-10
Fixes #2376 Consolidate invalid project version messages

### LINKED ISSUE #2376: Consolidate invalid project version messages

rwmcintosh:
Fixes #2376

# Description
There are currently separate implementations of this message in PK-Sim and MoBi. They are the same, but it makes sense only to define it once.

Also, this changes the message to add the display versions to the message where possible.

## Type of change
purely caption

image

image

# Questions (if appropriate):

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -1374,21 +1374,17 @@ namespace OSPSuite.Assets
 
       private static string projectNameAndVersionAsString(string versionDisplay, int version) => $"V{versionDisplay} {numberDisplay(version)}";
 
-      public static string ProjectVersionCannotBeLoaded(int projectVersion, 
-         string oldestSupportedDisplayVersion, int oldestSupportedVersion, 
-         string currentSupportedDisplayVersion,int currentSupportedVersion,
-         bool projectIsTooOld, 
+      public static string ProjectVersionCannotBeLoaded(
+         int projectVersion, 
+         string oldestSupportedDisplayVersion, 
+         int oldestSupportedVersion, 
+         string currentSupportedDisplayVersion,
+         int currentSupportedVersion,
          string downloadUrl)
       {
          var compositeOldestVersion = projectNameAndVersionAsString(oldestSupportedDisplayVersion, oldestSupportedVersion);
          var compositeCurrentVersion = projectNameAndVersionAsString(currentSupportedDisplayVersion, currentSupportedVersion);
          var preamble = $"This application is compatible with projects created between {compositeOldestVersion} and {compositeCurrentVersion}";
-         if (projectIsTooOld)
-         {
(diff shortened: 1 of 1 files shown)


msevestre:
the range is kind of weird. Where does the max comes from?

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1369,6 +1369,36 @@ public static class DeviationLines
...
          }
       }
+
+      private static string numberDisplay(int version) => $"(#{version})";
+
+      private static string projectNameAndVersionAsString(string versionDisplay, int version) => $"V{versionDisplay} {numberDisplay(version)}";
+
+      public static string ProjectVersionCannotBeLoaded(int projectVersion, 
+         string oldestSupportedDisplayVersion, int oldestSupportedVersion, 
+         string currentSupportedDisplayVersion,int currentSupportedVersion,
```
msevestre, line 1379: some resharper goodness required here to format

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1369,6 +1369,36 @@ public static class DeviationLines
...
+         }
+
+         if (projectVersion > currentSupportedVersion)
+         {
+            return $"{preamble}.\n\n" +
+                   $"The version of this project {numberDisplay(projectVersion)} is newer and cannot be loaded.\n\n" +
+                   $"Visit our download page at {downloadUrl}";
+         }
+
+         return $"Work in progress.\nThis project file is too old {numberDisplay(projectVersion)} and cannot be loaded.\nSorry :-(";
```
msevestre, line 1400: should that be gone and the 1387 to 1390 be the default?

src/OSPSuite.Assets/UIConstants.cs
```diff
@@ -1369,6 +1369,36 @@ public static class DeviationLines
...
       }
+
+      private static string numberDisplay(int version) => $"(#{version})";
+
+      private static string projectNameAndVersionAsString(string versionDisplay, int version) => $"V{versionDisplay} {numberDisplay(version)}";
+
+      public static string ProjectVersionCannotBeLoaded(int projectVersion, 
+         string oldestSupportedDisplayVersion, int oldestSupportedVersion, 
+         string currentSupportedDisplayVersion,int currentSupportedVersion,
+         bool projectIsTooOld,
```
msevestre, line 1380: why do we need this flag ? isn it projectVersion < oldestSupportedVersion ?

msevestre (changes requested):
some questions

src/OSPSuite.Assets/UIConstants.cs
rwmcintosh: Yeah, I don't think this can ever be seen. Especially if the too old flag is removed.

PR #2390, opened 2025-01-13
Fixes #2300 Importing with column unit mapping

### LINKED ISSUE #2300: Importing with column unit mapping
When you import data with units described in a column, it should be possible to select the dimension.

For example, when importing mass in mg the importer will select the wrong dimension since input dose contains the mg unit, and is alphabetically first.

image

image

rwmcintosh:
Fixes #2300

# Description
If users try to map a dimension from a column that contains units then this has a couple of improvements

1. Formerly, only the first unit was used to determine the dimension, now we use all the units from the column
2. If there was more than one dimension supporting the units of the column then the first in the list would be selected (that list was always alaphabetic in the importer.

What this PR does - if there is an ambiguous unit->dimension mapping in the dataset then we present a modal dialog and the user must select the dimension for the column. There is also an 'Apply to All' button for each row so that if there are several columns in the data with the same dimensions you can override them all.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

See this data (there are two sheets):
image

Adding the sheets
image

Since both sheets have the same 'possible dimensions' using the apply to all maps both
image

# Questions (if appropriate):


PavelBal:
But what happens if you have multiple data sets with different dimension from the same sheet?

rwmcintosh:
> But what happens if you have multiple data sets with different dimension from the same sheet?

I thought of that, but not hard enough. You can't tell the 'Group' difference.
image


I'll add it

rwmcintosh:
image

src/OSPSuite.Presentation/DTO/DimensionSelectionDTO.cs
```diff
@@ -0,0 +1,39 @@
...
+      public string DisplayName
+      {
+         get
+         {
+            if (string.IsNullOrEmpty(_description))
+               return $"{_sheetName} - {Column.Name}";
+            return $"{_sheetName} - {Column.Name} - {_description}";
+         }
+      }
+   }
```
Yuri05, line 38: for CSV files, sheet name will be empty. I assume the name should be then just `{Column.Name} - {_description}`

rwmcintosh: Yes, I refactored it now to deal with csv files better

Yuri05 (approved):

PR #2403, opened 2025-01-22
what if we change the interface

rwmcintosh:


src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationDataSelectionPresenter.cs
```diff
@@ -129,16 +128,5 @@ public void Handle(SimulationReplacedInParameterAnalyzableEvent eventToHandle)
...
-
-      public void Handle(WeightObservedDataChangedEvent eventToHandle)
-      {
-         var existingMapping = _outputMappingPresenter.OutputMappings.Where(x => x.OutputPath == eventToHandle.OutputMapping.OutputPath).FirstOrDefault();
-
-         if (existingMapping == null)
-            return;
-
-         _weightedObservedDataPresenter.Edit(eventToHandle.OutputMapping.WeightedObservedData);
-      }
```
rwmcintosh, line 141: this was the part that was bugging me. It's because in practice we are updating one presenter based on the selection in another presenter, and then for this PR we need to react to data that was already replaced.

rwmcintosh: It feels like this presenter knows too much about what the `_outputMappingPresenter` and the `_weightedObservedDataPresenter` are doing.

src/OSPSuite.Presentation/Presenters/ParameterIdentifications/ParameterIdentificationWeightedObservedDataPresenter.cs
```diff
@@ -76,14 +78,20 @@ public string Caption
...
 
-         Edit(_observedData);
+         Edit(_outputMapping);
+      }
+
+      public void Handle(WeightObservedDataChangedEvent eventToHandle)
+      {
+         if(Equals(_outputMapping, eventToHandle.OutputMapping))
+            Edit(eventToHandle.OutputMapping);
       }
```
rwmcintosh, line 95: Now we can have an event handler that only cares about its own business.

PR #2459, opened 2025-03-31
#2388 Mobi stops working when trying to load huge file observed data

### LINKED ISSUE #2388: MoBi stops working when trying to add observed data
Happens both in 11.3 and 12.0:
when trying to add observed data set from this file, MoBi freezes forever (killed after 20 minutes waiting)
[02_ObsDataPK_OSP.xlsx](https://github.com/user-attachments/files/18361805/02_ObsDataPK_OSP.xlsx)

benjaperez1983:
Fixes #2388 

# Description

Add Working indicator when large volume of data is importing, since the app is not crashing but rather taking too long to load.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

image

# Questions (if appropriate):


PavelBal:
Can the user cancel the action? If not, let's try to do it.

E.g., after clicking on "load", show a window "Processing observed data" with "Cancel" button.

msevestre:
This can't be canceled at the moment so I won't review if @PavelBal's stuff is required

rwmcintosh:
@benjaperez1983 can you add a new screenshot? I think the screenshot in the description was the old one.

src/OSPSuite.Core/Services/IHeavyWorkManager.cs
```diff
@@ -19,15 +20,15 @@ public interface IHeavyWorkManager
       ///    action was canceled
       /// </summary>
       /// <param name="heavyWorkAction">Action that need to be performed</param>
-      bool Start(Action heavyWorkAction);
+      bool Start(Action heavyWorkAction, CancellationTokenSource cts = default);
```
rwmcintosh, line 23: Given that the other param is described in the XML comment the new one has to be described too otherwise it's a compiler warning

src/OSPSuite.Core/Services/IHeavyWorkManager.cs
```diff
@@ -19,15 +20,15 @@ public interface IHeavyWorkManager
...
 
       /// <summary>
       ///    run the heavy work action. Returns true if the action ran successfully, and false, if an exception was thrown or the
       ///    action was canceled
       /// </summary>
       /// <param name="heavyWorkAction">Action that need to be performed</param>
       /// <param name="caption">Caption that will be displayed</param>
-      bool Start(Action heavyWorkAction, string caption);
+      bool Start(Action heavyWorkAction, string caption, CancellationTokenSource cts = default);
```
rwmcintosh, line 31: here too

src/OSPSuite.Core/Services/IHeavyWorkManager.cs
rwmcintosh, line 23: Given that the other param is described in the XML comment the new one has to be described too otherwise it's a compiler warning

src/OSPSuite.Core/Services/IHeavyWorkManager.cs
rwmcintosh, line 31: here too

src/OSPSuite.Presentation/PresentationRegister.cs
```diff
@@ -84,6 +80,10 @@ public override void RegisterInContainer(IContainer container)
          container.Register<ChartEditorAndDisplaySettings, ChartEditorAndDisplaySettings>();
          container.Register<ChartEditorSettings, ChartEditorSettings>();
          container.Register<ChartPresenterContext, ChartPresenterContext>();
+         // These registrations need to be provided since
+         // the Factory needs the actual Implementation now, not the Interface.
+         container.Register<HeavyWorkPresenter, HeavyWorkPresenter>();
+         container.Register<HeavyWorkCancellablePresenter, HeavyWorkCancellablePresenter>();
```
rwmcintosh, line 86: If the factory does it's work using interface, then it should be fine to not register explicitly and instead rely on convention

```
         if (supportsCancellation)
            return Create<IHeavyWorkCancellablePresenter>();  
         return Create<IHeavyWorkPresenter>();  
```

src/OSPSuite.Presentation/Presenters/IHeavyWorkPresenterFactory.cs
```diff
@@ -5,13 +5,18 @@ namespace OSPSuite.Presentation.Presenters
...
+   public class HeavyWorkPresenterFactory : DynamicFactory<IHeavyWorkPresenter>, IHeavyWorkPresenterFactory
    {
-      public HeavyWorkPresenterFactory(IContainer container) : base(container)
+      public HeavyWorkPresenterFactory(IContainer container) : base(container) { }
+
+      public IHeavyWorkPresenter Create(bool supportsCancellation = false)
       {
+         if (supportsCancellation)
+            return Create<HeavyWorkCancellablePresenter>();  
+         return Create<HeavyWorkPresenter>();
```
rwmcintosh, line 19: use `Create<IHeavy...>`?

tests/OSPSuite.Presentation.Tests/Presentation/HeavyWorkManagerSpecs.cs
```diff
@@ -133,5 +135,43 @@ public void should_dispose_of_the_presenter_when_the_action_is_completed()
       {
          A.CallTo(() => _heavyWorkPresenter.Dispose()).MustHaveHappened();
       }
+
+      public class When_the_heavy_work_manager_is_starting_an_action : concern_for_HeavyWorkManager
+      {
+         private Action _action;
```
rwmcintosh, line 141: This is never assigned. You can use an anonymous action inline instead `() => { }`

src/OSPSuite.R/RRegister.cs
```diff
@@ -41,7 +46,13 @@ public override void RegisterInContainer(IContainer container)
          container.Register<SimulationBatch, SimulationBatch>();
          container.Register<ISimulationBatchFactory, SimulationBatchFactory>();
          container.Register<IObjectTypeResolver, RObjectTypeResolver>();
-
+         container.Register<HeavyWorkManager, HeavyWorkManager>();
+         container.Register<IHeavyWorkManager, HeavyWorkManager>();
+         container.Register<IHeavyWorkPresenterFactory, HeavyWorkPresenterFactory>();
+         container.Register<HeavyWorkPresenter, HeavyWorkPresenter>();
+         container.Register<HeavyWorkCancellablePresenter, HeavyWorkCancellablePresenter>();
+         container.Register<IHeavyWorkCancellableView, HeavyWorkCancellableView>();
```
rwmcintosh, line 54: After the changes to the `HeavyWorkPresenterFactory` remove these registrations and remove the project references. If there are IoC failures we'll need to find another way to solve that failure because we can't add the ProjectReferences to support this.

src/OSPSuite.R/OSPSuite.R.csproj
```diff
@@ -27,6 +27,8 @@
     <ProjectReference Include="..\OSPSuite.Infrastructure.Autofac\OSPSuite.Infrastructure.Autofac.csproj" />
     <ProjectReference Include="..\OSPSuite.Infrastructure.Import\OSPSuite.Infrastructure.Import.csproj" />
     <ProjectReference Include="..\OSPSuite.Infrastructure\OSPSuite.Infrastructure.csproj" />
+    <ProjectReference Include="..\OSPSuite.Presentation\OSPSuite.Presentation.csproj" />
+    <ProjectReference Include="..\OSPSuite.UI\OSPSuite.UI.csproj" />
```
rwmcintosh, line 31: We can't add these. That means that the UI and presentation project are required for R but that should not be the case. R should not need access to any of that.

rwmcintosh (changes requested):

rwmcintosh:
image

Can you implement the view with a `UxLayouControl` since it now has more than one item?

Then, it's worthwhile to make it look the same as if there is no cancel button and only show the border and extra space if the cancel button is shown.

rwmcintosh:
We had removed all the padding when we added the UxLayouControl, but I think if the dialog is going to show the borders, then we need to add it back. What do you think?

image

I just searched for a simple one with cancel.
image

if we get the controls off the border I think that will help.

src/OSPSuite.UI/Views/HeavyWorkView.cs
```diff
@@ -47,5 +74,10 @@ public bool Canceled
       public bool ExtraEnabled { get; set; }
       public bool ExtraVisible { get; set; }
       public bool CancelVisible { get; set; }
+
+      private void btnCancel_click()
```
rwmcintosh, line 78: don't use the convention of `_` in the events. I know it's the default.

rwmcintosh (approved):
Just rename that forms click event handler.

rwmcintosh (approved):

src/OSPSuite.R/MinimalImplementations/HeavyWorkManager.cs
```diff
@@ -0,0 +1,41 @@
...
+   }
+
+   public class HeavyWorkManagerFailingForSpecs : HeavyWorkManager
+   {
+      public override bool Start(Action heavyWorkAction, string caption, CancellationTokenSource cts = default)
+      {
+         base.Start(heavyWorkAction, caption);
+         return false;
+      }
+   }
```
Yuri05, line 39: Why do we need this here?

rwmcintosh: We don't

Yuri05 (approved):

PR #2521, opened 2025-05-16
Fixes #2520 Add snapshot as string to a module

### LINKED ISSUE #2520: Add snapshot as string to a module

rwmcintosh:
Fixes #2520

# Description
Allow the transfer of snapshot as string

## Type of change

# How Has This Been Tested?

- [x] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):


src/OSPSuite.Core/Domain/Module.cs
```diff
@@ -44,6 +45,15 @@ public class Module : ObjectBase, IEnumerable<IBuildingBlock>
       /// </summary>
       public bool IsPKSimModule { get; set; }
 
+      /// <summary>
+      /// Holds a snapshot of everything required to recreate the module in PK-sim
+      /// </summary>
+      public string Snapshot
+      {
+         set => _snapshot = value;
+         get => !IsPKSimModule ? string.Empty : _snapshot;
```
rwmcintosh, line 54: I left this as a hidden snapshot when the module is modified. If the user reverts the change, the snapshot will once again be available, if the project is closed and opened, then the snapshot mapping will be removed by the serialization cycle.

src/OSPSuite.Core/Domain/Services/ParameterValuesCreator.cs
rwmcintosh, line 1: Unrelated to the change. Just a cleanup

src/OSPSuite.Core/Serialization/Xml/ModuleXmlSerializer.cs
```diff
@@ -13,5 +14,21 @@ public override void PerformMapping()
...
+
+      protected override XElement TypedSerialize(Module objectToSerialize, SerializationContext context)
+      {
+         base.TypedSerialize(objectToSerialize, context);
+         var element =  base.TypedSerialize(objectToSerialize, context);
+         var snapShot = new XElement("Snapshot", objectToSerialize.Snapshot);
+         element.Add(snapShot);
+
+         return element;
+      }
```
rwmcintosh, line 32: Not sure here, When I add this as `Map(x => x.Snapshot)` then he whole base64 encoded string appears as an attribute.
This creates a separate element for it.

msevestre: you can specify in the mapping .AsNode or AsAttribute If my memories serves me whell

src/OSPSuite.Core/Domain/Module.cs
PavelBal: How about keeping it, so the user could "reset" the module even it has been converted from PK-Siom to extension?=

msevestre: I think the if is backward. IsPKsimModule ? _snapshot : string.empty

src/OSPSuite.Core/Serialization/Xml/ModuleXmlSerializer.cs
msevestre: also why do you care if it's an attribute or a node?

msevestre (commented):
I feel Map(x=>x.Snapshot).AsNode() should work

src/OSPSuite.Core/Domain/Module.cs
rwmcintosh: well, it's logically correct, but you mean why not remove the `!`

rwmcintosh: > How about keeping it, so the user could "reset" the module even it has been converted from PK-Siom to extension?=

Yeah, you can, as long as you have not re-opened. You cannot undo commands after the project is closed.

rwmcintosh: Oh, I get it. Like even if you have a module converted to extension, you could revert that even if the project had been saved

msevestre: > well, it's logically correct, but you mean why not remove the `!`

yes for readability. It's weird to start with a negation first especailly in a tertiary operator

msevestre: > How about keeping it, so the user could "reset" the module even it has been converted from PK-Siom to extension?=

This would not be enough to set the flag to true. It would have to be a "reload" module kind of action. But yeah keeping it does not hurt.

src/OSPSuite.Core/Serialization/Xml/ModuleXmlSerializer.cs
rwmcintosh: Theoretically never, in practice, all the time.

rwmcintosh:
> I feel Map(x=>x.Snapshot).AsNode() should work

In the end it did, but it was not as easy as it sounded

msevestre (approved):

msevestre:
but it's so much better now because we can reuse all the time :)

rwmcintosh:
> but it's so much better now because we can reuse all the time :)

always

PR #2537, opened 2025-05-29
Fixes #2536 batch edit for charts

### LINKED ISSUE #2536: Batch edit mode for charts
When users want to modify curves on a chart it can take a long time because each individual modification forces a chart update. For charts with many points, an update can take several seconds.

We should support a mode where users can disable chart updates, make many changes to curves, then update all at once.

- It could be automatically enabled for charts with > 20k? points
- It should be clear that the chart is out of date when the mode is used
- If enabled, then a user triggers it with any action from the curve editors or databrowser

rwmcintosh:
Fixes #2536

# Description
Some more optimizations and a new batch editing mode for charts

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

Enable batch mode
image

You can force the refresh when in batch mode.
image

Chart is blank until refresh, or batch mode disabled.
image


src/OSPSuite.UI/Binders/CurveBinder.cs
```diff
@@ -303,67 +304,73 @@ private void updateMarker(SimpleMarker marker)
 
       private void refreshData()
```
rwmcintosh, line 305: I'm not quite seeing the need for this method at all from the investigation I did. It feels like the datacolumns and curves are readonly in the charts. after initialization. I don't think a `CurveBinder` is ever updated with new data in the underlying datarepository.

Reducing the calls to this did not seem to affect the actual repaint time of the chart, but for curves of 10k it takes time to repopulate the datatable, and 10 curves of 10k when you really only changed 1 can be significant waste.

With this update this refresh is not called when editing curve properties, but it will still be called for other chart updates. But I don't think it's really needed.

src/OSPSuite.Presentation/Presenters/Charts/ChartDisplayPresenter.cs
```diff
@@ -557,20 +570,31 @@ public void ShowCurveInLegend(Curve curve, bool show)
...
       {
          if (!canHandle(chartUpdatedEvent))
             return;
 
-         updateChart();
+         if (Chart.BatchUpdateEnabled)
+         {
+            _curveBinders.Each(x => x.HideAllSeries());
+            View.ShowWatermark(Captions.RefreshChartToUpdate);
+            return;
```
rwmcintosh, line 594: In the batch update mode, when the first change is handled, the series are all hidden and the watermark drawn. That happens very fast. Any time I try to preserve the outdated chart with a  watermark, there's a significant price to pay. This is probably just the paint time when adding the watermark itself. The chart content is repainted.

I'm jus not sure it's acceptable that the chart is blank when you make batch changes. You don't have any outdated context... Maybe it's ok.

rwmcintosh:
@PavelBal is going to do some tests with this to see how usable it is. The batch mode blanks the chart for minimum repaint time.

src/OSPSuite.Presentation/Presenters/Charts/ChartDisplayPresenter.cs
rwmcintosh: yeah, the diagram is gone in this version when the batch mode is on, See the screenshots.

src/OSPSuite.UI/Binders/CurveBinder.cs
rwmcintosh: Hmm, better check what happens when you change the axis units for example. Might need to recalculate the data points in that case.

rwmcintosh: Yeah, you need it when there are scaling changes on the axes. Still, when we change curve properties we can skip this refresh.

msevestre:
I don’t think it should be context menu. Can it be a check box somewhere in the UI ? Even in the plot itself ? Or what do you think.  I would never think of going to this place to do batch update

rwmcintosh:
Yeah we could do something else, but currently there are no controls in the chart view.

image

I had considered ChartOptions, where you have SideMargins and other stuff for the whole chart, but that's a bit buried.
image

PavelBal:
This should be toggled from somewhere in the "Curves and Axis Settings". Here the suggestion:

image

- The checkbox "Auto update" is by default turned on. The behavior is as it used to be (each action updates the chart directly)
- When "Auto update" is on, the button "Apply" is disabled
- When "Auto update" is turned off, the "batch mode" applies. I.e., the chart is not updated - ideally the chart is still shown but with a watermark "Editing" or something
- As long "Auto update" Is turned off, the user has to press "Apply" to apply the changes, the watermark disappears until the user changes any setting again.

rwmcintosh:
What about this layout? Ignore the menu items in the bar because those are implemented per-application and this is just the view from the test application.

image

rwmcintosh:
I changed the wording to match 
image

Kept updated context menu items. I don't think it hurts to keep them.
image

src/OSPSuite.Core/Chart/CurveChartUpdater.cs
```diff
@@ -0,0 +1,47 @@
...
+
+      public CurveChartUpdater(IEventPublisher eventPublisher)
+      {
+         _eventPublisher = eventPublisher;
+      }
+
+      public CurveChartUpdate UpdateTransaction(CurveChart chart, bool curveDataChanged, Func<CurveChart, IReadOnlyCollection<Curve>> calculateModifiedCurves = null, bool propagateChartChangeEvent = true)
+      {
+         // default update all curves
+         if (calculateModifiedCurves == null)
```
msevestre, line 34: should that be called something else? curveToUpdateRetriever ?

src/OSPSuite.Core/Extensions/EnumerableExtensions.cs
```diff
@@ -50,5 +50,15 @@ public static EntityDescriptorMapList<T> ToEntityDescriptorMapList<T>(this IEnum
...
       }
+
+      /// <summary>
+      /// Returns the items that are not in both sets <paramref name="enumeration1"/> and <paramref name="enumeration2"/>
+      /// </summary>
+      public static IEnumerable<T> Complement<T>(this IEnumerable<T> enumeration1, IEnumerable<T> enumeration2)
+      {
+         var list1 = enumeration1.ToList();
+         var list2 = enumeration2.ToList();
+         return list1.Union(list2).Except(list1.Intersect(list2)).ToList();
```
msevestre, line 61: add unit tests for this

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -405,15 +406,31 @@ private void onChartPropertiesChanged()
...
+
+      /// <summary>
+      ///    Encapsulates a chart update where the curves that should be recalculated are the complement of the curves before and
+      ///    after the <paramref name="updateAction" />
+      /// </summary>
+      private void doAsCurveComplementUpdate(Action updateAction, bool curveDataChanged)
+      {
+         // We need ToList so that a new list is created in memory and not a reference to the existing Chart.Curves list
+         var existingCurves = Chart.Curves.ToList();
+         using (_chartUpdater.UpdateTransaction(Chart, curveDataChanged, x => x.Curves.Complement(existingCurves).ToList()))
```
msevestre, line 431: I Am not quite getting it. Isn't it then a setting of the chartUpdater to kmow to only update the one that are changing ? Feels like a state instead
updateMode: ALL, ADDED
curveDataChanged is always true except when remove so this can proably be inferred to. Not sure (maybe mode is REMOVED)

msevestre (changes requested):
it's a big PR and I don't quite understand the story specifically around two aspects
1: the curveDataChanged flag. It feels like this has to do with the fact that some settings was changed the requires some new bnding or sthg. The only instance of falce I could find was delete. is this correct?  curveDataChangd => i am expecing the underlying data to have changed which is not this at all
2: I don;t understand the whole complement stuff. It feels like an implementation details that seems to be all over now. Can we put it in the updater?

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
```diff
@@ -375,7 +377,7 @@ private void initPresentersWithColumnSettings()
 
       private void onAxisAdded()
       {
-         using (_chartUpdater.UpdateTransaction(Chart))
+         using (_chartUpdater.UpdateTransaction(Chart, curveDataChanged: true))
```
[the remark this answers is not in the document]
rwmcintosh: Yeah, I think it's probably the wrong name. I only found this to be needed in case of axis changes because the datatable has to be refreshed. It's the curve data at the DevExpress level that needs to be changed, not in the DataRepository level, so I guess the `Series` data has to change.

src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
rwmcintosh: I added states which trigger the ChartUpdater to calculate what has changed. Makes it easier to understand what's going on for sure.

We don't use the `Complement` anymore, but I left it since I wrote the tests first :)

msevestre (approved):

PR #2549, opened 2025-06-27
2543 pull up more snapshots and mappers for core objects 3

### LINKED ISSUE #2543: Pull up more snapshots and mappers for core objects

rwmcintosh:
Fixes #2543

# Description

Pulling up more mappers and snapshot objects


src/OSPSuite.Core/Domain/Builder/PathAndValueEntityBuildingBlockFromPKSim.cs
```diff
@@ -5,6 +5,7 @@ namespace OSPSuite.Core.Domain.Builder
    public abstract class PathAndValueEntityBuildingBlockFromPKSim<T> : PathAndValueEntityBuildingBlock<T> where T : PathAndValueEntity
    {
       public string PKSimVersion { set; get; }
+      public string SnapshotOriginModuleId { set; get; }
```
msevestre, line 8: Would it make sense to say that what this is and also explain udner which condition it's null?

msevestre (approved):
Only one little comment. Looks good to me. I would give some contextg around what the `SnapshotOriginModuleId` is about and when to use it and for what

PR #2608, opened 2025-08-27
Fixes #2607 Project conversion

### LINKED ISSUE #2607: Project conversion
v12.1 pkml is not compatible with 12.0, but you don't need an explicit conversion to move forward.

rwmcintosh:
Fixes #2607

# Description
We need a consistent chain of project converters where there's always a converter for a new PKML version. In this case, no active conversion is necessary since the pkml is forward compatible, but not backward compatible.

### DIFF
--- a/src/OSPSuite.Core/Converters/v12/Converter120To121.cs
+++ b/src/OSPSuite.Core/Converters/v12/Converter120To121.cs
@@ -11,12 +11,12 @@ namespace OSPSuite.Core.Converters.v12
 
       public (int convertedToVersion, bool conversionHappened) Convert(object objectToUpdate)
       {
-         return (PKMLVersion.V12_1, true);
+         return (PKMLVersion.V12_1, false);
       }
 
       public (int convertedToVersion, bool conversionHappened) ConvertXml(XElement element)
       {
-         return (PKMLVersion.V12_1, true);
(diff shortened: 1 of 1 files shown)


src/OSPSuite.Core/Converters/v12/Converter120To121.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+   public class Converter120To121 : IObjectConverter
+   {
+      // 12.1 pkml is not compatible with 12.0, but you don't need an explicit conversion to move forward.
+      // To satisfy the next converter, the object must pass through v12.1 conversion
+      public bool IsSatisfiedBy(int version) => version == PKMLVersion.V12_0;
+
+      public (int convertedToVersion, bool conversionHappened) Convert(object objectToUpdate)
+      {
+         return (PKMLVersion.V12_1, true);
```
msevestre, line 14: true or false? Why do we need to convert ?

msevestre (commented):
wondering about true vs false

src/OSPSuite.Core/Converters/v12/Converter120To121.cs
rwmcintosh: yeah, looks like it might be ignored anyway.

PR #2612, opened 2025-08-28
2603 behavior of includeexclude lists in passive transports

### LINKED ISSUE #2603: Behavior of include/exclude lists in Passive Transports
[PT_Overwrite behavior.zip](https://github.com/user-attachments/files/21994908/PT_Overwrite.behavior.zip)

The behavior is inconsistent and confusing.

In the project, there are two modules coming from PK-Sim - a PBPK model for sitagliptin, and a PBPK model for dapagliflozin. A simulation `Dapa_Sita` is the combination of these two modules.

THe PBPK model for Dapagliflozin has a passive transport `Glomerular Filtration-Generic-Dapagliflozin`:

Image

1. In the `Dapagliflozin_iv_PBPK` module (Overwrite mode), I change the "include" list of the transport to "All". The resulting simulation `Dapa_Sita_test1` generates this transport for both molecules Dapagliflozin and Sitagliptin, **as expected**.
2. Now I exclude "Sitagliptin" (simulation `Dapa_Sita_test2`), while having "All" checked. The transport is created for dapagliflozin only, **as expected**.
3. Now, while having "Dapagliflozin" in the "Include" list, but also checked the "All" check box, I try to add "Dapagliflozin" to the "Exclude" list:

Image

And get an error that the molecules cannot be in both lists. This is **unexpected**, because the "Include" list is grayed out and I assume that it is not considered.

4. I add an extension module "Dapagliflozin_Extension" (merge behavior "Overwrite") where the same passive transport is defined without any molecules in the "Include" list, simulation `Dapa_Sita_test3`. The transport for Dapa is still created, **unexpected**
5. Adding "Sitagliptin" to the "Include" list in the "Dapagliflozin_Extension" module. The transport is created for both molecules. **semi-expected**
6. Selecting "All" in the "Dapagliflozin_Extension" module. The transport is created for both molecules. **expected**
7. "All" but "Exclude" Sitagliptin - process created for Sita only. **Expected**
8. Selecting "All" but "Exclude" dapagliflozin - simulation creation fails. **Unexpected**

Image

So the behavior of "All" is always overwritten, but the include/exclude lists are extended. Confusing.

msevestre:
Fixes #2603 

# Description

- [x] And get an error that the molecules cannot be in both lists. This is unexpected, because the "Include" list is grayed out and I assume that it is not considered.

- [ ] I add an extension module "Dapagliflozin_Extension" (merge behavior "Overwrite") where the same passive transport is defined without any molecules in the "Include" list, simulation Dapa_Sita_test3. The transport for Dapa is still created, unexpected

## Type of change

# How Has This Been Tested?

- [x] Integration tests
- [x] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/OSPSuite.Core/Domain/Builder/SimulationBuilder.cs
+++ b/src/OSPSuite.Core/Domain/Builder/SimulationBuilder.cs
@@ -94,40 +94,42 @@ namespace OSPSuite.Core.Domain.Builder
 
       private void performMerge()
       {
-         mergeBuilders(x => x.Reactions, _reactions, mergeReactions);
-         mergeBuilders(x => x.Molecules, _molecules, mergeMolecules);
-         mergeBuilders(x => x.PassiveTransports, _passiveTransports, mergeTransports);
-         mergeBuilders(x => x.Observers, _observers, mergeObservers);
+         _reactions.AddRange(mergeBuilders(x => x.Reactions, mergeReactions));
+         _molecules.AddRange(mergeBuilders(x => x.Molecules, mergeMolecules));
+         _passiveTransports.AddRange(mergeBuilders(x => x.PassiveTransports, mergeTransports));
+         _observers.AddRange(mergeBuilders(x => x.Observers, mergeObservers));
          mergeParameterValueBuilders(x => x.SelectedParameterValues, _parameterValues);
          mergeInitialConditions();
          cacheEntities();
       }
 
-      private void mergeBuilders<T>(Func<Module, IBuildingBlock<T>> propAccess, CacheByName<T> cache, Action<T, BuilderSource<T>> mergeStrategyAction) where T : class, IBuilder, IEntity
+      private IReadOnlyList<T> mergeBuilders<T>(Func<Module, IBuildingBlock<T>> propAccess, Action<T, BuilderSource<T>> mergeStrategyAction) where T : class, IBuilder, IEntity
       {
          var analyzedMerges = analyzeBuilderMerges(propAccess);
-
+         var results = new List<T>();
          foreach (var mergeInfo in analyzedMerges)
          {
-            var (finalBuilder, buildingBlock) = mergeInfo.BaseBuilder;
+            var (baseBuilder, buildingBlock) = mergeInfo.BaseBuilder;
 
-            if (mergeInfo.RequiresClone)
+            if (mergeInfo.RequiresBaseClone)
             {
-               finalBuilder = cloneBuilder(finalBuilder);
-               foreach (var (sourceBuilder, sourceBuildingBlock) in mergeInfo.BuildersToMerge)
+               baseBuilder = cloneBuilder(baseBuilder);
+               foreach (var sourceBuilderThatExtend in mergeInfo.BuildersThatExtend)
                {
                   //Clone the source builder if needed to prevent cross-contamination during sequential merges
                   //When merging multiple builders, earlier merged builders could be affected by later merges through shared references
-                  var sourceBuilderToMerge = mergeInfo.RequiresMergeClone ? cloneBuilder(sourceBuilder) : sourceBuilder;
-                  var mergedBuilderSource = new BuilderSource<T>(sourceBuilderToMerge, sourceBuildingBlock);
-                  tryMergeContainers(finalBuilder, mergedBuilderSource);
-                  mergeStrategyAction(finalBuilder, mergedBuilderSource);
+                  var finalBuilderThatExtend = mergeInfo.RequiresExtensionClone ? cloneBuilder(sourceBuilderThatExtend.Builder) : sourceBuilderThatExtend.Builder;
+                  var finalSourceBuilderThatExtend = new BuilderSource<T>(finalBuilderThatExtend, sourceBuilderThatExtend.BuildingBlock);
+                  tryMergeContainers(baseBuilder, finalSourceBuilderThatExtend);
+                  mergeStrategyAction(baseBuilder, finalSourceBuilderThatExtend);
                }
             }
 
-            cache.Add(finalBuilder);
-            AddToBuilderSource(finalBuilder, buildingBlock);
+            results.Add(baseBuilder);
+            AddToBuilderSource(baseBuilder, buildingBlock);
          }
+
+         return results;
       }
 
       private T cloneBuilder<T>(T builder) where T : class, IBuilder
@@ -170,7 +172,7 @@ namespace OSPSuite.Core.Domain.Builder
          mergeDescriptorCriteria(target.TargetCriteria, source.TargetCriteria);
          target.CreateProcessRateParameter = source.CreateProcessRateParameter;
          target.ProcessRateParameterPersistable = source.ProcessRateParameterPersistable;
-         //TODO: do we need to clone?
+         //no need to clone the formula. it's either use as is or already a clone
          target.Formula = _cloneManager.Clone(source.Formula);
       }
 
@@ -327,12 +329,16 @@ namespace OSPSuite.Core.Domain.Builder
       internal class BuilderMergeInfo<T> where T : IBuilder
       {
          public BuilderSource<T> BaseBuilder { get; }
-         public IReadOnlyList<BuilderSource<T>> BuildersToMerge { get; }
+
+         /// <summary>
+         /// List of builders to EXTEND on top of the base builder
+         /// </summary>
+         public IReadOnlyList<BuilderSource<T>> BuildersThatExtend { get; }
 
          /// <summary>
          ///    Indicates that the base builder needs to be cloned before merging
          /// </summary>
-         public bool RequiresClone => BuildersToMerge.Count > 0;
+         public bool RequiresBaseClone => BuildersThatExtend.Count > 0;
 
          /// <summary>
          ///    It is required to also clone each builder being merged if we have 2 or more builders to merge.
@@ -343,7 +349,7 @@ namespace OSPSuite.Core.Domain.Builder
          ///    - If B1 wasn't cloned, these modifications propagate back to the original building block through shared references
          ///    With only 1 builder to merge, there are no subsequent merges to cause side effects, so cloning is not needed.
          /// </summary>
-         public bool RequiresMergeClone => BuildersToMerge.Count >= 2;
+         public bool RequiresExtensionClone => BuildersThatExtend.Count >= 2;
 
          public BuilderMergeInfo(BuilderSource<T> baseBuilder) : this(baseBuilder, Enumerable.Empty<BuilderSource<T>>())
          {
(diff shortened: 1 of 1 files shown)


msevestre (commented):
This is not reayd yet but I want to submit for review because this will impact a lot of things moving forward.

src/OSPSuite.Core/Domain/Builder/SimulationBuilder.cs
```diff
@@ -85,15 +93,102 @@ public IReadOnlyList<string> AllPresentEndogenousStationaryMoleculeNames() =>
...
+            }
+
+            cache.Add(finalBuilder);
+            AddToBuilderSource(finalBuilder, buildingBlock);
+         }
+      }
+
+      private T cloneBuilder<T>(T builder) where T : class, IBuilder
+      {
+         //we will need to make sure we keep the Id the same somehow to ensure proper tracking of entities. TBD
```
msevestre, line 134: I think that we will need to ensure that we KEEP The ids of the cloned entities. Because we use the id to find the element in the configuration but if we clone, it's gone! (this is to ensure we can Jump to Source only)

src/OSPSuite.Core/Domain/Builder/SimulationBuilder.cs
```diff
@@ -85,15 +93,102 @@ public IReadOnlyList<string> AllPresentEndogenousStationaryMoleculeNames() =>
...
+         var allEntities = sourceBuilder.GetAllChildren<IEntity>();
+         allEntities.Each(entity => AddToBuilderSource(entity, sourceBuildingBlock));
+
+         //at this step, all entities in the target should already be marked as coming from their respective source
+         _containerMergeTask.MergeContainers(target, sourceBuilder);
+      }
+
+      private void mergeReactions(ReactionBuilder target, BuilderSource<ReactionBuilder> source)
+      {
+         //TODO add reaction merge behavior
```
msevestre, line 161: @benjaperez1983 This is where the logic in reaction merge from #2640 should be implemented. Most of the code in this PR does not apply but we should talk about what can be saved anyways.

src/OSPSuite.Core/Domain/Builder/SimulationBuilder.cs
```diff
@@ -203,38 +266,131 @@ private void addMolecules(IMoleculeDependentBuilder builder, MoleculeList molecu
...
+         /// <summary>
+         ///    It is required to also clone each builder being merged if we have 2 or more builders to merge.
+         ///    This prevents cross-contamination between builders during sequential merges.
+         ///    Example: When merging B1, then B2 into finalBuilder:
+         ///    - Merge B1: B1's children/references are transferred to finalBuilder
+         ///    - Merge B2: B2's merge operations might modify children that came from B1
+         ///    - If B1 wasn't cloned, these modifications propagate back to the original building block through shared references
+         ///    With only 1 builder to merge, there are no subsequent merges to cause side effects, so cloning is not needed.
+         /// </summary>
+         public bool RequiresMergeClone => BuildersToMerge.Count >= 2;
```
msevestre, line 327: Hopefully this comments makes it clear

src/OSPSuite.Core/Domain/Builder/SimulationBuilder.cs
```diff
@@ -203,38 +266,131 @@ private void addMolecules(IMoleculeDependentBuilder builder, MoleculeList molecu
...
+            if (builders.Count == 1)
+            {
+               var builderSource = new BuilderSource<T>(builders[0].builder, builders[0].buildingBlock);
+               results.Add(new BuilderMergeInfo<T>(builderSource));
+               continue;
+            }
+
+            //last one is overwrite. We will use it as is also
+            var (lastBuilder, lastBuildingBlock, mergeBehavior) = builders[builders.Count - 1];
+            if (mergeBehavior == MergeBehavior.Overwrite)
```
msevestre, line 367: @PavelBal @rwmcintosh @Yuri05 Correct me if I am wrong but I believe that we said that in case of overwrite, we use as is? correct? I know I was disagreeing with some suggestions but I can't find trace of it in the discussion. 
Extend will defacto combine. But one overwrite and it becomes the new source of truth

src/OSPSuite.Core/CoreRegister.cs
```diff
@@ -92,6 +92,9 @@ public override void RegisterInContainer(IContainer container)
          container.Register<ISensitivityAnalysisRunner, SensitivityAnalysisRunner>(LifeStyle.Singleton);
          container.Register<IConfirmationManager, ConfirmationManager>(LifeStyle.Singleton);
 
+         //TODO no idea why I need to register this manually. I don't see where it's being excluded 
+         container.Register<SimulationBuilder, SimulationBuilder>();
```
msevestre, line 96: @rwmcintosh I don't understand why I need to do this. Any idea?

PavelBal:
@msevestre This is how I think the behaviors should work:

#### Merge behavior "Overwrite"

- The equation in the **Kinetic** tab is overwritten.

- The **Parameters** list is overwritten. This also implies that parameters that are not defined in the module that is lower in the hierarchy are removed.

- The **Operator** (and/or) for the "Source" and "Target" lists are overwritten.

- **Source** and **Target** lists are overwritten (i.e., it is possible to remove source/target condition).

- **Include/Exclude** molecule lists for molecules are overwritten. The behavior of the **All checkbox** is overwritten.
  #### Warning: was not implemented like this.

#### Merge behavior "Extend"
  ##### Warning: Was not implemented at all.

- The equation in the **Kinetic** tab is overwritten.

- The **Parameters** list is extended. That means that new parameters are added, but existing parameters are overwritten.

- The **Source** and **Target** lists are extended, i.e., new conditions are added. It is not possible to remove source/target conditions.

- The **Operator** (and/or) for the "Source" and "Target" lists are overwritten.

- **Include/Exclude** molecule lists for molecules are extended. However, the behavior of the **All checkbox** is overwritten.

- **Create process rate parameter** property is overwritten
- **Plot process rate parameter** property is overwritten

src/OSPSuite.Core/CoreRegister.cs
rwmcintosh: AssemblyScanner for utility includes this test that filters out classes without any interfaces

```
	private bool cannotRegister(Type type)
	{
		return type.IsInterface || type.IsAbstract || type.IsEnum || type.GetInterfaces().Length == 0;
	}
```

msevestre:
@all: A few opened questions but all in all, I think we have the good base for merging all entities. Once we approve, I'll update the event and @benjaperez1983 can adapt the reaction stuff

src/OSPSuite.R/Domain/Simulation.cs
```diff
@@ -50,7 +50,8 @@ public void AddAnalysis(ISimulationAnalysis simulationAnalysis)
       public Simulation(IModelCoreSimulation modelCoreSimulation)
       {
          CoreSimulation = modelCoreSimulation;
-         BuildConfiguration = new SimulationBuilder(Configuration);
+         //TODO!!
+         // BuildConfiguration = new SimulationBuilder(Configuration);
```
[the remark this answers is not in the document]
rwmcintosh: In OSPSuite-R these properties are used:


AllPresentFloatingMoleculeNames
AllPresentStationaryMoleculeNames
AllPresentXenobioticFloatingMoleculeNames
AllPresentEndogenousStationaryMoleculeNames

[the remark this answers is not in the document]
rwmcintosh: Where can we get the post-merge molecules to support those extensions? I guess we can just `GetAllChildren` with distinct names on the model for MoleculeAmounts? That will tell us all everything present, but to support the methods we would need to link back to the source molecule builder somehow?

tests/OSPSuite.HelpersForTests/ModuleHelperForSpecs.cs
```diff
@@ -72,12 +74,16 @@ public SimulationConfiguration CreateSimulationConfigurationForExtendMergeBehavi
...
+         // module1.Add(createEventGroupBuildingBlock("EventForModule1", "eventGroup1", "eventBuilder1"));
+         // module2.Add(createEventGroupBuildingBlock("EventForModule2", "eventGroup2", "eventBuilder1"));
+
          simulationConfiguration.AddModuleConfiguration(new ModuleConfiguration(module1));
-         // Using this constructor to validate that the merge behavior is set correctly
          simulationConfiguration.AddModuleConfiguration(new ModuleConfiguration(module2, null, null));
          return simulationConfiguration;
       }
 
+      //TODO REMOVE. This method is called extend and override at the same time!
```
rwmcintosh, line 86: I don't think it should be removed as per the TODO. The naming is off probably due to this method being introduced when we were able to override the module behavior when creating the simulation configuration. We removed that since.

I think that's the intended meaning of `Override` in the method name, the simulation configuration was overriding the module default behaviour.

You can just rename it. It's now simply a simulation configuration with an extend module.

src/OSPSuite.Core/Domain/Builder/SimulationBuilder.cs
rwmcintosh: Meaning this should only be used from the context of a warning or error message in the notifications pane?

src/OSPSuite.Core/Domain/Builder/SimulationBuilder.cs
```diff
@@ -203,38 +285,131 @@ private void addMolecules(IMoleculeDependentBuilder builder, MoleculeList molecu
...
+            builder = Builder;
+            buildingBlock = BuildingBlock;
          }
       }
 
-      internal BuilderSource BuilderSourceFor(IEntity sourceEntity) => _builderSources[sourceEntity.Id];
+      internal class BuilderMergeInfo<T> where T : IBuilder
+      {
+         public BuilderSource<T> BaseBuilder { get; }
+         public IReadOnlyList<BuilderSource<T>> BuildersToMerge { get; }
```
rwmcintosh, line 330: Or maybe 'BuilderMergeInfo` is the right name, but this should be `BuildersToExtendSource`.

src/OSPSuite.Core/Domain/Builder/SimulationBuilder.cs
rwmcintosh: I made some comments about a general sense I have when reading about merge/extend. Is there anywhere that the word **Extend** could replace **Merge**

rwmcintosh (commented):
Looks good to me. I just have a reservation about the names of some methods/properties/classes ... 

That can be summarized that I would prefer we use **Extend** over **Merge** whenever possible.

I might be the only one.

src/OSPSuite.Core/Domain/Builder/SimulationBuilderFactory.cs
```diff
@@ -0,0 +1,26 @@
+﻿using IoC = OSPSuite.Utility.Container.IContainer;
```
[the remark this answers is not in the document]
rwmcintosh: Yeah, that's what I mean. This makes it clear what container. Clear = Good

PR #2645, opened 2025-09-19
update version numbers and create new workflow names

rwmcintosh:
After release, we merge these updated numbers into develop to begin creating Core 12.2 NuGet packages


.github/workflows/coverage_12.2.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           dotnet restore
 
       - name: Build
-        run: dotnet build OSPSuite.Core.sln /p:Version=12.1.999
```
msevestre, line 22: what's with the 999? and why do we have 9999 on the other file?

msevestre (approved):

PR #2657, opened 2025-10-02
2656 unnecessary project has changed save changes prompt

### LINKED ISSUE #2135: Unnecessary `Project has changed. Save changes?` prompt
* Open the project attached (initially created with 12.1) [P1.zip](https://github.com/user-attachments/files/22472048/P1.zip)
* Open the only simulation
* Close the project

Security prompt `Project has changed. Save changes?` is shown - but nothing was changed in the project.

benjaperez1983:
Fixes Open-Systems-Pharmacology/MoBi#2135

# Description

When editing a simulation, an event was being fired that automatically set the flag ProjectHasChanged = true. This behavior was incorrect, because not every change during a simulation edit should be interpreted as a project-level modification.

Introduced a scoped suppression mechanism to prevent ProjectHasChanged from being set during simulation editing. 
The suppression is only active while the simulation is being loaded or edited, ensuring that normal changes outside of this context continue to correctly set the flag.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
+++ b/src/OSPSuite.Presentation/Presenters/Charts/ChartEditorPresenter.cs
@@ -406,7 +406,7 @@ namespace OSPSuite.Presentation.Presenters.Charts
 
       private void onDataBrowserUsedChanged(UsedColumnsEventArgs e)
       {
-         using (_chartUpdater.UpdateTransaction(Chart, e.Used ? CurveChartUpdateModes.Add : CurveChartUpdateModes.Remove))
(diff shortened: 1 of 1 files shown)


msevestre (changes requested):
@benjaperez1983 This feels weird to me to have suddently a loading flag being passed throughout the code TBH. We have a way of dealing with this using some BatchaCtions so that we never have to do some manual ++ and -- like suggested
Also I Am wondering if we are not solving the problem by patching something that should not happen in the first place. Let's discsus tomorrow during the dev meeting but for now, this code is too hacky for what it fixes

PR #2659, opened 2025-10-06
2658   snapshot based runners seem to only use a single core

### LINKED ISSUE #2658: Snapshot based runners seem to only use a single core
When fixing [this issue](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3229) for population simulations. It seemed like running individual simulations was always using a single core and never truly parallelizing async/await.

For that issue, the population runner does parallelize all the individual simulations, but that doesn't mean that all separate individual simulations are also parallelized. We have to make another adjustment somewhere.

benjaperez1983:
Fixes #2658

# Description

Prepare thread safe for running parallel simulations.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/OSPSuite.CLI.Core/MinimalImplementations/EntityValidationTask.cs
+++ b/src/OSPSuite.CLI.Core/MinimalImplementations/EntityValidationTask.cs
@@ -8,20 +8,20 @@ namespace OSPSuite.CLI.Core.MinimalImplementations
 {
    public class EntityValidationTask : IEntityValidationTask
    {
-      private readonly IEntityValidator _entityValidator;
+      private readonly IEntityValidatorFactory _entityValidatorFactory;
       private readonly IOSPSuiteExecutionContext _executionContext;
       private readonly IOSPSuiteLogger _logger;
 
-      public EntityValidationTask(IEntityValidator entityValidator, IOSPSuiteExecutionContext executionContext, IOSPSuiteLogger logger)
+      public EntityValidationTask(IEntityValidatorFactory entityValidatorFactory, IOSPSuiteExecutionContext executionContext, IOSPSuiteLogger logger)
       {
-         _entityValidator = entityValidator;
+         _entityValidatorFactory = entityValidatorFactory;
          _executionContext = executionContext;
          _logger = logger;
       }
 
       public bool Validate(IObjectBase objectToValidate)
       {
-         var validationResult = _entityValidator.Validate(objectToValidate);
+         var validationResult = _entityValidatorFactory.Create().Validate(objectToValidate);
          if (validationResult.ValidationState == ValidationState.Valid)
             return true;
 
--- a/src/OSPSuite.CLI.Core/MinimalImplementations/EntityValidationTask.cs
+++ b/src/OSPSuite.CLI.Core/MinimalImplementations/EntityValidationTask.cs
@@ -21,7 +21,7 @@ namespace OSPSuite.CLI.Core.MinimalImplementations
 
       public bool Validate(IObjectBase objectToValidate)
       {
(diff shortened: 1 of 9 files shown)


rwmcintosh (changes requested):
I'm thinking that we should remove the statefulness one level up by creating `EntityValidatorFactory` to create new `EntityValidator` for each thread. Basically what you have here, but up one level in the call stack. It would be less code changes as well.

There are usages of the interface in PK-Sim that would have to be changed as well so that the factory is used.

src/OSPSuite.Core/Domain/Services/EntityValidator.cs
```diff
@@ -12,24 +12,17 @@ namespace OSPSuite.Core.Domain.Services
...
+   ///    its children as well.
+   ///    This class is not thread-safe. See <seealso cref="IEntityValidatorFactory" /> to create thread instances
    /// </summary>
-   public interface IEntityValidator
-   {
-      /// <summary>
-      ///    Validates the given entity (for a container, all its children as well).
-      /// </summary>
-      ValidationResult Validate(IObjectBase objectBase);
-   }
```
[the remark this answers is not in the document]
msevestre: I don't understand why we remove this. The facotyr and the service are two dfferent things

src/OSPSuite.CLI.Core/MinimalImplementations/EntityValidationTask.cs
```diff
@@ -8,20 +8,20 @@ namespace OSPSuite.CLI.Core.MinimalImplementations
...
-         _entityValidator = entityValidator;
+         _entityValidatorFactory = entityValidatorFactory;
          _executionContext = executionContext;
          _logger = logger;
       }
 
       public bool Validate(IObjectBase objectToValidate)
       {
-         var validationResult = _entityValidator.Validate(objectToValidate);
+         var validationResult = _entityValidatorFactory.Create().Validate(objectToValidate);
```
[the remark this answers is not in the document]
msevestre: let's add a method called `Validate` that creates and validate I agree

tests/OSPSuite.Core.IntegrationTests/ContextForIntegration.cs
```diff
@@ -105,6 +105,7 @@ public void RegisterComponents(CastleWindsorContainer container)
          IoC.RegisterImplementationOf(IoC.Container);
 
          container.WindsorContainer.AddFacility<TypedFactoryFacility>();
+         container.Register<IEntityValidatorFactory, EntityValidatorFactory>();
```
msevestre, line 108: Why are we registering this by hand?

msevestre (changes requested):
I don't undertand why we removed the interface. I coud see that wone would potentially want to validate PKSim spewcific object and now we can't
Also how do we should that there is an issue now that the presnetation stuff has been removed.
Last, it looks like there are still usage of presenter in the tests that are ghost.

tests/OSPSuite.Core.IntegrationTests/ContextForIntegration.cs
rwmcintosh: this is only in tests right? I didn't think the scan was used here.

msevestre: hum weird. I don't know why we would register everything by hand in our integrationstests. Kinds of defeat the purpose of integration somehow lol. Forget it then :)

rwmcintosh:
added the interface back so it can be injected
Added the presentation stuff back in tests

PR #2722, opened 2025-12-15
Fixes #2721 Path of a parameter of a Passive Transport cannot be resolved when "Create process rate" is enabled

### LINKED ISSUE #2721: Path of a parameter of a Passive Transport cannot be resolved when "Create process rate" is enabled
[PassiveTransportPathBug.zip](https://github.com/user-attachments/files/24126415/PassiveTransportPathBug.zip)

In the attached project, see the passive transport `Glomerular Filtration-GFR-Alfentanil`.

The transport has a parameter `GFR fraction` that is a local parameter (as all parameters of PT are local by default).

In the `Kinetics` definition of the transport, these parameters are usually referenced with the path that only contains the name of the parameter, e.g. `GFR fraction`.

However, it is also possible to define the path to the parameter as `<TransportName>|<ParameterName>`, e.g., `Glomerular Filtration-GFR-Alfentanil|GFR fraction`

Image

The simulation can be created and works. However, when trying to create the same simulation with the option "Create transport rate parameters", the path `Glomerular Filtration-GFR-Alfentanil|GFR fraction` cannot be resolved.

So either a) simulation creation should already fail (and it would be good to understand why), or the creation of `ProcessRate´ parameter should not fail.

rwmcintosh:
Fixes #2721

# Description
The path resolver can resolve paths that start with the container name, but the process rate parameter creation will not create the correct path adjustment for the process rate parameter.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/OSPSuite.Core/Domain/Mappers/ProcessRateParameterCreator.cs
+++ b/src/OSPSuite.Core/Domain/Mappers/ProcessRateParameterCreator.cs
@@ -57,19 +57,16 @@ namespace OSPSuite.Core.Domain.Mappers
          if (!objectPath.Any())
             return;
 
+         // if the path starts with the  process name then we need to replace the name with ".."
+         // to compensate. That's because a container name can be ignored when a path is resolved from within that container
+         if (objectPath.Count == 2 && string.Equals(objectPath[0], processName))
+            objectPath.RemoveAt(0);
+
          // if the path starts with ".."  or if the objectPath only has the name of the reference then it
          // should be adjusted to have an additional ".." at the front
          if (objectPath[0] == ObjectPath.PARENT_CONTAINER || objectPath.Count == 1)
          {
             objectPath.AddAtFront(ObjectPath.PARENT_CONTAINER);
-            return;
-         }
-
-         // if the path contains two elements and the first one is the process name then we need to move up two containers at the front
-         // to compensate. That's because a container name can be ignored when a path is resolved from within that container
-         if (objectPath.Count == 2 && string.Equals(objectPath[0], processName))
-         {
-            objectPath.AddAtFront(new ObjectPath(ObjectPath.PARENT_CONTAINER, ObjectPath.PARENT_CONTAINER));
          }
       }
    }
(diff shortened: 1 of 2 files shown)


tests/OSPSuite.Core.Tests/Domain/ProcessRateParameterCreatorSpecs.cs
```diff
@@ -49,6 +50,8 @@ protected override void Context()
          _kinetic.AddObjectPath(_formulaUsablePathFU);
          _formulaUsablePathBW = new FormulaUsablePath("Organism", "BW").WithAlias("BW");
          _kinetic.AddObjectPath(_formulaUsablePathBW);
+         _formulaUsablePathC = new FormulaUsablePath("Reaction", "C").WithAlias("C");
```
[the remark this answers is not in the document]
msevestre: oh that' sthe name of the reaction defined below. This is super confusiong. Can we set the name higer up

I don't understand why the path look like this
why would the path in the process rate parmaeter be "P1|C". Should not that be "..|C"? or sthg like this
It feels like it;s absolute relative to the parent container. Do we have sthg like this anyewhere else?

msevestre (commented):
a bit confused with the use cas.e Implementation is fine

src/OSPSuite.Core/Domain/Mappers/ProcessRateParameterCreator.cs
```diff
@@ -49,27 +50,27 @@ public IParameter CreateProcessRateParameterFor(IProcessBuilder processBuilder,
...
+            objectPath.AddAtFront(ObjectPath.PARENT_CONTAINER);
+            return;
+         }
 
-         return false;
+         // if the path contains two elements and the first one is the process name then we need to move up two containers at the front
+         // to compensate. That's because a container name can be ignored when a path is resolved from within that container
+         if (objectPath.Count == 2 && string.Equals(objectPath[0], processName))
+         {
+            objectPath.AddAtFront(new ObjectPath(ObjectPath.PARENT_CONTAINER, ObjectPath.PARENT_CONTAINER));
```
rwmcintosh: @Yuri05 what do you think? should we make this case `..|..|ProcessName|ParameterName` or remove the process name and go with `..|ParameterName`

Yuri05: `..|ParameterName` would be better for sure (and process name invariant)

PR #2764, opened 2026-02-04
Fixes #1863 remove check for monotonically increasing time values?

### LINKED ISSUE #1863: `DataRepository` - remove check for monotonically increasing time values?
What is the reason for this check? Is it really required? It often leads to problems when importing data (e.g. reported values are not sorted in the correct way).

rwmcintosh:
Fixes #1863 remove check for monotonically increasing time values?

# Description
Relatively few fires created from what I can see.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

## Summary by CodeRabbit

* **User Experience**
  * Improved chart tooltip detection and resolution for more accurate tooltips.

* **Bug Fixes**
  * Interpolation now averages values across matching grid points for more consistent results.

* **Breaking Changes**
  * Removed bulk data-manipulation and public mutation APIs (insert/swap).
  * Grids no longer enforce strict increasing order; duplicates/non-monotonic values accepted.
  * A specialized "time not strictly monotone" exception and sheet-name enrichment on import errors were removed.


src/OSPSuite.Core/Domain/Data/DataColumn.cs
```diff
@@ -232,11 +234,29 @@ public virtual float GetValue(float baseValue)
...
+         BaseGrid.Values.Each((x, i) =>
+         {
+            if (x != baseGridAtIndex) 
+               return;
+            accumulator += Values[i];
+            count++;
+         });
+
+         return accumulator / count;
+      }
```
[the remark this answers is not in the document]
rwmcintosh: we want to use exact matches, not tolerance based matches

tests/OSPSuite.Core.Tests/Domain/DataColumnSpecs.cs
```diff
@@ -118,34 +128,34 @@ public void TestAddRelatedColumn()
...
 
       [Observation]
       public void TestAddRelatedColumnArithmeticStdDevWithWrongDimension()
       {
-         sut.Values = new[] {8.0F, 0.0F, -5.0F};
+         sut.Values = new[] { 8.0F, 0.0F, -5.0F, 0.0F };
          var relatedColumn = new DataColumn("Regina", _dimensionless, _baseGrid);
          relatedColumn.DataInfo = new DataInfo(ColumnOrigins.ObservationAuxiliary, AuxiliaryType.ArithmeticStdDev, " ", "", 0);
-         relatedColumn.Values = new[] {0.1F, 0.2F, 0.1F};
+         relatedColumn.Values = new[] { 0.1F, 0.2F, 0.1F, 02F };
```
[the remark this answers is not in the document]
rwmcintosh: fixed

src/OSPSuite.Core/Domain/Data/DataColumn.cs
```diff
@@ -218,10 +219,11 @@ public virtual float GetValue(float baseValue)
...
          int index = BaseGrid.IndexOf(baseValue);
          if (index >= 0)
-            return Values[index];
+            return averageFor(index);
 
-         int leftIndex = BaseGrid.LeftIndexOf(baseValue);
-         int rightIndex = BaseGrid.RightIndexOf(baseValue);
+         // We can throw an exception here if the values are not monotonic?
+         int leftIndex = BaseGrid.IndexOfNextLowest(baseValue);
+         int rightIndex = BaseGrid.IndexOfNextHighest(baseValue);
```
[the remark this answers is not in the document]
rwmcintosh: removed

src/OSPSuite.Core/Domain/Data/BaseGrid.cs
```diff
@@ -35,18 +32,6 @@ public override IReadOnlyList<float> Values
...
-            {
-               if (value[i] <= value[i - 1])
-               {
-
-                  var beforeValue =this.ConvertToDisplayUnit(value[i - 1]);
-                  var afterValue = this.ConvertToDisplayUnit(value[i]);
-                  throw new TimeNotStrictlyMonotoneException(beforeValue, afterValue, DisplayUnit.Name, Repository?.Name);
-               }
-            }
-
```
Yuri05, line 49: should we at least check that the base grid values are **sorted**?
I think even in case of non-unique base grid we should sort the values before.

Otherwise, if when e.g. plotting observed data not only by dots but by **dots+line** - it could become ugly if the time values are not sorted. Also the `IndexOfNextXXX `routines below would be simpler.

Btw: use cases where not unique values (and not sorted values for sure) could make trouble are linear interpolations of data points, and those use cases must be checked once we allow non unique values:

* import of ph dependent solubility table

Image

* import of formulation table

Image

* Import data points of arbitrary table formula in MoBi

Image

Yuri05: yeah, sorted is a MUST. 
Non unique values in table formula with activated "Use derivated values" will produce NaNs whatsoever pretty sure

Yuri05: The more I think about it: non-unique values in all cases above must be forbidden.
It's only **observed data** (and maybe few other cases) where it could be allowed.
Any use case where table represents a **function** `y=f(BaseGrid)` should support only unique (and sorted) values.

Yuri05: @PavelBal Thoughts?

rwmcintosh: Shouldn't that then be controlled outside the base grid object? By the presenters or dtos of these particular usages? Or by another base grid type that does enforce.

rwmcintosh: You can't sort the the basegrid independent of the value columns. They are linked by order.

rwmcintosh: > Shouldn't that then be controlled outside the base grid object? By the presenters or dtos of these particular usages? Or by another base grid type that does enforce.

I bet that's already the case because we wouldn't rely on the exception to validate these controls. There must be rules in place that enforce that. We can verify that of course but I'm pretty sure the validation fails already if you try duplicates or wrong order.

rwmcintosh: When plotting observed data we could make this ordering when creating the series points for the chart, only for observation repositories

Yuri05: > Shouldn't that then be controlled outside the base grid object? By the presenters or dtos of these particular usages?

Uniqueness must be controlled outside because it's case dependent.
Sorting should be **done** outside as well, but if we agree that base grids should always be sorted, than it can be **controlled** here.

Yuri05: > You can't sort the the basegrid independent of the value columns. They are linked by order.

Yes, that's why the sorting must be done outside. But here we can assume that passed base grids are always sorted, and then control it here.

Yuri05: > When plotting observed data we could make this ordering when creating the series points for the chart, only for observation repositories

It's better to sort once during the data import - then no need to sort every time later when the data is plotted, exported, whatsoever. 

So in principle I would keep the code in _BaseGrid.cs_ almost AS IS with the only exception that it does not check anymore that the values are _unique_ (but still checks that they are sorted).

PavelBal:
There are two tasks here:

- Allowing duplicates of x values. This should only be allowed for observed data, I agree with @Yuri05 
- Allowing import of non-monotonically ascending data. This should be easily fixed by simply sorting the grid during import.

src/OSPSuite.Core/Domain/Data/DataColumn.cs
```diff
@@ -218,23 +219,46 @@ public virtual float GetValue(float baseValue)
...
+         if (indexOfNextLowest < 0)
             return float.NaN;
 
-         if (rightIndex >= BaseGrid.Count)
+         if (indexOfNextHighest >= BaseGrid.Count)
             return float.NaN;
 
-         if (leftIndex == rightIndex)
+         if (indexOfNextLowest == indexOfNextHighest)
             return float.NaN;
```
Yuri05: can be simplified into just one if statement:
```
if (indexOfNextLowest < 0 || indexOfNextHighest >= BaseGrid.Count || indexOfNextLowest == indexOfNextHighest)
   return float.NaN;
```

src/OSPSuite.Core/Domain/Data/DataColumn.cs
```diff
@@ -218,23 +219,46 @@ public virtual float GetValue(float baseValue)
...
+                (averageFor(indexOfNextHighest) - averageOfNextLowest) * (baseValue - BaseGrid[indexOfNextLowest]) /
+                (BaseGrid[indexOfNextHighest] - BaseGrid[indexOfNextLowest]);
+      }
+
+      /// <summary>
+      /// There could be multiple instances of the same base grid value, so we need to average them
+      /// </summary>
+      /// <param name="index">the index of the basegrid value that we want to create an average for (eg t=0)</param>
+      /// <returns>The mean average of all values that have the same basegrid value as the <paramref name="index"/></returns>
+      private float averageFor(int index)
```
Yuri05: would pass directly the base grid value rather than its index as argument.
With the current implementation, base grid value for index is repeatedly calculated here and in the calling `GetValue `function.

src/OSPSuite.Core/Domain/Data/DataColumn.cs
```diff
@@ -218,23 +219,46 @@ public virtual float GetValue(float baseValue)
...
+
+         BaseGrid.Values.Each((x, i) =>
+         {
+            if (x != baseGridAtIndex) 
+               return;
+            accumulator += Values[i];
+            count++;
+         });
+
+         return accumulator / count;
```
Yuri05: I wonder if for **multiple** values for the same base grid (time) we should return a **median** rather than mean
@PavelBal

PavelBal: This is for observed data? Then we should return all individual values. Or what is the use case of returning an aggregation?

Yuri05: GetValue (single!) is used everywhere (both for observed data and for all other use cases).
If we need to retrieve all values somewhere (for observed data only) then we would need additional `GetValues()` function (not a replacement of the current one)
E.g. see here: https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/446963b867c3c0774db201b6a0d90f1e010d469e/src/PKSim.Core/Services/CompoundAlternativeTask.cs#L199-L214

src/OSPSuite.Core/Domain/Data/BaseGrid.cs
rwmcintosh: I'm not sure how much we gain by enforcing it. We can have the importer sort the points to make the charts less cluttered if lines are enabled, but interpolating values that are not at basegrid points will still require averaging.

All the removed insert/insert range stuff is not used anywhere in Suite anyway.

src/OSPSuite.Core/Domain/Data/DataColumn.cs
rwmcintosh: It's for interpolating non-basegrid points in any basegrid or data column.

PavelBal: Then... median vs mean. Median probably makes more sense, but I guess users intuitively expect mean. I would return mean, but the decision also depends on the intended use.

As we only allow multiple values per one time grid for the observed data, is there actually a use case where an interpolation would be required?

Yuri05: > Then... median vs mean. Median probably makes more sense, but I guess users intuitively expect mean. I would return mean, but the decision also depends on the intended use.

OK, let's return mean then for now. I think the use case "non unique obs data" will be pretty rare anyways

> As we only allow multiple values per one time grid for the observed data, is there actually a use case where an interpolation would be required?

For obs data I think no.

src/OSPSuite.Core/Domain/Data/BaseGrid.cs
Yuri05: >We can have the importer sort the points

yep, this should be done imo

rwmcintosh: ok, what I mean is, I'll leave out the Exception throwing for non-increasing. Both uniqueness and increasing base grid are handled by the construction of the base grid and is not enforced by the base grid.

src/OSPSuite.Core/Domain/Data/DataColumn.cs
rwmcintosh: I changed this around to reduce the number of times we iterate over the base grid. I changed the method `IndexOf` to return multiple indexes for a value. That way we only have to iterate once to average.

Yuri05 (approved):

src/OSPSuite.Core/Domain/Data/BaseGrid.cs
Yuri05: ok, let's try it like this and see what happens :)

tests/OSPSuite.Core.Tests/Domain/BaseGridSpecs.cs
```diff
@@ -174,32 +73,32 @@ public void TestRemove()
...
 
-         Assert.AreEqual(1, sut.LeftIndexOf(0.5F));
-         Assert.AreEqual(1, sut.LeftIndexOf(0.6F));
-         Assert.AreEqual(3, sut.LeftIndexOf(1.6F));
-         Assert.AreEqual(-1, sut.LeftIndexOf(-3F));
+         Assert.AreEqual(1, sut.IndexOfNextLowest(0.5F));
+         Assert.AreEqual(1, sut.IndexOfNextLowest(0.6F));
+         Assert.AreEqual(3, sut.IndexOfNextLowest(1.6F));
+         Assert.AreEqual(-1, sut.IndexOfNextLowest(-3F));
       }
```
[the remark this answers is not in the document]
rwmcintosh: I will open a new issue for all NUnit style assertions.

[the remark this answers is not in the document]
rwmcintosh: I opened it and generalized it a little. There are other places where NUnit is used directly

https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/2766

PR #2772, opened 2026-02-13
Fixes #2771 move individualcharacteristics class to core

### LINKED ISSUE #2771: Move `IndividualCharacteristics` class to Core
In order for `MoBi.R` to build `IndividualBuildingBlock`s from PK-Sim `Individual`s we want to be able to pass this configuration mapping object from R, to MoBi and then to PK-Sim.

Needs to go into Core so MoBi knows about it.

rwmcintosh:
Fixes #2771 move individualcharacteristics class to core

# Description
Moves some configuration mapping classes from PK-Sim to core so they can be accessed by MoBi.

## Type of change

- [x] Other (please describe):

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [ ] Manual tests 
- [x] No tests required

# Reviewer checklist

# Questions (if appropriate):

## Summary by CodeRabbit

* **New Features**
  * Expanded capture of individual characteristics: species, population, gender, age, weight, height, and a seed value.
  * Added support for disease-state metadata and disease-state parameters.
  * Added tracking of calculation method identifiers for results provenance.
  * Introduced molecule ontogeny associations to link molecules with ontogenies for richer profiling.


src/OSPSuite.R/Domain/IndividualCharacteristics.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Collections.Generic;
+using OSPSuite.Core.Snapshots;
+
+namespace OSPSuite.R.Domain;
+
+/// <summary>
+///    Wrapper object for .net that encapsulates origin data and disease state
+/// </summary>
+public class IndividualCharacteristics : OriginData
```
rwmcintosh, line 9: Stuck for a naming scheme. We will have an object in PK-Sim called `IndividualCharacteristics` that predates this, inherits it, and adds `MoleculeOntogeny`

Should we rename the PK-Sim object and call it `IndividualCharacteristicsWithOntogeny` or something

Yuri05: You mean this one?
~https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/develop/src/PKSim.R/Domain/IndividualCharacteristics.cs~
No this is the one that was moved, right?

rwmcintosh: I'm going to pull up most of it into Core because I want to use it in MoBi. However, through MoBi you will get only the `IndividualBuildingBlock` and not ontogeny parameters. So we should not include it in the object.

For PK-Sim we have to continue to support that because those ontegny path and value parameters are used to change values in a simulation.

rwmcintosh: You just haven't seen this part yet because I need phase one first (this PR)

Yuri05: OK, I'll then wait for that PR and answer it there :)
>Should we rename the PK-Sim object and call it IndividualCharacteristicsWithOntogeny or something

Yuri05 (approved):

src/OSPSuite.R/Domain/IndividualCharacteristics.cs
msevestre: why can't MoleculeOntogeny be a parameter of this object and not set ?

msevestre (changes requested):
I think it's ok to have a class that has optional parameters instead of creating a complex hierarchy of class. Why would we not add the MoleculeOntogeny in core?

src/OSPSuite.R/Domain/IndividualCharacteristics.cs
rwmcintosh: > OK, I'll then wait for that PR and answer it there :)
> 
> > Should we rename the PK-Sim object and call it IndividualCharacteristicsWithOntogeny or something

I need this one first

rwmcintosh:
> I think it's ok to have a class that has optional parameters instead of creating a complex hierarchy of class. Why would we not add the MoleculeOntogeny in core?

I can change it, but I think this is confusing if someone adds them and expects them to do something.

PR #2857, opened 2026-05-07
Fixes #2838 Distributed parameter overwrite loses distribution

### LINKED ISSUE #2838: Distributed parameter loses distribution when value is overwritten and parameter is not in spatial structure
## Problem

When mapping a `ParameterValue` (typically an `IndividualParameter`) back to a model `IParameter` via `ParameterValueToParameterMapper`, a distributed parameter whose value has been overwritten is turned into a plain non-distributed parameter.

Current logic in [ParameterValueToParameterMapper.MapFrom](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Domain/Mappers/ParameterValueToParameterMapper.cs):

```csharp
var parameter = !parameterValue.IsDistributed() || parameterValue.Value != null
   ? _parameterFactory.CreateParameter(name, dimension: dimension, displayUnit: displayUnit)
   : _parameterFactory.CreateDistributedParameter(name, distributionType.Value, dimension: dimension, displayUnit: displayUnit);
```

When `IsDistributed() && Value != null`, a non-distributed parameter is created — losing the distribution. The value overwrite should produce a distributed parameter with `IsFixedValue = true`, not a constant parameter.

This bug surfaces in PK-Sim as [Open-Systems-Pharmacology/PK-Sim#3511](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3511): creating a simulation from an individual where a distributed parameter was overwritten yields a constant parameter instead of a distributed one with an overridden value. Introduced in v12 (commit afdb080f, Fixes #2295).

## Expected

If the `ParameterValue` is distributed, always create an `IDistributedParameter`. If a value is also provided, set it on the distributed parameter afterward (which sets `IsFixedValue = true` and keeps the distribution formula).

## Reproduction

* Create all standard building blocks in PK-Sim.
* In the individual, overwrite one distributed volume parameter.
* Create a simulation with this individual.

Observe that the parameter in the simulation is constant, not distributed.

msevestre:
## Summary
- Keep a parameter distributed when an individual parameter provides both `DistributionType` and `Value` (the value is then applied as a fixed value while the distribution formula and sub-parameters remain intact).
- `ParameterFactory.CreateDistributedParameter` now pre-creates the sub-parameters referenced by the distribution formula (`Mean`, `Deviation`, `Percentile`, ...) so a distributed parameter created from an individual building block is fully operational even when not present in the spatial structure.
- Adds integration tests covering both cases (overwrite of an existing distributed parameter, and a distributed parameter only defined in the individual building block — reproduces Open-Systems-Pharmacology/PK-Sim#3511).

## Test plan
- [ ] `OSPSuite.Core.Tests` pass
- [ ] `OSPSuite.Core.IntegrationTests` pass (new `ModelConstructorIntegrationTests` cases)
- [ ] Verify in PK-Sim that the originally reported scenario (PK-Sim#3511) is resolved

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when individual values overwrite distributed parameters so distribution status, fixed flags, and overwritten values are preserved.
  * Ensure required distribution sub-parameters (including percentile) are created and percentiles are refreshed when distributions become fixed.

* **Tests**
  * Added integration tests covering distributed-parameter overwrite scenarios, including cases created outside the existing spatial structure.

### DIFF
--- a/src/OSPSuite.Core/Domain/Mappers/ParameterValueToParameterMapper.cs
+++ b/src/OSPSuite.Core/Domain/Mappers/ParameterValueToParameterMapper.cs
@@ -21,25 +21,17 @@ namespace OSPSuite.Core.Domain.Mappers
       public IParameter MapFrom(ParameterValue parameterValue)
       {
          var name = parameterValue.Name;
+         var value = parameterValue.Value;
          var dimension = parameterValue.Dimension;
          var displayUnit = parameterValue.DisplayUnit;
          var distributionType = parameterValue.DistributionType;
 
-         //if a distribution is defined we create a distributed parameter; the overwritten value (if any) is
-         //applied as a fixed value via the IDistributedParameter Value setter, which keeps the distribution
-         //formula intact and updates the percentile sub-parameter accordingly.
+         //if a distribution is defined we create a distributed parameter; the value (if any) is applied as a
+         //fixed value while the distribution formula and its sub-parameters are kept intact.
          //Otherwise, we create a plain parameter whose constant formula carries the value.
-         IParameter parameter;
-         if (parameterValue.IsDistributed())
-         {
-            parameter = _parameterFactory.CreateDistributedParameter(name, distributionType.Value, dimension: dimension, displayUnit: displayUnit);
-            if (parameterValue.Value.HasValue)
-               parameter.Value = parameterValue.Value.Value;
-         }
-         else
-         {
-            parameter = _parameterFactory.CreateParameter(name, value: parameterValue.Value, dimension: dimension, displayUnit: displayUnit);
-         }
+         var parameter = parameterValue.IsDistributed()
+            ? _parameterFactory.CreateDistributedParameter(name, distributionType.Value, value: value, dimension: dimension, displayUnit: displayUnit)
+            : _parameterFactory.CreateParameter(name, value: value, dimension: dimension, displayUnit: displayUnit);
 
          parameter.ValueOrigin.UpdateAllFrom(parameterValue.ValueOrigin);
 
--- a/src/OSPSuite.Core/Domain/Mappers/ParameterValueToParameterMapper.cs
+++ b/src/OSPSuite.Core/Domain/Mappers/ParameterValueToParameterMapper.cs
@@ -1,6 +1,7 @@
-﻿using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Core.Extensions;
 using OSPSuite.Utility;
+
 
 namespace OSPSuite.Core.Domain.Mappers
 {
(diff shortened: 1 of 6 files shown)


msevestre:
@coderabbitai full review please. I have a pro license

msevestre:
@coderabbitai where is my full review?

src/OSPSuite.Core/Domain/Mappers/ParameterValueToParameterMapper.cs
```diff
@@ -20,19 +21,21 @@ public ParameterValueToParameterMapper(IParameterFactory parameterFactory)
       public IParameter MapFrom(ParameterValue parameterValue)
       {
          var name = parameterValue.Name;
+         var value = parameterValue.Value;
          var dimension = parameterValue.Dimension;
          var displayUnit = parameterValue.DisplayUnit;
          var distributionType = parameterValue.DistributionType;
 
-         //if the distribution is undefined or the value is set, we create a default parameter to ensure that the value will take precedence.
-         //Otherwise, we create a distributed parameter and assume that required sub-parameters will be created as well
```
msevestre, line 28: @rwmcintosh @Yuri05 This was the wrong assumption here. During the mapping from Modules, this was NOT done and hence we would lose the structure

src/OSPSuite.Core/Domain/ParameterFactory.cs
```diff
@@ -111,12 +113,38 @@ public IParameter CreateDistributedParameter(string name, DistributionType distr
...
+      }
+
+      private void addDistributionSubParametersTo(IDistributedParameter distributedParameter, IDimension dimension)
+      {
+         if (distributedParameter.Formula is DistributionFormula distributionFormula)
+         {
+            distributionFormula.ObjectPaths
+               .Select(path => path.Last())
+               .Where(subName => distributedParameter.GetSingleChildByName(subName) == null)
+               .Each(subName => distributedParameter.Add(CreateParameter(subName, dimension: dimension)));
```
msevestre, line 143: Ensure about this here as we are losing some context on the parameters but we do not have anythig at this stage besides the fact that the parmaeters are required by the formula.

Yuri05 (approved):

Yuri05:
> I have a pro license

Does it always work as simple as this? I will try it out 😆

Yuri05:
@coderabbitai What is this new "Review Change Stack" button, which you are now adding to your reviews, is about? Is it described in the CodeRabbit-documentation somewhere? I could not find it. When clicking on the button, I either get a 404 Error, or a prompt from CodeRabbit to extend its permissions, which I declined (e.g. I don't want to give all the permissions to the private repositories, which I was asked for)

> [Review Change Stack](https://app.coderabbit.ai/change-stack/Open-Systems-Pharmacology/OSPSuite.Core/pull/2857)
>

src/OSPSuite.Core/Domain/Services/QuantityValuesUpdater.cs
```diff
@@ -94,11 +94,30 @@ private IReadOnlyList<string> allMoleculeNames(SimulationConfiguration configura
...
          //Order by distribution to ensure that distributed parameter are loaded BEFORE their sub parameters.
          //note: use descending otherwise parameter without distribution are returned fist
-         valueUpdater.ModelConfiguration.SimulationConfiguration.Individual?.OrderByDescending(x => x.DistributionType)
-            .Each(addOrUpdateParameter);
+         individual.OrderByDescending(x => x.DistributionType).Each(addOrUpdateParameter);
+
+         //A distributed parameter's value (and percentile) is applied before its sub-parameters
+         //(Mean/Deviation/...) are set above. Refresh the percentile of any distributed parameter that
+         //came from the individual so it reflects the now-correct sub-parameter values.
+         refreshPercentilesOf(valueUpdater, individual);
```
msevestre, line 109: @Yuri05 Added change here to ensure that percentile is showing the right value
We have a bit of a weird situation. We need Parent Parameter (volume) to be defined first to add Mean and Std etc. But we need Mean and STd to be deined to calcualte value ofr Parameter. 
By refreshing the percentile after the fact, we are good

Issue #2861, opened 2026-05-08
Keep R interfaces dependent on OSPSuite.Core only (let Presentation stay Windows-only)


rwmcintosh:
## Why (revised direction)

The original framing of this issue was *"restore `OSPSuite.Presentation` to cross-platform `net10.0`"*, which would have required refactoring `DxValidatableDTO` off `DevExpress.XtraEditors.DXErrorProvider.IDXDataErrorInfo` along with a coordinated multi-repo migration of every `DxValidatableDTO`-derived DTO in PK-Sim (17 files) and MoBi (14 files). After landing #2858 / #2863, that scope feels excessive — `OSPSuite.Presentation` is genuinely a WinForms/DevExpress MVP layer and there is little ongoing value in pretending it is cross-platform.

**Revised goal:** accept that `OSPSuite.Presentation` (and the downstream `PKSim.Presentation` / `MoBi.Presentation`) are honestly `net10.0-windows`, and instead make sure the *R* interfaces only need `OSPSuite.Core` — no Presentation in their transitive closure.

That way Linux/macOS R users get a real cross-platform path through `OSPSuite.Core` services, without anyone having to maintain the fiction that Presentation can be cross-platform.

## Concrete next steps

1. Walk the transitive package/project reference chain from `OSPSuite.R` and confirm what (if anything) still drags `OSPSuite.Presentation` into the closure today. (Currently `OSPSuite.R.csproj` directly references `OSPSuite.Core`, `OSPSuite.Infrastructure`, `OSPSuite.Infrastructure.Autofac`, `OSPSuite.Infrastructure.Import`, `OSPSuite.CLI.Core`, `OSPSuite.Assets` — none reference Presentation directly. Confirm with `dotnet nuget why src/OSPSuite.R/OSPSuite.R.csproj OSPSuite.Presentation`.)
2. For any service / type living in `OSPSuite.Presentation` that R-bound code calls into, move it down to `OSPSuite.Core` (or to `OSPSuite.Infrastructure` if it has non-Core infrastructure deps).
3. After the moves, `OSPSuite.Presentation` is allowed to permanently target `net10.0-windows`. Drop the temporary comments in the four csproj files added by #2863 (`OSPSuite.Presentation`, `OSPSuite.Presentation.Serialization`, `OSPSuite.Presentation.Tests`, `OSPSuite.Core.IntegrationTests`).
[...]

PR #2863, opened 2026-05-11
Fixes #2858 Refactor ApplicationIcon to remove Windows-only code

### LINKED ISSUE #2858: Refactor ApplicationIcon to remove Windows-only code from OSPSuite.Assets.Images
## Why

`OSPSuite.Assets.Images` advertises a cross-platform TFM (`net10.0`) but has always pulled in Windows-only code:

- `using System.Drawing;` — `Image`, `Bitmap` are Windows-only on .NET 7+ unless `System.Drawing.UseWindowsLibrary` is opted in.
- `using DevExpress.Utils.Svg;` — `SvgImage` and the `SvgImage.Render(...)` extension live in DevExpress assemblies that target `net8.0-windows` from 25.x onward.

Today this works on Ubuntu/macOS via `OSPSuite.R` only because the suite carefully avoids calling into icon code paths from R. The DevExpress 25.2.7 upgrade in the V13 .NET 10 work makes the fiction visible: with the proper `net8.0-windows`-tagged DevExpress packages, restoring `OSPSuite.Assets.Images` (net10.0) emits NU1701:

> Package 'DevExpress.Utils 25.2.7' was restored using '.NETFramework,Version=v4.6.1, ...' instead of the project target framework 'net10.0'.

The same warning cascades up through `OSPSuite.Infrastructure.Import` → `OSPSuite.R`, etc.

## Goal

`OSPSuite.Assets.Images` should be a pure resource library: SVG bytes + lookup, no Windows API surface. Rendering belongs in `OSPSuite.UI` (already `net10.0-windows`).

After the refactor:
- `OSPSuite.Assets.Images` truly compiles and runs on `net10.0` on any OS.
- `OSPSuite.Infrastructure.Import`, `OSPSuite.Presentation`, `OSPSuite.R`, `OSPSuite.Presentation.Serialization`, `OSPSuite.CLI.Core`, `OSPSuite.Infrastructure.Export` keep their cross-platform `net10.0` TFM honestly.
- The Linux/macOS R surface stops depending on "no one calls icon code" as a load-bearing assumption.
- NU1701 disappears (no DevExpress assembly is dragged through a non-Windows TFM).

## Suggested approach

Files that need changes (all in `OSPSuite.Core/src/OSPSuite.Assets.Images/`):

- `ApplicationIcon.cs` — strip `System.Drawing` and `DevExpress.Utils.Svg` references. Keep the class as a holder for the SVG `byte[]` + `IconName` + `Index`. Drop the `_image`/`SvgImage` field, the `ToImage(...)` overloads, and the implicit `operator Image` / `operator SvgImage`.
- `ApplicationIcons.cs` — drop `using DevExpress.Utils.Svg;`. Replace `private static ApplicationIcon createEmptyIcon() => new ApplicationIcon((SvgImage)null);` with a `byte[]`-based no-op constructor.
- Add a new helper class in `OSPSuite.UI` (e.g. `OSPSuite.UI/Mappers/ApplicationIconExtensions.cs`) that owns the rendering side:
  - `SvgImage ToSvgImage(this ApplicationIcon icon)` — constructs `SvgImage` from `icon.SvgBytes` on demand.
  - `Image ToImage(this ApplicationIcon icon, IconSize size = null)` — does the `SvgImage.Render(size, ...)` call.
  - Optional: explicit conversion helpers replacing the old implicit `Image`/`SvgImage` operators (the implicit operators were ergonomic but they're the actual mechanism by which Windows-only types leaked across the assembly boundary; explicit makes the call site visible).
- `OSPSuite.UI/Mappers/ApplicationIconsToImageCollectionMapper.cs` already lives in `OSPSuite.UI` and is the right home for any "build a DevExpress.SvgImageCollection from all icons" logic.
- Remove `<PackageReference Include="DevExpress.Data" Version="25.2.7" />` and `<PackageReference Include="DevExpress.Utils" Version="25.2.7" />` from `OSPSuite.Assets.Images.csproj`.

## Call-site impact

Anywhere that does `(Image)icon` or `(SvgImage)icon` today needs to become `icon.ToImage()` / `icon.ToSvgImage()` (and import the extensions from `OSPSuite.UI`). All such call sites already live in WinForms/DevExpress UI code, so the new dependency is a no-op for them.

The implicit operators + property `_image` field are convenience layers that have grown to ~250 call sites across the suite (rough grep). The mechanical update is what makes this a real piece of work; the architectural change itself is small.

## Definition of done

- `OSPSuite.Assets.Images.csproj` no longer references DevExpress packages, and the project compiles standalone with TFM `net10.0` (no `-windows`).
- `OSPSuite.Core.sln` builds clean with no NU1701 warnings.
- Existing unit tests pass (`OSPSuite.Presentation.Tests/Presentation/ApplicationIconSpecs.cs` is the main spec).
- A smoke load of `OSPSuite.R` on a Linux .NET 10 runtime succeeds (it should, since `OSPSuite.R` only ever calls back into icon code through UI which is never loaded from R).

## Tracking

Downstream tracking issues will be linked from MoBi and PK-Sim once filed.

rwmcintosh:
## Summary

- `OSPSuite.Assets.Images` is now a pure cross-platform resource library — `ApplicationIcon` holds an SVG `byte[]` + name/index only, no `System.Drawing` or `DevExpress.Utils` references. DevExpress package refs are dropped from the csproj.
- Rendering (`ApplicationIcon` → `SvgImage` / `Image`) lives in `OSPSuite.UI/Extensions/ApplicationIconExtensions.cs` with a per-icon `SvgImage` cache (parse-once, identical to the prior `_image` field memoization — just relocated).
- `OSPSuite.Presentation`, `OSPSuite.Presentation.Serialization`, `OSPSuite.Presentation.Tests`, and `OSPSuite.Core.IntegrationTests` move to `net10.0-windows` — accepted as their honest target. `OSPSuite.Presentation` is a WinForms/DevExpress MVP layer; pretending it's cross-platform required a much larger `DxValidatableDTO` refactor without delivering proportional value. Cross-platform R is solved separately — `OSPSuite.R` already has zero transitive dependency on `OSPSuite.Presentation` (verified via `dotnet nuget why`), so nothing in OSPSuite.Core needs to change. The remaining work to keep `PKSim.R` / `MoBi.R` honestly cross-platform lives in those repos (see Follow-ups below).
- All ~15 call sites of the previous implicit `ApplicationIcon → SvgImage` / `IconSize → System.Drawing.Size` operators updated to explicit `.ToSvgImage()` / `.ToDrawingSize()` / `.ToImage()` calls.
- Drive-by security fix: NPOI's transitive `System.Security.Cryptography.Xml 8.0.2` (CVE-2026-33116 / [GHSA-37gx-xxp4-5rgx](https://github.com/advisories/GHSA-37gx-xxp4-5rgx)) is overridden to `10.0.7` in `OSPSuite.Infrastructure.Import` — matches the existing override in `OSPSuite.Infrastructure.Export`. Clears all NU1903 warnings across the solution.

## Follow-ups

The original framing planned a coordinated three-repo refactor to return `OSPSuite.Presentation` to cross-platform. That direction was dropped after verifying `OSPSuite.R` is already Presentation-free — see #2861 (closed not-planned). The remaining R-cross-platform work is downstream-only:

- PK-Sim — [Open-Systems-Pharmacology/PK-Sim#3537](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3537): `PKSim.R` pulls `OSPSuite.Presentation` via `PKSim.Infrastructure`. Decouple by moving R-bound services down to `PKSim.Core` (or splitting `PKSim.Infrastructure`).
- MoBi — [Open-Systems-Pharmacology/MoBi#2388](https://github.com/Open-Systems-Pharmacology/MoBi/issues/2388): same pattern via `MoBi.CLI.Core` / `MoBi.Infrastructure`.

Companion consumption PRs against this work are open in PK-Sim ([#3533](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3533)) and MoBi ([#2385](https://github.com/Open-Systems-Pharmacology/MoBi/issues/2385)) — the local cross-repo integration test passed: PK-Sim and MoBi build clean against locally-packed `13.0.0-pbkny` packages produced from this branch.

## Test plan

- [x] `dotnet build OSPSuite.Core.sln` — clean (0 errors, no NU1701 from DevExpress, no NU1903). Remaining NU1701 warnings are pre-existing for `Northwoods.GoWin 5.2.0` in OSPSuite.UI / Starter / UI.Tests.
- [x] `dotnet test tests/OSPSuite.Core.Tests` — 2021 / 2021 pass
- [x] `dotnet test tests/OSPSuite.Presentation.Tests` — 1247 / 1247 pass
- [x] `dotnet test tests/OSPSuite.UI.Tests` — 65 / 65 pass (includes 4 new specs for `ApplicationIconExtensions` covering null-bytes fallback + `SvgImage` cache hit)
- [x] `dotnet test tests/OSPSuite.Core.IntegrationTests` — 501 / 502 pass (1 pre-existing skip)
- [x] `dotnet nuget why src/OSPSuite.R/OSPSuite.R.csproj OSPSuite.Presentation` — no path. `OSPSuite.R` cross-platform claim is honest.
- [x] PK-Sim and MoBi solutions build clean against this branch's local NuGet output (see companion PRs)
- [ ] Visual smoke of `OSPSuite.Starter` (reviewer)

## Summary by CodeRabbit

* **Bug Fixes**
  * Addressed security vulnerability in cryptography library with CVE patch.

* **Improvements**
  * Enhanced Windows platform compatibility with framework updates.
  * Optimized icon rendering through intelligent caching mechanisms.
  * Upgraded core dependencies including SQLite and security libraries for improved stability.


.github/workflows/build-and-publish_v13.yml
```diff
@@ -30,6 +27,8 @@ jobs:
           echo "APP_VERSION=13.0.${{ github.run_number }}" | Out-File -FilePath $env:GITHUB_ENV -Append
 
       - name: Build
+        env:
+          DevExpress_License: ${{ secrets.DEV_EXPRESS_KEY }}
```
rwmcintosh, line 31: Unrelated change that was implemented after the .NET 10 was already merged for Core

.github/workflows/build-and-publish_v13.yml
```diff
@@ -39,6 +38,7 @@ jobs:
         run: dotnet pack .\OSPSuite.Core.sln --no-build --no-restore -o ./ -p:PackageVersion=${{env.APP_VERSION}} --configuration=Debug --no-build
 
       - name: Push test log as artifact
+        if: always()
```
rwmcintosh, line 41: Unrelated change that was implemented after the .NET 10 was already merged for Core

src/OSPSuite.Assets.Images/ApplicationIcon.cs
```diff
@@ -1,45 +1,15 @@
...
-      {
-         return _image?.Render(imageSize, paletteProvider: null, useHighSpeedRendering: DefaultBoolean.Default, allowCache: DefaultBoolean.Default)
-                ?? new Bitmap(imageSize.Width, imageSize.Height);
-      }
-
-      public virtual SvgImage ToSvgImage() => _image;
    }
-}
\ No newline at end of file
+}
```
rwmcintosh, line 15: We're just keeping IconBytes, the Index and the IconName. The static caching is still done, but it's moved to a `OSPSuite.UI` layer extension method

src/OSPSuite.Assets.Images/IconSize.cs
```diff
@@ -1,28 +1,21 @@
...
-        public string Id => $"{Width}x{Height}";
+      internal IconSize(int width, int height)
+      {
+         Width = width;
+         Height = height;
+      }
 
-        public static implicit operator Size(IconSize icon)
-        {
-            return new Size(icon.Width, icon.Height);
```
rwmcintosh, line 20: That used to return a `System.Drawing.Size` but it's Windows only so we're just going to use `int, int` in this layer now

src/OSPSuite.Presentation.Serialization/OSPSuite.Presentation.Serialization.csproj
```diff
@@ -1,7 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net10.0</TargetFramework>
+    <!-- Inherits net10.0-windows from OSPSuite.Presentation. See note in OSPSuite.Presentation.csproj. -->
+    <TargetFramework>net10.0-windows</TargetFramework>
```
rwmcintosh, line 5: I'm hoping to clean this one up next if necessary. Using `-windows` clears up the warnings on build right now, but it will not allow us to use the presentation assembly in cross platform uses so we have to clear it next.

src/OSPSuite.Presentation/OSPSuite.Presentation.csproj
```diff
@@ -2,7 +2,13 @@
 
 
   <PropertyGroup>
-    <TargetFramework>net10.0</TargetFramework>
+    <!--
+      OSPSuite.Presentation targets net10.0-windows because it currently leaks Windows-only types
+      (System.Drawing.Bitmap/Image/Icon/StringAlignment) and DevExpress validation interfaces
+      (IDXDataErrorInfo via DxValidatableDTO base class). Cleaning these up is tracked in follow-up
+      issues for OSPSuite.Core, MoBi, and PK-Sim. Once those land, this can move back to net10.0.
```
rwmcintosh: yeah, like this

src/OSPSuite.UI/Extensions/ApplicationIconExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+
+      public static Image ToImage(this ApplicationIcon icon, IconSize imageSize)
+      {
+         var svg = icon.ToSvgImage();
+         return svg?.Render(imageSize.ToDrawingSize(), paletteProvider: null, useHighSpeedRendering: DefaultBoolean.Default, allowCache: DefaultBoolean.Default)
+                ?? new Bitmap(imageSize.Width, imageSize.Height);
+      }
+
+      public static Size ToDrawingSize(this IconSize iconSize) => new(iconSize.Width, iconSize.Height);
+   }
```
rwmcintosh, line 36: This is the new static SVG caching. SVG's are still only rendered once.

src/OSPSuite.UI/Extensions/ApplicationIconExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+            return null;
+
+         if (!_svgCache.Contains(icon))
+         {
+            using var ms = new MemoryStream(icon.IconBytes);
+            _svgCache[icon] = new SvgImage(ms);
+         }
+
+         return _svgCache[icon];
+      }
```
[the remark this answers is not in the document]
rwmcintosh: Reverted to `ConcurrentDictionary<ApplicationIcon, SvgImage>` + `GetOrAdd` in b7b19e5a — agreed, the `Cache`/`Contains` swap re-introduced the check-then-act race.

src/OSPSuite.UI/Extensions/ApplicationIconExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+
+      public static Image ToImage(this ApplicationIcon icon) => icon.ToImage(IconSizes.Size16x16);
+
+      public static Image ToImage(this ApplicationIcon icon, IconSize imageSize)
+      {
+         var svg = icon.ToSvgImage();
+         return svg?.Render(imageSize.ToDrawingSize(), paletteProvider: null, useHighSpeedRendering: DefaultBoolean.Default, allowCache: DefaultBoolean.Default)
+                ?? new Bitmap(imageSize.Width, imageSize.Height);
+      }
```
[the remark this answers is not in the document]
rwmcintosh: Declining this one. A few reasons:

1. **Not a regression.** The pre-refactor `ApplicationIcon.ToImage(IconSize)` had the identical body — `return _image?.Render(...) ?? new Bitmap(imageSize.Width, imageSize.Height);` (see V13 `src/OSPSuite.Assets.Images/ApplicationIcon.cs`). This PR preserves the disposal contract precisely; any leak that exists today existed before #2858 and is out of scope.

2. **The fallback path is rare.** It's only hit when `icon.IconBytes == null` — i.e., `ApplicationIcons.EmptyIcon` on icons that didn't resolve. Not the hot path for the GDI-handle math here.

3. **Caching the fallback Bitmap introduces a real disposal-contract hazard.** A shared static `Bitmap` per `IconSize` means any caller that disposes the returned image (which WinForms controls routinely do on their own `Dispose`) would tear it out from under every other caller. That's a worse footgun than allocating a fresh blank bitmap on the cold path.

4. **The actual GDI churn this PR's callers see is from `SvgImage.Render(...)`**, which always allocates a fresh `Bitmap` — by DevExpress design. The dominant UI path (toolbars, tabs, tree views, menus) doesn't hit `.ToImage()` per call at all; it goes through `ImageListRetriever` which pre-renders all icons into DevExpress `SvgImageCollection`s once at app start.

5. **If `ValueOriginBinder.imageFor` is leaking, the fix belongs at the call site** (`using` the temporary `sourceImage` / `methodImage` once they've been drawn onto the composite, or copying to a `SvgImageCollection`-managed slot). That's its own change unrelated to the Assets.Images refactor.

Happy to file a separate issue for GDI hygiene in `ValueOriginBinder` if it shows up in profiling, but rolling it into this PR would broaden scope without fixing the actual root cause.

src/OSPSuite.UI/Extensions/BarItemExtensions.cs
```diff
@@ -18,7 +18,7 @@ public static BarItem UpdateIcon(this BarItem barItem, ApplicationIcon icon)
 
       private static SvgImage transparentImage(ApplicationIcon icon, IconSize iconSize)
       {
-         return icon;
+         return icon.ToSvgImage();
       }
```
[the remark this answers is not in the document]
rwmcintosh: done

src/OSPSuite.Infrastructure.Import/OSPSuite.Infrastructure.Import.csproj
```diff
@@ -28,6 +28,8 @@
     <PackageReference Include="NPOI" Version="2.7.6" />
     <PackageReference Include="OSPSuite.Utility" Version="5.0.0.7" />
     <PackageReference Include="LumenWorksCsvReader" Version="4.0.0" />
+    <!-- Override NPOI's transitive System.Security.Cryptography.Xml 8.0.2, which has CVE-2026-33116 (GHSA-37gx-xxp4-5rgx). Patched in 10.0.6+. -->
```
msevestre, line 31: @rwmcintosh what's this?

src/OSPSuite.UI/Services/ToolTipCreator.cs
```diff
@@ -37,8 +37,8 @@ public SuperToolTip CreateToolTip(string content, string title = null, Applicati
             setupArgs.Title.Text = title;
 
          setupArgs.Contents.Text = convertHtml(content);
-         setupArgs.Contents.ImageOptions.SvgImage = imageToUse;
-         setupArgs.Contents.ImageOptions.SvgImageSize = IconSizes.Size16x16;
+         setupArgs.Contents.ImageOptions.SvgImage = imageToUse.ToSvgImage();
```
msevestre: let's creeate an extgension for ImageOptions that is going to to do the two steps ? feel.s like we have this in may places

src/OSPSuite.Infrastructure.Import/OSPSuite.Infrastructure.Import.csproj
rwmcintosh: We need that update to avoid a security notice. It's a transitive dependency from NPOI. We're just updating it to avoid the vulnerability

src/OSPSuite.UI/Views/ExceptionView.cs
```diff
@@ -61,7 +61,7 @@ public string Description
       public void Initialize(string caption, ApplicationIcon icon, string issueTrackerUrl, string productName)
       {
          Text = caption;
-         IconOptions.SvgImage = icon;
+         IconOptions.SvgImage = icon.ToSvgImage();
```
msevestre, line 64: no size here for some reason

msevestre (changes requested):

src/OSPSuite.UI/Services/ToolTipCreator.cs
rwmcintosh: done

src/OSPSuite.UI/Views/ExceptionView.cs
rwmcintosh: that one doesn't need a size

msevestre (approved):

PR #2878, opened 2026-06-02
Remove dead code in OSPSuite.Core

rwmcintosh:
## Summary

Removes dead code from **OSPSuite.Core** — unused types and members with no remaining callers anywhere in the suite. This is a pure cleanup: the change set is **deletions only** (no behavioral change), confined entirely to OSPSuite.Core.

There is no linked issue for this PR, and that is intentional.

## How the dead code was identified

1. **ReSharper `inspectcode`** (the same engine as the IDE) was run over a combined solution containing OSPSuite.Core **plus MoBi and PK-Sim** source-linked together, so that cross-repo usage of Core's public API was visible to the analyzer. We focused on the `UnusedMember.Global` and `UnusedType.Global` rules.
2. **Semantic confirmation with Roslyn.** Every candidate was confirmed dead with `SymbolFinder.FindReferencesAsync` (zero real source references), not text search — common names such as `Refresh`, `Clear`, and `BestFitColumns` collide with DevExpress/framework members, so a grep-based count is unreliable.
3. **Iterated to stability.** After each removal batch we re-ran `inspectcode` (removing dead code can orphan its callees) and repeated until the analyzer reported nothing new in Core.

## Using the registration conventions to find *true* dead code

ReSharper flags a member as unused when it sees no direct reference, but much of the suite is wired by **convention-based DI and reflection**, which static analysis cannot see. Our understanding of these registration channels was essential to separate genuinely dead code from live code that merely *looks* unused:

- **`I<Name>` interface convention** — a class `Foo` is auto-registered under `IFoo` and injected by interface, so the concrete type is never referenced by name. If a flagged class implements a live `I<Name>` interface (checked transitively, including interfaces defined in OSPSuite.Utility), it is **used**, not dead.
- **Group / marker resolution** — serializers, `*DiffBuilder`s, and `*ContextMenuFactory`s are discovered by reflection scans and resolved as a set (`ResolveAll<T>` / `IRepository<T>`). For example, all 19 flagged `*ContextMenuFactory` classes are live through this channel and were kept.
- **Lifecycle / eventing** — `IStartable` repositories, `IListener<T>` event handlers, and command/query mediators are invoked by the container or the event facility, never by name.

Only candidates that survived **both** ReSharper and this semantic DI audit were removed.

## Surfaces we deliberately avoided

- **The R reflection surface.** OSPSuite.R exposes .NET APIs to R scripts via rSharp **reflection** (invocation by method-name string), which no static analysis can observe. We computed the R closure — the transitive project references of `OSPSuite.R` + `OSPSuite.CLI.Core`, i.e. `OSPSuite.Core`, `OSPSuite.Assets`, and the `OSPSuite.Infrastructure.*` projects — and treated public members there as off-limits unless we could **positively verify the member name is absent from the entire OSPSuite-R package** (since rSharp calls by name, an absent name cannot be called). The R-reachable public domain API was left untouched.
- **Icons.** `ApplicationIcons` resolves icons by **string name** at runtime (`IconByName`, name-keyed overlays, etc.). Hundreds of icon entries look unused to the analyzer but are live, and deleting them would break the UI at runtime with no build or test failure. The icon registry was excluded entirely.

## Judgment calls (kept on purpose)

Some genuinely-unreferenced members were **kept** where removing them would break a natural symmetry and the code is trivial — e.g. a `SetFalse` paired with a used `SetTrue`, or a single overload within a small overload set. API completeness/symmetry is a reasonable contract that a raw reference count does not capture.

## Verification

- `dotnet build` of the combined Core + MoBi + PK-Sim solution: **0 errors**.
- Unit tests green: **OSPSuite.Core 2030/2030**, **OSPSuite.Presentation 1255/1255**, **OSPSuite.Infrastructure 158/158**.
- Loading the app, running simulations, PI, SA, reviewing charts

## Follow-up: test-only (Tier C) + closed-cluster pass

A second sweep extends this PR with two detections the first pass did not cover:

- **Tier C (test-only).** Re-ran `inspectcode` with all test projects removed and diffed against the with-tests run, surfacing production code referenced *only* by specs. After the same DI / group-resolution rescue and an OSPSuite-R name grep (clean for every candidate), the genuinely-dead types were `DataRowData` and `HistoryBrowserProperties` (the latter only a vestigial test fake).
- **Closed clusters.** A Roslyn tool detects mutually-referencing presenter/view and dead-duplicate clusters that per-type analysis cannot flag, because the `I<name>` rescue marks each member "used" via the other. Found: `EditDescriptionPresenter` and `ExportParameterIdentificationToRUICommand` (dead duplicates -- the live versions are in PK-Sim) and `RunMultipleParameterIdentificationsUICommand` (not wired to any menu).

The cluster scan also flagged false positives that the verification rejected and that were **kept**: `OptionsEditorView` (its view interface is co-located with the live `IOptionsEditorPresenter`, injected by `MappingParameterEditorPresenter`/`LloqEditorPresenter`), `ExceptionManager`/`ProgressManager` (registered under their OSPSuite.Utility interfaces), and `DataFormatHeadersWithUnits`/`DataFormatNonmem` (`ResolveAll<IDataFormat>`).

## Release Notes

* **Removed Features**
  * Removed R export functionality for parameter identification analysis.

* **Improved UI**
  * Simplified chart export and templating workflows.
  * Refined importer column mapping interface for improved clarity.

* **Updated Labels**
  * Updated user-facing captions and error messages throughout the application for consistency and clarity.


Yuri05:
> 1. **ReSharper `inspectcode`** (the same engine as the IDE) was run over a combined solution containing OSPSuite.Core **plus MoBi and PK-Sim** source-linked together, so that cross-repo usage of Core's public API was visible to the analyzer. We focused on the `UnusedMember.Global` and `UnusedType.Global` rules.

What about further consumers of OSPSuite.Core (InstallationValidator, QualificationRunner)?

rwmcintosh:
> > 1. **ReSharper `inspectcode`** (the same engine as the IDE) was run over a combined solution containing OSPSuite.Core **plus MoBi and PK-Sim** source-linked together, so that cross-repo usage of Core's public API was visible to the analyzer. We focused on the `UnusedMember.Global` and `UnusedType.Global` rules.
> 
> What about further consumers of OSPSuite.Core (InstallationValidator, QualificationRunner)?

Yeah, there was one over-removal that was used by QualificationRunner. It's restored now.

Yuri05 (approved):

PR #2882, opened 2026-06-12
Fixes #2431 Chart templates for more curve charts

### LINKED ISSUE #2431: Chart templates for PI time profiles
There isn't any support for templates in PI time profiles for chart templates.

rwmcintosh:
﻿Fixes #2431

There isn't any support for chart templates in parameter identification charts. Simulation time profile charts offer a chart template menu (apply / create from current chart / update existing / manage), but the PI analyses do not.

## Scope

- All PI curve-chart analyses get the chart template menu:
  - Time Profile
  - Time Profile Confidence Interval
  - Time Profile Prediction Interval
  - Time Profile VPC Interval
  - Predicted vs Observed
  - Residuals vs Time

  Since a PI spans multiple simulations, templates are sourced from the **project** (the same source the multi-simulation comparison charts use).
- The simulation-level analyses also newly get the chart template menu, sourced from the **simulation** (the same source as the simulation time profile chart):
  - Predicted vs Observed
  - Residuals vs Time
- Because a template pool can now mix templates created from charts with different axes (time profile vs predicted vs observed vs residuals vs time), templates are discriminated by chart type:
  - `CurveChart` gets a `CurveChartType` (enum `CurveChartTypes`: `TimeProfile`, `PredictedVsObserved`, `ResidualVsTime`), promoted from the former abstract `AnalysisChart.AnalysisChartType`. The PI interval analyses (Confidence/Prediction/VPC Interval) are time profile charts with the same time-based axes, so they classify as `TimeProfile` — one time profile template is shared across the whole time profile family.
  - Templates record the chart type they were created from; the *Apply Chart Template* and *Update Existing Template* menus only offer templates matching the active chart's type. Templates created before the type existed load as time profiles (no converter required).
  - The *Manage Templates* dialog only manages the templates of the active chart's type (templates of other types are preserved untouched), which also allows **one default template per chart type**. Importing a template file created for another chart type is rejected with an error message.
  - The type-unaware `IWithChartTemplates.DefaultChartTemplate` is removed in favor of `DefaultChartTemplateFor(CurveChartTypes)` (breaking change; MoBi/PK-Sim call sites follow in their core-update PRs).
- The constructed marker columns (zero marker, identity line, deviation lines) now carry stable quantity paths (`Time`/`Zero`, merged dimension display name/`Identity`, fold-specific names), so template entries are self-describing, satisfy the template manager validation, and deviation fold styling is matched deterministically. Scatter repositories persisted with residual charts self-heal their base grid path on the next chart update.
- Adding/removing project templates marks the project as changed so templates are not silently lost when closing without saving.

### DIFF
--- a/src/OSPSuite.Assets/UIConstants.cs
+++ b/src/OSPSuite.Assets/UIConstants.cs
@@ -1627,6 +1627,11 @@ namespace OSPSuite.Assets
          return $"Could not load chart template. Make sure that the file '{templateFilePath}' is a template file.";
       }
 
+      public static string CannotLoadTemplateCreatedForAnotherChartType(string templateName, string templateChartType, string managedChartType)
+      {
+         return $"Cannot load the template '{templateName}'. The template was created for charts of type '{templateChartType}' and cannot be used for charts of type '{managedChartType}'.";
+      }
+
       public static string CannotFindResource(string resourceFullPath) => $"Cannot find resource located at '{resourceFullPath}'";
 
       public static string CannotFindSimulationParameterForIdentificationParameter(string fullQuantityPath, string name)
--- a/src/OSPSuite.Core/Chart/AnalysisChart.cs
+++ b/src/OSPSuite.Core/Chart/AnalysisChart.cs
@@ -5,67 +5,64 @@ using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Data;
 using OSPSuite.Utility.Extensions;
 
-namespace OSPSuite.Core.Chart
+namespace OSPSuite.Core.Chart;
+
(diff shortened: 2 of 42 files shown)


rwmcintosh:
﻿@coderabbitai Regarding the outside-diff comment on ResidualsVsTimeChartService (zero-marker x-path "bakes chart-specific ids into the x-path"): this is a misread of the BaseGrid constructor overload. `new BaseGrid($"{id}-Time", "Time", ...)` is the (id, name, dimension) overload, so baseGrid.Name is the literal "Time" - the chart-specific id only lands in the grid's Id. The x-path is therefore the stable `Time` on every chart, including in the self-heal branch.

Yuri05 (approved):

PR #2889, opened 2026-06-15
Fixes #2888 Add manual major interval and minor count override for chart axes

### LINKED ISSUE #2888: Allow manual override of axis major tick interval and minor tick count
## Summary

Chart axis tick spacing is currently always computed automatically (DevExpress auto-grid for numeric axes, a unit-based mapper for time axes, fixed decades for logarithmic axes). There is no way for a user to pin the gridline spacing on an axis.

## Requested behavior

Add two optional, per-axis settings on `Axis`, named to match the DevExpress convention:

- **Major Interval** (`float?`) — spacing between major gridlines, in axis units (maps to DevExpress `NumericScaleOptions.GridSpacing`).
- **Minor Count** (`int?`) — number of minor ticks between two adjacent major ticks (maps to DevExpress `Axis.MinorCount`).

They should behave exactly like the existing `Min`/`Max`:

- Both nullable and independent; when set, they override the automatic tick calculation.
- Clearing a field re-engages the automatic calculation for that field (same UX as clearing `Min`/`Max`).
- Applied to **linearly-scaled axes only** — logarithmic axes keep their decade gridlines and fixed minor count.
- Editable both in the axis-options grid and in the single-axis settings dialog.
- Persisted in the project XML and snapshots, and carried over by chart templates.
- `Major Interval` is unit-specific, so it should be reset when the axis unit changes (like `Min`/`Max`); `Minor Count` is a unit-independent count and should be preserved.

## Acceptance criteria

- Setting Major Interval / Minor Count on a linear axis changes the rendered gridlines accordingly.
- Clearing either field reverts that aspect to the automatic calculation.
- Values round-trip through save/load and snapshots, and through `Clone`/templates.
- Logarithmic axes are unaffected.
- Invalid input is rejected (Major Interval must be > 0, Minor Count must be >= 0).

rwmcintosh:
Fixes #2888

## Summary

Adds two optional, per-axis settings that let the user override the automatic tick calculation for chart axes, mirroring how `Min`/`Max` already work:

- **Major Interval** (`float?`) → DevExpress `NumericScaleOptions.GridSpacing` (spacing between major gridlines, in axis units).
- **Minor Count** (`int?`) → DevExpress `Axis.MinorCount` (number of minor ticks between two major ticks).

## Behavior

- Both are nullable and independent; when set they override the automatic calculation, and clearing a field re-engages auto (same UX as `Min`/`Max`).
- Applied to **linearly-scaled axes only** — logarithmic axes keep their decade gridlines and fixed minor count.
- `MajorInterval` is reset when the axis unit changes (it is expressed in axis units, like `Min`/`Max`); `MinorCount` is unit-independent and preserved.
- Editable both in the axis-options grid and the single-axis settings dialog.
- Persisted in project XML and snapshots, and carried by chart templates / `Clone`.
- Validation: `MajorInterval` must be > 0; `MinorCount` must be within the range DevExpress accepts (greater than 0 and less than 100). The binder also falls back to the automatic count for any out-of-range value, so an invalid value can never crash the chart.

## Tests

- Domain validation and `UpdateFrom` (`AxisSpecs`)
- Snapshot round-trip (`AxisMapperSpecs`) and XML round-trip (`AxisXmlSerializerSpecs` + `AssertForSpecs.AreEqualAxis`)
- Binder behavior (`AxisBinderSpecs`): override applied on linear axes, falls back to auto when cleared or out of range, ignored on log axes

All green: `dotnet build` 0 errors; Core and UI axis specs pass.

image

image

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **New Features**
  * Added chart axis options for **major interval** (gridline/tick spacing) and **minor ticks** (ticks between major ticks).
  * Extended axis settings UI with new fields to configure these values.
* **Validation**
  * Added rules to ensure **major interval > 0** and **minor ticks** is within the allowed range.
* **Updates**
  * Updated axis settings, mapping, and serialization so these values persist and reload correctly.
* **Tests**
  * Added/extended coverage for interval/tick behavior and validation.


rwmcintosh:
@coderabbitai full review please

Yuri05 (approved):

PR #2904, opened 2026-07-14
Fixes #2903 Log scale 0 representation is not clear

### LINKED ISSUE #2903: Log scale 0 representation is not clear
Plotted parameter drops to 0 between 50 and 100 hours, but it appears like a straight line because the y-axis is logarithmic. When the y-axis is switched to linear the y=0 "valley" line is shown.

Image

rwmcintosh:
Fixes #2903

Values that cannot be plotted on a log-scaled axis (e.g. 0) were silently dropped from the bound chart data, so DevExpress drew a straight line connecting the neighboring valid points. Such values are now kept as empty points, which DevExpress renders as a gap in the curve.

Hovering near a gap now shows the underlying data value (e.g. 0) in the tooltip instead of NaN.

Also adds a 'With Values Dropping To Zero' test data set to OSPSuite.Starter for visual verification.

Linear view
image

Before
image

After - it's a single series, but has a gap
image

There is a little bit of code to deal with the actual NaN that would be in the tooltip, replacing it with the data from the data set instead of the curve.
image


msevestre (approved):
nice

PR #2919, opened 2026-07-30
Fixes #2918 Remove a neighborhood redefined with invalid neighbors when merging modules

rwmcintosh:
Fixes #2918

When merging modules, only the neighborhoods whose neighbors resolved were passed to the merge step. A module redefining a neighborhood with unresolvable neighbors therefore had its definition dropped, silently leaving the version created by an earlier module in the model.

Such a redefinition now removes the existing neighborhood, for both `Extend` and `Overwrite`. This makes it possible to remove a neighborhood from another module, and makes the existing "was not added to the simulation" warning accurate.

Covered by two new fixtures in `ModuleIntegrationTests`, one per merge behavior; both fail without the fix. Verified manually in MoBi.

We are implementing this with UI in MoBi to support

Blank neighborhoods are valid
image

Removed neighborhoods are listed. They also have notification messages
image

If you specify an invalid neighbor in any module, the simulation will not build and you will see this error
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved neighborhood handling when referenced neighbors are missing or unresolved.
  * Neighborless definitions no longer attempt invalid path access.
  * Invalid neighborhoods now produce clear errors, while removed neighborhoods generate warnings.
  * Logical neighbors are reported as warnings and remain in the merged model where applicable.
  * Improved handling of neighborhoods that extend or overwrite existing definitions.

* **Tests**
  * Added coverage for missing, logical, neighborless, extended, and overwritten neighborhoods, including serialization and validation scenarios.

### DIFF
--- a/src/OSPSuite.Core/Domain/Mappers/NeighborhoodCollectionToContainerMapper.cs
+++ b/src/OSPSuite.Core/Domain/Mappers/NeighborhoodCollectionToContainerMapper.cs
@@ -10,9 +10,10 @@ using OSPSuite.Utility.Extensions;
 namespace OSPSuite.Core.Domain.Mappers
 {
    /// <summary>
-   ///    Returns a result that includes a top container named "NEIGHBORHOODS" with all mapped neighborhoods
-   ///    and a list of all neighborhoods that were not included because one or both neighbors were not found in the
-   ///    simulation spatial structure
+   ///    Returns a result that includes a top container named "NEIGHBORHOODS" with all mapped neighborhoods,
+   ///    a list of all invalid neighborhoods that could not be created because one or both neighbors were not found in the
+   ///    simulation spatial structure, and a list of all neighborhoods that were removed because a module redefines them
+   ///    without neighbors
    /// </summary>
    internal interface INeighborhoodCollectionToContainerMapper : IMapper<ModelConfiguration, NeighborhoodMapResult>
    {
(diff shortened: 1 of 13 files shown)


rwmcintosh:
@coderabbitai review please

src/OSPSuite.Core/Domain/Mappers/NeighborhoodCollectionToContainerMapper.cs
```diff
@@ -49,34 +50,59 @@ public NeighborhoodMapResult MapFrom(ModelConfiguration modelConfiguration)
...
 
          //now merge all other neighborhoods
          allOtherSpatialStructuresWithMergeBehavior
             .Select(x => new { x.mergeBehavior, neighborhoods = allNeighborHoodsFrom(x.spatialStructure), x.spatialStructure })
             .Each(x =>
             {
                mergeNeighborhoodsInStructure(neighborhoodsParentContainer, definedNeighborhoods(x.neighborhoods), x.mergeBehavior);
-               notDefinedNeighborhoods(x.neighborhoods).Each(n => neighborhoodMapResult.Add(n, x.spatialStructure));
+               invalidNeighborhoods(x.neighborhoods).Each(n => neighborhoodMapResult.AddInvalid(n, x.spatialStructure));
+               neighborhoodsWithoutNeighbors(x.spatialStructure).Each(n => removeNeighborhood(neighborhoodsParentContainer, n, x.spatialStructure, neighborhoodMapResult));
```
rwmcintosh, line 84: I was not sure if there should be a difference between invalid neighborhoods in the first spatial structure vs. those from extension modules. It's a behavior difference where previously, a bad neighborhood was just dropped, now it's an error.

rwmcintosh:
@coderabbitai full review please

src/OSPSuite.Core/Domain/Mappers/NeighborhoodCollectionToContainerMapper.cs
Yuri05: No difference for me

Yuri05 (approved):

PR #2920, opened 2026-07-30
Fixes #2917 Take the administered molecule from the last module when extending

### LINKED ISSUE #2917: Merge behavior "Extend": administered molecule of an application is taken from the first, not the last module
Follow-up to #2811 ("Extend" merge behavior for Events).

## Current behavior

When two modules define an equally-named application (`ApplicationBuilder`) that administers **different** molecules, and the later module has merge behavior **"Extend"**, the *first* module's molecule is administered. The later module's molecule is silently dropped.

Observed in a controlled two-module test:

- Module `A` (first in the hierarchy) defines application `App1` administering molecule `X`.
- Module `B` (selected after `A`, merge behavior "Extend") defines an application `App1` administering molecule `Y`.
- Resulting simulation: `App1` administers **`X`**.

Under merge behavior **"Overwrite"** the later module's molecule (`Y`) wins, which is correct and unchanged from v12.

For reference, in v12 "Extend" produced an application administering **both** `X` and `Y` — a malformed event that used a single molecule's molecular weight. So the v13 behavior is not a regression of a previously correct result; it is a new rule that was, as far as I can tell, never explicitly decided.

## Why I consider this a bug

Everywhere else in the "Extend" rules, a scalar property that cannot be meaningfully merged is **overwritten by the later module**:

- Molecules: type, stationary, calculation methods, parameter type — later module wins.
- Reactions: equation, stoichiometry, operators — later module wins.
- Passive transports / observers: kinetic or monitoring equation — later module wins.
- Event groups: `EventGroupType` — later module wins.
- Parameter values and initial conditions: latest module wins.

The administered molecule of an application is exactly such a non-mergeable scalar, and it is the *only* place where "Extend" gives precedence to the **earlier** module. This is inconsistent, surprising, and silent — no warning or error is raised, and the extending module has no effect at all.

**Expected behavior:** under "Extend", the administered molecule of an equally-named application should be taken from the **later** module, i.e. the same precedence as every other overwritten property (and the same result as "Overwrite").

## Root cause

`SimulationBuilder.mergeEvents` copies only `EventGroupType` and merges `SourceCriteria`; it never transfers `ApplicationBuilder.MoleculeName` from the source (later module) to the target (base) builder:

https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Domain/Builder/SimulationBuilder.cs

```csharp
private void mergeEvents(EventGroupBuilder targetBuilder, BuilderSource<EventGroupBuilder> source)
{
   var sourceBuilder = source.Builder;
   targetBuilder.EventGroupType = sourceBuilder.EventGroupType;

   mergeDescriptorCriteria(targetBuilder.SourceCriteria, sourceBuilder.SourceCriteria);
}
```

`MoleculeName` is a property of `ApplicationBuilder`, not a child entity, so the generic container merge (`tryExtendContainers` → `ContainerMergeTask.MergeContainers`) does not touch it. The base builder therefore keeps the first module's value.

### Secondary effect worth checking with the fix

The `ApplicationMoleculeBuilder` children *are* merged by name (add-or-replace). `EventGroupBuilderToEventGroupMapper.createApplication` then creates one molecule amount per `ApplicationMoleculeBuilder`, each using the single `applicationBuilder.MoleculeName`:

```csharp
foreach (var appMolecule in applicationBuilder.Molecules)
{
   var moleculeContainer = appMolecule.RelativeContainerPath.Resolve<IContainer>(eventGroup);
   var molecule = _moleculeMapper.MapFrom(simulationBuilder.MoleculeByName(applicationBuilder.MoleculeName), moleculeContainer, simulationBuilder);
   ...
}
```

So if the two modules' `ApplicationMoleculeBuilder`s are named differently, the merged application ends up with **two amounts of the same molecule** (whichever `MoleculeName` wins), at the two relative container paths. Fixing `MoleculeName` precedence alone does not address that; please check whether the application molecule list also needs to be overwritten rather than extended when the administered molecule differs.

## Related

- #2811 — "Extend" merge behavior for Events (implemented the current behavior)
- Open-Systems-Pharmacology/OSPSuite.Core#2918 — same class of problem for neighborhoods: an extending module cannot change/remove what an earlier module defined
- Open-Systems-Pharmacology/MoBi#2128 — Application definition rework (event/application combination is expected to change again there; a decision here should be compatible with it)

rwmcintosh:
Fixes #2917

Under merge behavior "Extend", an equally named application kept the **first** module's administered molecule and the extending module had no effect. `mergeEvents` copied only `EventGroupType` and merged `SourceCriteria`; `ApplicationBuilder.MoleculeName` is a property rather than a child entity, so the generic container merge never touched it either.

`mergeApplications` now walks the target and source event group trees in lockstep, pairing children by name, and transfers the administered molecule for every matched application pair. The walk is recursive because applications can sit at the root of the Events building block *or* nested inside event groups — MoBi offers both — and `mergeEvents` is only ever handed the top-level builder. The nested case was verified broken before the fix, so handling only the builder passed to `mergeEvents` would have left the bug reachable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved simulation extension handling for application molecules.
  * When the administered molecule changes, outdated application molecules are removed.
  * Applications nested within event groups now merge correctly.
  * Applications with the same administered molecule retain and combine molecules from both modules.

* **Tests**
  * Added coverage for application and nested event-group merge scenarios.

### DIFF
--- a/src/OSPSuite.Core/Domain/Builder/SimulationBuilder.cs
+++ b/src/OSPSuite.Core/Domain/Builder/SimulationBuilder.cs
@@ -167,6 +167,50 @@ namespace OSPSuite.Core.Domain.Builder
          targetBuilder.EventGroupType = sourceBuilder.EventGroupType;
 
          mergeDescriptorCriteria(targetBuilder.SourceCriteria, sourceBuilder.SourceCriteria);
+
+         mergeApplications(targetBuilder, sourceBuilder);
+      }
+
+      /// <summary>
+      ///    Merges the given event group builders, and all equally named event group builders defined under them, as
+      ///    applications. This is required because <see cref="ApplicationBuilder.MoleculeName" /> is a property and not a
+      ///    child entity and is therefore not covered by the generic container merge
+      /// </summary>
+      private void mergeApplications(EventGroupBuilder targetBuilder, EventGroupBuilder sourceBuilder)
+      {
+         if (targetBuilder is ApplicationBuilder targetApplication && sourceBuilder is ApplicationBuilder sourceApplication)
+            mergeApplication(targetApplication, sourceApplication);
+
+         var targetEventGroups = targetBuilder.GetChildren<EventGroupBuilder>().ToList();
+         foreach (var sourceEventGroup in sourceBuilder.GetChildren<EventGroupBuilder>().ToList())
+         {
+            //The container merge already placed every source event group in the target, so this is only null when the
+            //target uses that name for something that is not an event group. For a group that was added as is, target
+            //and source are the same instance and merging them is a no-op
+            var targetEventGroup = targetEventGroups.FirstOrDefault(x => string.Equals(x.Name, sourceEventGroup.Name));
+            if (targetEventGroup != null)
+               mergeApplications(targetEventGroup, sourceEventGroup);
+         }
+      }
+
+      private void mergeApplication(ApplicationBuilder targetApplication, ApplicationBuilder sourceApplication)
+      {
+         //Application molecules only define where the administered molecule is applied. They are meaningless for another
+         //molecule, so the ones inherited from the base builder are dropped when the administered molecule changes
+         if (!string.Equals(targetApplication.MoleculeName, sourceApplication.MoleculeName))
+            removeApplicationMoleculesNotDefinedIn(targetApplication, sourceApplication);
+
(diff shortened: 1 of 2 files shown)


rwmcintosh:
@coderabbitai review please

Yuri05 (approved):

PR #2937, opened 2026-08-13
Fixes #2936 Take over the value origin when overriding values in a simulation

rwmcintosh:
Fixes #2936

## Problem

When a simulation is created, the value origin carried by a `ParameterValue`, `IndividualParameter`, `ExpressionParameter` or `InitialCondition` was discarded. The value was applied, but the quantity in the simulation kept the value origin it inherited from the underlying builder (spatial structure / molecule).

All three parameter-based overrides share the same apply delegate in `QuantityValuesUpdater`, so the problem was identical for each of them — expression profiles are simply where it was reported.

The value origin *was* already taken over on the branch that creates a parameter that does not yet exist in the model (`ParameterValueToParameterMapper`). Only the far more common branch — the parameter already exists and is overridden — was missing it. `WithUpdatedMetaFrom` copies `Origin` (the `ParameterOrigin`: building block / simulation ids), which is a different concept from `ValueOrigin`.

`InitialCondition` had no value origin handling at all, on either branch.

## Change

`QuantityValuesUpdater`:

- `addOrUpdateParameterFromParameterValue` now takes over the value origin, covering `ParameterValue`, `IndividualParameter` and `ExpressionParameter`. This only happens when the builder effectively overrides a value or a formula, so an entry defining neither cannot reset an existing value origin to `Undefined`.
- `updateMoleculeAmountFromInitialConditions` now takes over the value origin of the `InitialCondition` onto the `MoleculeAmount`, in the same two branches that apply the formula. Using the `MoleculeAmount` keeps this symmetric with `InitialConditionsCreator.CreateInitialCondition`, which reads the value origin back from it.

## Tests

New integration tests in `QuantityValuesUpdaterIntegrationTests` seed the model quantities with a value origin and assert that each of the four override types replaces it, plus one test asserting that builders defining neither value nor formula leave the value origin untouched.

### DIFF
--- a/src/OSPSuite.Core/Domain/Services/QuantityValuesUpdater.cs
+++ b/src/OSPSuite.Core/Domain/Services/QuantityValuesUpdater.cs
@@ -188,6 +188,10 @@ namespace OSPSuite.Core.Domain.Services
          var (_, simulationBuilder, replacementContext) = valueUpdater.ModelConfiguration;
          _entityTracker.Track(parameter, parameterValue, simulationBuilder);
 
+         //the value origin is only taken over if the value or formula is effectively overridden
+         if (parameterValue.Formula != null || parameterValue.Value.IsFinite())
+            parameter.ValueOrigin.UpdateAllFrom(parameterValue.ValueOrigin);
+
          //Formula is defined, we update in the parameter instance
          if (parameterValue.Formula != null)
          {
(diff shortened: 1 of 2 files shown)


rwmcintosh:
@coderabbitai review please

Yuri05 (approved):

